PR #57, opened 2017-05-02
Add Travis integration

Sam-Burns:
It's just a suggestion, but would you like to use Travis as a cloud CI service for this project?  I know it is the one that was chosen for other Symfony projects.

If you want, you can use this PR, but you would also need to enable Travis for the project on the [Travis website](https://travis-ci.org/profile/).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,0 +1,21 @@
+language: php
+
+sudo: false
+
+matrix:
+  include:
+    - php: 7.1
+    - php: 7.1
+      env: DEPENDENCIES='low'
+    - php: nightly
+  allow_failures:
+    - php: nightly
+  fast_finish: true
+
+install:
+  - composer self-update
+  - if [ "$DEPENDENCIES" != "low" ]; then composer update; fi;
+  - if [ "$DEPENDENCIES" == "low" ]; then composer update --prefer-lowest; fi;
+
+script:
+  - ./vendor/bin/simple-phpunit
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,7 +6,7 @@ matrix:
   include:
     - php: 7.1
     - php: 7.1
-      env: DEPENDENCIES='low'
+      env: COMPOSER_FLAGS='--prefer-lowest'
     - php: nightly
   allow_failures:
     - php: nightly
@@ -20,10 +20,10 @@ cache:
 env:
   global:
     - SYMFONY_PHPUNIT_DIR="$HOME/symfony-bridge/.phpunit"
+    - COMPOSER_FLAGS=''
 
 install:
-  - if [ "$DEPENDENCIES" != "low" ]; then composer update; fi;
-  - if [ "$DEPENDENCIES" == "low" ]; then composer update --prefer-lowest; fi;
+  - composer update $COMPOSER_FLAGS
 
 script:
   - ./vendor/bin/simple-phpunit
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,8 +12,16 @@ matrix:
     - php: nightly
   fast_finish: true
 
+cache:
+  directories:
+    - $HOME/.composer/cache/files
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -1,3 +1,5 @@
+[![Build Status](https://travis-ci.org/symfony/flex.svg?branch=master)](https://travis-ci.org/symfony/flex)
```
javiereguiluz: In the past it was decided that Symfony repositories don't include these badges in their READMEs, so we should remove this. Thanks!

Sam-Burns: Removed

.travis.yml
```diff
@@ -0,0 +1,21 @@
...
+      env: DEPENDENCIES='low'
+    - php: nightly
+  allow_failures:
+    - php: nightly
+  fast_finish: true
+
+install:
+  - composer self-update
+  - if [ "$DEPENDENCIES" != "low" ]; then composer update; fi;
+  - if [ "$DEPENDENCIES" == "low" ]; then composer update --prefer-lowest; fi;
```
Pierstoval: Could be simplified by setting an `env` var to `COMPOSER_FLAGS` and inject it automatically without any `if` statement, like this:

```yaml

env:
  COMPOSER_FLAGS: ''
matrix:
  include:
    - php: 7.1
    - php: 7.1
      env: COMPOSER_FLAGS='--prefer-lowest'
    - php: nightly
  allow_failures:
    - php: nightly
  fast_finish: true

install:
  - composer self-update
  - composer update $COMPOSER_FLAGS
```

.travis.yml
```diff
@@ -0,0 +1,21 @@
+language: php
+
+sudo: false
+
```
stof: I suggest also adding the cache for composer downloads, and for the simple-phpunit downloads of PHPUnit):

```
cache:
  directories:
    - $HOME/.composer/cache/files
    - $HOME/symfony-bridge/.phpunit

env:
  global:
    - SYMFONY_PHPUNIT_DIR="$HOME/symfony-bridge/.phpunit"

File: .travis.yml [hunk 4]
```diff
@@ -0,0 +1,21 @@
...
+    - php: 7.1
+    - php: 7.1
+      env: DEPENDENCIES='low'
+    - php: nightly
+  allow_failures:
+    - php: nightly
+  fast_finish: true
+
+install:
+  - composer self-update
```
stof: useless. Travis already ensures that Composer is uptodate now

.travis.yml
stof: @nicolas-grekas having the bridge storing `.phpunit` inside `vendor/symfony/phpunit-bridge/bin` by default makes it hard to cache this folder on Travis (caching the folder inside the vendor folder won't work fine as it will be overwritten when Composer installs `symfony/phpunit-bridge`, and so it requires overwriting the location). Can't we find an easier location ?

nicolas-grekas: `SYMFONY_PHPUNIT_DIR` is here for that reason. Not sure if we can have a better default.

.travis.yml
Sam-Burns: @Pierstoval I've changed it

.travis.yml
Sam-Burns: The cache directories are added now

.travis.yml
```diff
@@ -0,0 +1,29 @@
...
+    - $HOME/.composer/cache/files
+    - $HOME/symfony-bridge/.phpunit
+
+env:
+  global:
+    - SYMFONY_PHPUNIT_DIR="$HOME/symfony-bridge/.phpunit"
+    - COMPOSER_FLAGS=''
+
+install:
+  - composer update $COMPOSER_FLAGS
```
xabbuh: You can also run `./vendor/bin/simple-phpunit install` in this step so that the installation output is folded by default.

fabpot:
Thank you @Sam-Burns.

PR #60, opened 2017-05-02
PHPUnit version bump

Sam-Burns:
Increases PHPUnit version from `5.4` to `6.1`.

Fixes [nightly build](https://travis-ci.org/symfony/flex/jobs/227983361), which is currently an allowable failure.

(The problem was that PHPUnit 5.4 returns non-zero exit code in PHP master because [each() will be deprecated](https://github.com/php/php-src/commit/06a034016280435feb80eea4d674ca5688ab4c06) in 7.2.)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,8 +8,6 @@ matrix:
     - php: 7.1
       env: COMPOSER_FLAGS='--prefer-lowest'
     - php: nightly
-  allow_failures:
-    - php: nightly
   fast_finish: true
 
 cache:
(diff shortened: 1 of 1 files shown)


fabpot:
What about moving PHP nighly to the regular section?

Sam-Burns:
Happy to do that as well, if you would like that in the same PR

fabpot:
Same PR is fine

fabpot:
Thank you @Sam-Burns.

PR #64, opened 2017-05-02
Simplifying autoloader

Sam-Burns:
Since the namespace was last changed, the `classmap` section of the Composer autoloader settings is not necessary.  This file is autoloaded anyway because of the `psr-4` section below.


fabpot:
Thank you @Sam-Burns.

PR #65, opened 2017-05-02
Unit test for Recipe class

Sam-Burns:
Adds a unit test for `Symfony\Flex\Recipe`

### DIFF
--- a/tests/RecipeTest.php
+++ b/tests/RecipeTest.php
@@ -1,0 +1,62 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Flex\Tests;
+
+use Composer\Package\Package;
+use PHPUnit\Framework\TestCase;
+use Symfony\Flex\Recipe;
+
+class RecipeTest extends TestCase
+{
+    /**
+     * @var Recipe
+     */
+    private $recipe;
+
+    /**
+     * @var Package|PHPUnit_Framework_MockObject_MockObject
+     */
+    private $package;
+
+    protected function setUp()
+    {
+        $this->package = $this->getMockBuilder(Package::class)->disableOriginalConstructor()->getMock();
+        $this->recipe = new Recipe(
+            $this->package,
+            'name',
+            [
+                'files' => ['file1'],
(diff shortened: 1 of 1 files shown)


tests/RecipeTest.php
```diff
@@ -0,0 +1,63 @@
...
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Flex\Tests;
+
+use Composer\Package\Package;
+use PHPUnit_Framework_MockObject_MockObject as Mock;
```
stof: we never add use statements for classes in the global namespace

tests/RecipeTest.php
```diff
@@ -0,0 +1,63 @@
...
+     * @var Recipe
+     */
+    private $recipe;
+
+    /**
+     * @var Package|Mock
+     */
+    private $package;
+
+    public function setUp()
```
stof: should be protected

tests/RecipeTest.php
TomasVotruba: @stof Just curious, is there a fixer for that to automate this in reviews?

stof: I don't think PHP-CS-Fixer supports it currently.

tests/RecipeTest.php
```diff
@@ -0,0 +1,62 @@
...
+
+class RecipeTest extends TestCase
+{
+    /**
+     * @var Recipe
+     */
+    private $recipe;
+
+    /**
+     * @var Package|PHPUnit_Framework_MockObject_MockObject
```
maxhelias: I think, it should be : `@var Package|PHPUnit\Framework\MockObject\MockObject`

tests/RecipeTest.php
```diff
@@ -0,0 +1,62 @@
...
+     */
+    private $package;
+
+    protected function setUp()
+    {
+        $this->package = $this->getMockBuilder(Package::class)->disableOriginalConstructor()->getMock();
+        $this->recipe = new Recipe(
+            $this->package,
+            'name',
+            [
```
maxhelias: The `job` parameters are missing : https://github.com/symfony/flex/blob/master/src/Recipe.php#L26

tests/RecipeTest.php
localheinz: Actually, it should be `@var Package|\PHPUnit\Framework\MockObject\MockObject` (note the missing leading `\`).

PR #66, opened 2017-05-03
Expand special %DIR% variables in post-install-output

chalasr:
Fixes https://github.com/symfony/recipes/issues/12

### DIFF
--- a/src/Flex.php
+++ b/src/Flex.php
@@ -130,7 +130,10 @@ class Flex implements PluginInterface, EventSubscriberInterface
 
             $manifest = $recipe->getManifest();
             if (isset($manifest['post-install-output'])) {
-                $this->postInstallOutput = array_merge($this->postInstallOutput, $manifest['post-install-output'], ['']);
+                foreach ($manifest['post-install-output'] as $line) {
+                    $this->postInstallOutput[] = $this->options->expandTargetDir($line);
+                }
+                $this->postInstallOutput[] = '';
             }
         }
     }
--- a/tests/FlexTest.php
+++ b/tests/FlexTest.php
@@ -1,0 +1,78 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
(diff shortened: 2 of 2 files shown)


src/Flex.php
```diff
@@ -130,7 +130,7 @@ public function configurePackage(PackageEvent $event): void
 
             $manifest = $recipe->getManifest();
             if (isset($manifest['post-install-output'])) {
-                $this->postInstallOutput = array_merge($this->postInstallOutput, $manifest['post-install-output'], ['']);
+                $this->postInstallOutput = array_merge($this->postInstallOutput, $this->options->expandTargetDir($manifest['post-install-output']), ['']);
```
stof: this is broken. `$manifest['post-install-output']` is an array of lines. You need to loop over the array

stof:
btw, it means this code is untested (or the test is wrong)

sstok:
It seems to be untested.

src/Flex.php
```diff
@@ -130,7 +130,12 @@ public function configurePackage(PackageEvent $event): void
 
             $manifest = $recipe->getManifest();
             if (isset($manifest['post-install-output'])) {
-                $this->postInstallOutput = array_merge($this->postInstallOutput, $manifest['post-install-output'], ['']);
+                $postInstallOutput = [];
+                foreach ($manifest['post-install-output'] as $line) {
+                    $postInstallOutput[] = $this->options->expandTargetDir($line);
+                }
+
+                $this->postInstallOutput = array_merge($this->postInstallOutput, $postInstallOutput, ['']);
```
stof: instead of building an array `$postInstallOutput` by adding elements in it one by one, and then merging it in `$this->postInstallOutput`, you could add elements in `$this->postInstallOutput` directly in your loop (and then adding the empty string in the array after the loop)

chalasr:
The whole Flex class is untested, added a first test and addressed @stof comments

fabpot:
Thank you @chalasr.

Issue #68, opened 2017-05-04
Composer version requirement


Pierstoval:
After testing, it seems that Flex is compatible with Composer 1.3.1+ (I had an error with 1.3.0 when installing symfony/stopwatch, not sure it comes from composer though, and installing any bundle with 1.2 didn't work).

How can we tell, when installing Flex, that the user needs to upgrade its composer version if it's not compatible?

stof:
Which kind of error do you get with composer 1.2 ?

Pierstoval:
It's an error related to the `--no-suggest` option 😕 

```bash
pierstoval@PIERSTOVAL /mnt/e/dev/www/flex_tests/cms (master) $ composer -V
Composer version 1.2.4 2016-12-06 22:00:51
pierstoval@PIERSTOVAL /mnt/e/dev/www/flex_tests/cms (master) $ composer req orbitale/cms-bundle -vvv
Reading ./composer.json
Loading config file ./composer.json
Checked CA file /etc/ssl/certs/ca-certificates.crt: valid
Executing command (/mnt/e/dev/www/flex_tests/cms): git branch --no-color --no-abbrev -v
Failed to initialize global composer: Composer could not find the config file: /home/pierstoval/.composer/composer.json
To initialize a project, please create a composer.json file as described in the https://getcomposer.org/ "Getting Started" section
Reading /mnt/e/dev/www/flex_tests/cms/vendor/composer/installed.json
Loading plugin Symfony\Flex\Flex
Running 1.2.4 (2016-12-06 22:00:51) with PHP 7.1.4-1+deb.sury.org~xenial+1 on Linux / 3.4.0+


  [Symfony\Component\Console\Exception\InvalidArgumentException]
  The "no-suggest" option does not exist.


Exception trace:
 () at phar:///home/pierstoval/bin/composer/vendor/symfony/console/Input/Input.php:169
 Symfony\Component\Console\Input\Input->setOption() at /mnt/e/dev/www/flex_tests/cms/vendor/symfony/flex/src/Command/RequireCommand.php:33
 Symfony\Flex\Command\RequireCommand->execute() at phar:///home/pierstoval/bin/composer/vendor/symfony/console/Command/Command.php:259
 Symfony\Component\Console\Command\Command->run() at phar:///home/pierstoval/bin/composer/vendor/symfony/console/Application.php:847
[...]

PR #69, opened 2017-05-04
Test with different versions of Composer

Pierstoval:
Just trying this config to test over different versions of Composer.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,11 +2,19 @@ language: php
 
 sudo: false
 
+env:
+    global:
+        - COMPOSER_VERSION=""
+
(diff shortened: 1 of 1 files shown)


stof:
this won't help much, as there is no functional tests covering the overwritten commands, and so the issue you have would not be catched on CI.
Tests should be written first.

Pierstoval:
Yep, it's just a POC but you're right, we should test more things to make sure each composer version is covered. I'm gonna work on a reproduction of the bug I issued in #68

fabpot:
Closing for now as there is no much value in merging this one as is.

PR #77, opened 2017-05-09
Some CS fixes

Pierstoval:
Mostly phpdoc for IDEs when needed (self-documented parameters are not useful to document, unless there's documentation about exceptions, this is why sometimes there is phpdoc and sometimes not).

Also, there was some unused classes and some missing strict checks .

### DIFF
--- a/src/Downloader.php
+++ b/src/Downloader.php
@@ -185,7 +185,7 @@ class Downloader
         $this->degradedMode = true;
     }
 
-    private function getOptions(array $headers): array 
+    private function getOptions(array $headers): array
     {
         $options = ['http' => ['header' => $headers]];
 
--- a/src/Flex.php
+++ b/src/Flex.php
@@ -27,9 +27,7 @@ use Composer\IO\NullIO;
 use Composer\Json\JsonFile;
 use Composer\Json\JsonManipulator;
 use Composer\Package\PackageInterface;
-use Composer\Plugin\CommandEvent;
 use Composer\Plugin\PluginInterface;
-use Composer\Plugin\PluginEvents;
 use Composer\Script\Event;
 use Composer\Script\ScriptEvents;
 use Symfony\Component\Console\Application as ConsoleApplication;
@@ -198,7 +196,7 @@ class Flex implements PluginInterface, EventSubscriberInterface
         $this->io->write($this->postInstallOutput);
     }
 
-    private function filterPackageNames(PackageInterface $package, string $operation)
+    private function filterPackageNames(PackageInterface $package, string $operation): \Generator
     {
         // FIXME: getNames() can return n names
         $name = $package->getNames()[0];
@@ -211,7 +209,9 @@ class Flex implements PluginInterface, EventSubscriberInterface
             }
         }
 
-        if (200 === $response->getStatusCode() || 304 === $response->getStatusCode()) {
+        $statusCode = $response->getStatusCode();
+
+        if (200 === $statusCode || 304 === $statusCode) {
             yield $name => new Recipe($package, $name, $response->getBody());
         } elseif ('symfony-bundle' === $package->getType()) {
             $manifest = [];
--- a/src/PackageResolver.php
+++ b/src/PackageResolver.php
@@ -19,10 +19,8 @@ use Composer\Package\Version\VersionParser;
 class PackageResolver
 {
     private const SYMFONY_VERSIONS = ['lts', 'previous', 'stable', 'next'];
-
     private static $aliases;
     private static $versions;
-
     private $downloader;
 
     public function __construct(Downloader $downloader)
@@ -61,7 +59,7 @@ class PackageResolver
                         $versionParser->parseConstraints($argument);
                     } catch (\UnexpectedValueException $e) {
                         // is it a special Symfony version?
-                        if (!in_array($argument, self::SYMFONY_VERSIONS)) {
+                        if (!in_array($argument, self::SYMFONY_VERSIONS, true)) {
                             $this->throwAlternatives($argument, $i);
                         }
                     }
@@ -72,7 +70,7 @@ class PackageResolver
         }
 
         // third pass to resolve versions
-        $requires= [];
+        $requires = [];
         foreach ($versionParser->parseNameVersionPairs($packages) as $package) {
             $requires[] = $package['name'].$this->parseVersion($package['name'], $package['version'] ?? '');
         }
@@ -96,7 +94,7 @@ class PackageResolver
 
         if ('next' === $version) {
             $version = '^'.self::$versions[$version].'@dev';
-        } elseif (in_array($version, self::SYMFONY_VERSIONS)) {
+        } elseif (in_array($version, self::SYMFONY_VERSIONS, true)) {
             $version = '^'.self::$versions[$version];
         }
 
@@ -120,7 +118,7 @@ class PackageResolver
         if ($alternatives || 0 === $position) {
             $message = sprintf('"%s" is not a valid alias.', $argument);
             if ($alternatives) {
-                if (1 == count($alternatives)) {
+                if (1 === count($alternatives)) {
                     $message .= " Did you mean this:\n";
                 } else {
                     $message .= " Did you mean one of these:\n";
--- a/src/Response.php
+++ b/src/Response.php
@@ -63,7 +63,7 @@ class Response implements \JsonSerializable
             throw new \LogicException('Old cache detected. Clear the Composer cache under "~/.composer/cache/repo/https---symfony.sh/".');
         }
 
-        $response = new Response($json['body']);
+        $response = new self($json['body']);
         $response->headers = $json['headers'];
 
         return $response;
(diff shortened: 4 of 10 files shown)


src/PackageResolver.php
```diff
@@ -20,9 +20,19 @@ class PackageResolver
 {
     private const SYMFONY_VERSIONS = ['lts', 'previous', 'stable', 'next'];
 
+    /**
+     * @var array
```
ogizanagi: `string[]` ?

src/PackageResolver.php
```diff
@@ -20,9 +20,19 @@ class PackageResolver
...
+     */
     private static $aliases;
+
+    /**
+     * @var array
```
ogizanagi: `string[]` ?

(Also, I personally prefer the inlined version:

```php
/** @var string[] */
private static $versions;
```
, but may be only me)

src/PackageResolver.php
Pierstoval: Good catch 🙂

src/PackageResolver.php
Pierstoval: I don't like the inlined version, I have not executed php-cs-fixer on the package itself (not sure it's fully compatible with PHP7.1 yet, but moreover, I was lazy 😛), but I'm not sure it would allow inlined comments 🤔

ogizanagi: Sure it allows it :) I'm using this each of my projects.

src/Flex.php
```diff
@@ -39,13 +37,41 @@
...
+     */
     private $downloader;
+
+    /**
+     * @var array
+     */
     private $postInstallOutput = [''];
+
+    /**
+     * @var \Closure|InstallCommand
```
chalasr: `\Closure` only

Pierstoval: You're right, I didn't see that it was the return of this var that was type checked, not the var itself 👍

src/Flex.php
```diff
@@ -198,7 +227,7 @@ public function executeAutoScripts(Event $event): void
         $this->io->write($this->postInstallOutput);
     }
 
-    private function filterPackageNames(PackageInterface $package, string $operation)
+    private function filterPackageNames(PackageInterface $package, string $operation): ?\Generator
```
stof: this is not nullable. It is always a generator

src/ScriptExecutor.php
```diff
@@ -39,9 +54,12 @@ public function __construct(Composer $composer, IOInterface $io, Options $option
     }
 
     /**
+     * @param string $type
+     * @param string $cmd
+     *
      * @throw ScriptExecutionException if the executed command returns a non-0 exit code
```
stof: the tag is invalid. It should be `@throws`. Can you fix it too ?

src/Flex.php
Pierstoval: Fixed, but are you _absolutely certain_ that nothing in the `elseif` could have an edge case and therefore not return the expected `\Generator`?

src/ScriptExecutor.php
Pierstoval: Nice catch, fixed 👍

src/Flex.php
jvasseur: As soon as the body of a fonction contains a yield, it will return a generator (possibly an empty one): https://3v4l.org/FgHWh

chalasr: yes, using `yield` in a function makes it a generator function (no matter if the `yield` statement is actually reached) thus always return a `\Generator` instance.

Pierstoval: Nice, didn't know that, thanks 👍

src/Downloader.php
```diff
@@ -68,6 +68,8 @@ public function allowContrib(bool $allow): void
      *
      * @param string $path    The path to get on the server
      * @param array  $headers An array of HTTP headers
+     *
+     * @return Response
```
fabpot: Why is it useful?

src/Flex.php
```diff
@@ -117,6 +143,9 @@ public function configureProject(Event $event): void
         file_put_contents($json->getPath(), $manipulator->getContents());
     }
 
+    /**
+     * @param PackageEvent $event
+     */
```
fabpot: Why is it useful?

src/PackageResolver.php
```diff
@@ -96,14 +106,17 @@ private function parseVersion(string $package, string $version): string
...
             $version = '^'.self::$versions[$version];
         }
 
         return ':'.$version;
     }
 
     /**
+     * @param string $argument
+     * @param int    $position
+     *
```
fabpot: Again, why is it useful, seems redundant to me.

src/Response.php
```diff
@@ -22,7 +22,9 @@ class Response implements \JsonSerializable
     private $code;
 
     /**
-     * @param mixed $body The response as JSON
+     * @param mixed    $body The response as JSON
+     * @param iterable $headers
+     * @param int      $code
      */
```
fabpot: Don't see the value of adding those comments

src/ScriptExecutor.php
```diff
@@ -39,9 +54,12 @@ public function __construct(Composer $composer, IOInterface $io, Options $option
     }
 
     /**
-     * @throw ScriptExecutionException if the executed command returns a non-0 exit code
+     * @param string $type
+     * @param string $cmd
```
fabpot: To be removed, redundant.

src/PackageResolver.php
```diff
@@ -20,9 +20,19 @@ class PackageResolver
...
+     */
     private static $aliases;
+
+    /**
+     * @var string[]
+     */
     private static $versions;
 
+    /**
+     * @var Downloader
```
ro0NL: PHPStorm detects it..  :) usually left out in sf codebase as well.

src/ScriptExecutor.php
```diff
@@ -25,9 +25,24 @@
  */
 class ScriptExecutor
 {
+    /**
+     * @var Composer
+     */
```
yceruto: PHPStorm is able to auto-detect all properties initialized into `__constructor()`

javiereguiluz:
@Pierstoval this is great ... but if you use PHPStorm (72% of Symfony devs do) there's no need to add PHPdoc for method arguments, method response types and properties initialized in the constructor. So you can safely remove some of these changes. Thanks!

src/Downloader.php
Pierstoval: IDEs & cs-fixers consider that if you set a "partial" phpdoc, it's inconsistent. But yes, we shouldn't consider IDEs as almighty 🙂 I'm gonna remove useless doc

Pierstoval:
I did remove all redundant comments, an ran php-cs-fixer too, to fix some very minor stuff (mostly quotes)

src/Flex.php
```diff
@@ -39,13 +37,41 @@
...
     private $downloader;
+
+    /**
+     * @var array
+     */
     private $postInstallOutput = [''];
+
+    /**
+     * @var \Closure
+     */
```
fabpot: Not a big fan of documenting private stuff. If others think this is useful (I don't), can we at least put all `@var` blocks on one line?

chalasr: I don't think it's useful either, it's an internal detail.

Pierstoval: Ok, I removed it in the last commit 👍

yceruto: > [...] all `@var` blocks [...]

You only changed this one ;)

Pierstoval: Oh yes, sorry. Done.
Only for this file then? :)

fabpot: My question was more about the usefulness of these annotations. Apparently, @chalasr agrees with me. So I think removing them altogether is the way to go.

Pierstoval: Actually, IDEs recognize constructor arguments, but not properties that are defined in other methods, so to have autocomplete & property type recognition in IDEs we need these doc blocks because they're not initialized in constructor.

But if you really feel they're not useful, I can remove them

fabpot: I'm not using an IDE, so there are useless for me :) Just wanted to get some more feedback on that.

ro0NL: Technically they should be `Some|null` also.. but i agree with @chalasr as well on not documenting private stuff.

in general :+1: for inlined :)

Pierstoval: So, do I inline or remove? 😕

fabpot: AFAIU, it looks like at least the 3 of us want to remove those annotations.

Pierstoval: Ok, removed then :

src/PackageResolver.php
```diff
@@ -20,7 +20,14 @@ class PackageResolver
...
     private const SYMFONY_VERSIONS = ['lts', 'previous', 'stable', 'next'];
 
+    /**
+     * @var string[]
+     */
     private static $aliases;
+
+    /**
+     * @var string[]
+     */
```
fabpot: same here, those two comments should be removed

Pierstoval: Done

fabpot:
Thank you @Pierstoval.

PR #79, opened 2017-05-09
Make Flex compatible with Composer <1.3

### LINKED ISSUE #68: Composer version requirement

Pierstoval:
If the option doesn't exist in the definition, an error is thrown.

This makes flex compatible with Composer versions that do not have this option in the command definition.

Fixes #68


fabpot:
Thank you @Pierstoval.

PR #84, opened 2017-05-16
clean opcode caches after writing the bundles file

### LINKED ISSUE #82: `composer req admin` fails with "There is no extension able to load the configuration for "doctrine"
I've tried running this immediately after `composer clear-cache` with no success.

The exact command I'm running is:
`composer -v create-project symfony/skeleton:3.3.x-dev demo && cd demo && composer req webserver && composer req cli && composer req admin && ./bin/console server:start`

### `php -v` output:
```
PHP 7.1.4-1+deb.sury.org~xenial+1 (cli) (built: Apr 11 2017 22:12:32) ( NTS )
Copyright (c) 1997-2017 The PHP Group
Zend Engine v3.1.0, Copyright (c) 1998-2017 Zend Technologies
    with Zend OPcache v7.1.4-1+deb.sury.org~xenial+1, Copyright (c) 1999-2017, by Zend Technologies`
```

### `composer -V` output:
```
Composer version 1.4.1 2017-03-10 09:29:45
```

### Error:
```
Executing script assets:install --symlink --relative %WEB_DIR% [KO]
 [KO]
Script assets:install --symlink --relative %WEB_DIR% returned with error code 1
!!  
!!  
!!                                                                                 
!!    [Symfony\Component\Config\Exception\FileLoaderLoadException]                 
!!    There is no extension able to load the configuration for "doctrine" (in /ho  
!!    me/gus/Projects/Learning/demo/etc/packages/doctrine.yaml). Looked for names  
!!    pace "doctrine", found "framework", "web_server", "easy_admin" in /home/gus  
!!  
!!    /Projects/Learning/demo/etc/packages/doctrine.yaml (which is loaded in reso  
!!    urce "/home/gus/Projects/Learning/demo/etc/packages/doctrine.yaml").         
!!                                                                                 
!!  
!!  
!!                                                                                 
!!    [Symfony\Component\DependencyInjection\Exception\InvalidArgumentException]   
!!    There is no extension able to load the configuration for "doctrine" (in /ho  
!!    me/gus/Projects/Learning/demo/etc/packages/doctrine.yaml). Looked for names  
!!    pace "doctrine", found "framework", "web_server", "easy_admin"               
!!  
!!                                                                                 
!!  
!!  
!!  

Installation failed, reverting ./composer.json to its original content.
```

### Full console output is provided below:
[flex_debug.txt](https://github.com/symfony/flex/files/994671/flex_debug.txt)

xabbuh:
This should fix #82.

### DIFF
--- a/src/Configurator/BundlesConfigurator.php
+++ b/src/Configurator/BundlesConfigurator.php
@@ -90,6 +90,10 @@ class BundlesConfigurator extends AbstractConfigurator
         }
 
         file_put_contents($file, $contents);
+
+        if (function_exists('opcache_invalidate')) {
+            opcache_invalidate($file);
+        }
     }
 
     private function getConfFile(): string


src/Configurator/BundlesConfigurator.php
```diff
@@ -90,6 +90,14 @@ private function dump(string $file, iterable $bundles): void
         }
 
         file_put_contents($file, $contents);
+
+        if (function_exists('opcache_invalidate')) {
+            opcache_invalidate($file);
+        }
+
+        if (function_exists('apc_delete_file')) {
+            apc_delete_file($file);
```
stof: this looks useless. Flex requires PHP 7.1+, and the APC extension exists only for PHP 5.5-

xabbuh: Are you sure that it is really not possible to install the extension with PHP 7? I did never try that and I also think that doing so is not a common use case. But I did not find any information on that it is not possible.

Pierstoval: Isn't there an APCU equivalent for PHP7?

stof: APCu replaces the user caching part of APC, not the opcode cache part (as opcode caching is handled by OPCache).

And yes I'm sure it does not work with PHP 7. Extensions require a rewrite for PHP 7, especially when they are as low level as an opcode cache. APC would not even compile on PHP 7.

xabbuh: alright, I removed the `apc_delete_file()` call

fabpot:
Has anyone any idea why this is needed? That means that on such configurations, people need to clear their Opcache cache whenever they save a file? I don't really understand what's going on here.

fabpot:
Thank you @xabbuh.

stof:
@fabpot you can configure how often OPCache will revalidate files. It will not check all the time. And so, when running with OPCache enabled, editing a file would not give you the right content if you require it just after while you already required it before the edition (unless you spend more time doing the change than the opcache revalidation frequency, but the default setting is to revalidate only once every 2 seconds).
I think the big difference for people trying flex is that `opcache.enable_cli` is ON by default as of PHP 7.1.2. And flex is typically in a case where it can edit the bundles.php file multiple times in a row and then requiring it again (this happens when your `composer require` command triggers the installation of at least 2 bundles)

PR #85, opened 2017-05-16
Use correct typehint 

emirb:
Simple fix for compatibility mismatch between the actual usage (always passing iterable) and the type hint (array) revealed by phpstan.

### DIFF
--- a/src/Downloader.php
+++ b/src/Downloader.php
@@ -185,7 +185,7 @@ class Downloader
         $this->degradedMode = true;
     }
 
-    private function getOptions(array $headers): array
+    private function getOptions(iterable $headers): array
     {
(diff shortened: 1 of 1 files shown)


src/Downloader.php
```diff
@@ -185,7 +185,7 @@ private function switchToDegradedMode(\Exception $e, string $url): void
         $this->degradedMode = true;
     }
 
-    private function getOptions(array $headers): array
+    private function getOptions(iterable $headers): array
```
stof: this looks wrong to me. Passing an iterator will not be allowed. The headers inside the options of the Composer RemoteFilesystem must be an array, not any iterable

PR #98, opened 2017-06-10
Add the recipe origin in the Composer output

### LINKED ISSUE #97: Not reporting auto-configuration is ignored for contrib recipe
When I run composer install, Flex reports an auto-configuration is detected, however when the recipe is part of contrib it doesn't report the recipe itself ignored.

I thought this was bug, but I forgot to set the allow-contrib to true :)
Maybe it's a good idea to inform there is a (contrib) recipe but it was ignored?

```
- Installing pomm-project/pomm-bundle (2.3.1): Loading from cache
    Detected auto-configuration settings for "pomm-project/pomm-bundle" (contrib-recipes disabled)
    Enabling the package as a Symfony bundle
```

fabpot:
fixes #97

### DIFF
--- a/src/Downloader.php
+++ b/src/Downloader.php
@@ -169,7 +169,7 @@ class Downloader
         }
 
         $response = new Response($data, $this->rfs->getLastHeaders());
-        if (null !== $response->getHeader('last-modified')) {
+        if ($response->getHeader('last-modified')) {
             $this->cache->write($cacheKey, json_encode($response));
         }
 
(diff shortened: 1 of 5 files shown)


fabpot:
This new output will also help debugging issue and error reporting.

fabpot:
When using a PR, it tells you precisely which recipe is used:

image

sstok:
Thank you @fabpot 😄

PR #99, opened 2017-06-12
Change configurator messages to be only displayed on verbose mode

fabpot:


src/Configurator/AbstractConfigurator.php
```diff
@@ -35,4 +35,15 @@ public function __construct(Composer $composer, IOInterface $io, Options $option
     abstract public function configure(Recipe $recipe, $config): void;
 
     abstract public function unconfigure(Recipe $recipe, $config): void;
+
+    protected function write($messages)
```
stof: shouldn't it be a void-returning function ?

src/Configurator/AbstractConfigurator.php
```diff
@@ -35,4 +35,15 @@ public function __construct(Composer $composer, IOInterface $io, Options $option
...
+    {
+        if (!is_array($messages)) {
+            $messages = [$messages];
+        }
+        foreach ($messages as $i => $message) {
+            $messages[$i] = '    '.$message;
+        }
+        return $this->io->write($messages, true, IOInterface::VERBOSE);
```
stof: `IOInterface::write` does not return anything

src/Configurator/AbstractConfigurator.php
fabpot: done

src/Configurator/AbstractConfigurator.php
fabpot: indeed

PR #101, opened 2017-06-13
Optimize performance by getting all recipes in one HTTP request

### LINKED ISSUE #80: Smarter auto-enabling of bundles without recipes
If you execute this and the bundle doesn't have a recipe:

```
$ composer require vendor-name/package-name
```

Flex autodetects that this is a Symfony Bundle (if composer.json type is symfony-bundle) and it enables for all environments:

```php
return [
  // ...
  'VendorName\Bundle\PackageNameBundle\VendorNamePackageNameBundle' => ['all' => true],
];
```

However, if you execute this:

```
$ composer require --dev vendor-name/package-name
```

The result is the same. Shouldn't we detect the `--dev` option and then enable the bundle only for dev and test?

```php
return [
  // ...
  'VendorName\Bundle\PackageNameBundle\VendorNamePackageNameBundle' => ['dev' => true, 'test' => true],
];
```

fabpot:
Also fixed #80

### DIFF
--- a/src/Downloader.php
+++ b/src/Downloader.php
@@ -13,6 +13,9 @@ namespace Symfony\Flex;
 
 use Composer\Cache;
 use Composer\Composer;
+use Composer\DependencyResolver\Operation\OperationInterface;
+use Composer\DependencyResolver\Operation\UpdateOperation;
+use Composer\DependencyResolver\Operation\UninstallOperation;
 use Composer\Downloader\TransportException;
 use Composer\Factory;
 use Composer\IO\IOInterface;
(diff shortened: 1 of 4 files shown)


src/Downloader.php
```diff
@@ -64,19 +67,79 @@ public function allowContrib(bool $allow): void
     }
 
     /**
+     * Downloads recipes.
+     *
+     * @param OperationInterface[] $operations
+     */
+    public function getRecipes(array $operations): array
+    {
+        $max = 1000;
```
fabpot: Not sure what could be the optimal value here?

PR #114, opened 2017-06-24
Fix missed argument in BundlesConfigurator::unconfigure()

shude:
[Configurator\BundlesConfigurator] fix in method unconfigure() 

Missed second argument for method $this->parse is added.


fabpot:
Thank you @shude.

PR #123, opened 2017-07-04
Alphabetize bundles by key

ericpoe:
This helps organize bundles by vendor and bundle family. This is for UX and has no bearing on the performance of the application.

As an example, after setting up an initial skeleton project, I wanted to install api-platform, but since the default "minimum-stability" of skeleton is "stable" instead of "dev," a quick `composer req api` fails. So, I installed all of the libraries and bundles in https://github.com/api-platform/api-pack/blob/master/composer.json, then I ran `composer require admin` for added complexity. I took a look at my `bundles.php` and saw this:

```php
return [
    'Symfony\Bundle\FrameworkBundle\FrameworkBundle' => ['all' => true],
    'Doctrine\Bundle\DoctrineCacheBundle\DoctrineCacheBundle' => ['all' => true],
    'Doctrine\Bundle\DoctrineBundle\DoctrineBundle' => ['all' => true],
    'Symfony\Bundle\SecurityBundle\SecurityBundle' => ['all' => true],
    'Symfony\Bundle\TwigBundle\TwigBundle' => ['all' => true],
    'Sensio\Bundle\FrameworkExtraBundle\SensioFrameworkExtraBundle' => ['all' => true],
    'JavierEguiluz\Bundle\EasyAdminBundle\EasyAdminBundle' => ['all' => true],
];
```

The bundles appear to be randomly sorted and I have a harder time quickly grokking which bundles are installed. Wouldn't it be nice to see all of these bundles arranged alphabetically so that I can see them grouped by vendor and maybe library family? Perhaps, something like this:

```php
return [
    'Doctrine\Bundle\DoctrineBundle\DoctrineBundle' => ['all' => true],
    'Doctrine\Bundle\DoctrineCacheBundle\DoctrineCacheBundle' => ['all' => true],
    'JavierEguiluz\Bundle\EasyAdminBundle\EasyAdminBundle' => ['all' => true],
    'Sensio\Bundle\FrameworkExtraBundle\SensioFrameworkExtraBundle' => ['all' => true],
    'Symfony\Bundle\FrameworkBundle\FrameworkBundle' => ['all' => true],
    'Symfony\Bundle\SecurityBundle\SecurityBundle' => ['all' => true],
    'Symfony\Bundle\TwigBundle\TwigBundle' => ['all' => true],
];
```
As can be seen above, I can quickly tell that I have 2 Doctrine bundles, 3 Symfony bundles, a JavierEguiluz bundle, and a Sensio bundle. In the first example, I cannot quickly discern which bundles I have.

Imagine what this will look like once more and more bundles are added!

### DIFF
--- a/src/Configurator/BundlesConfigurator.php
+++ b/src/Configurator/BundlesConfigurator.php
@@ -77,6 +77,7 @@ class BundlesConfigurator extends AbstractConfigurator
 
     private function dump(string $file, iterable $bundles): void
     {
+        ksort($bundles);
         $contents = '<?php'.PHP_EOL.PHP_EOL.'return ['.PHP_EOL;
         foreach ($bundles as $class => $envs) {
             $contents .= "    '$class' => [";
--- a/tests/Configurator/BundlesConfiguratorTest.php
+++ b/tests/Configurator/BundlesConfiguratorTest.php
@@ -33,13 +33,15 @@ class BundlesConfiguratorTest extends TestCase
         $configurator->configure($recipe, [
             'FooBundle' => ['dev', 'test'],
             'Symfony\Bundle\FrameworkBundle\FrameworkBundle' => ['all'],
+            'Symfony\Bundle\DebugBundle\DebugBundle' => ['all'],
         ]);
         $this->assertEquals(<<<EOF
 <?php
 
 return [
-    'Symfony\Bundle\FrameworkBundle\FrameworkBundle' => ['all' => true],
     'FooBundle' => ['dev' => true, 'test' => true],
+    'Symfony\Bundle\DebugBundle\DebugBundle' => ['all' => true],
+    'Symfony\Bundle\FrameworkBundle\FrameworkBundle' => ['all' => true],
 ];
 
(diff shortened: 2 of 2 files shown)


fabpot:
I'm mostly 👎 , but in any case, frameworkbundle should stay first.

ericpoe:
For my education, why should frameworkbundle stay first? Composer is already alphabetizing all requirements, why not `$bundles`?

Pierstoval:
Composer requirements and bundles are different. Requirements are just here to define project's dependencies, and they're managed by Composer, the order doesn't matter at all.

With bundles, it's different. The kernel retrieves the list of bundles, and then `Kernel::boot()` will initialize bundles with a simple `foreach` loop, therefore processing them one after the other, like, procedurally. Same for bundle extensions and compiler passes. So to avoid any strange complication, it's better to keep FrameworkBundle at the top, and let the user change the order manually IMO.

ericpoe:
@Pierstoval I appreciate the excellent answer to my poorly worded question.

PR #126, opened 2017-07-08
Use \n instead of PHP_EOL

dunglas:
Using `PHP_EOL` may cause issues when some developers use Windows while others use UNIX. I propose to follow the UNIX convention.

Still a WIP.

### DIFF
--- a/src/Configurator/AbstractConfigurator.php
+++ b/src/Configurator/AbstractConfigurator.php
@@ -54,6 +54,6 @@ abstract class AbstractConfigurator
 
     protected function markData(Recipe $recipe, $data)
     {
-        return PHP_EOL.sprintf('###> %s ###%s%s%s###< %s ###%s', $recipe->getName(), PHP_EOL, rtrim($data, PHP_EOL), PHP_EOL, $recipe->getName(), PHP_EOL);
+        return "\n".sprintf('###> %s ###%s%s%s###< %s ###%s', $recipe->getName(), "\n", rtrim($data, "\r\n"), "\n", $recipe->getName(), "\n");
     }
 }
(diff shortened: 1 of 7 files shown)


javiereguiluz:
In my opinion this proposal is technically wrong. `PHP_EOL` is the only correct value for every user (Windows, Linux, macOS). The case you mentioned (a team where some people use Windows and others use Linux/macOS) is and edged-case that I don't think we should solve ourselves (that team should be careful when configuring their editors ... not only for Flex, but for Symfony, PHP and any other file types).

dunglas:
> that team should be careful when configuring their editors

It's exactly what this PR fixes. The generated files are not created through editors but created by Flex.

lyrixx:
I agree with @dunglas here.
More over, in [my git config](https://github.com/lyrixx/dotfiles/blob/master/.gitconfig#L41), git automatically transform all `\r\n` to `\n`.
So using only `\n` will avoid useless commit ;)

javiereguiluz:
@lyrixx your Git configuration is correct ... but we cannot assume the same for other developers.

In any case, to me this PR is obviously wrong ... but for you is obviously right ... so someone must be incredibly wrong ... and I guess it's me 😞

stof:
👍 about using Unix line ending when generating files, as these files are meant to be committed.

stof:
Btw, a .env file with Windows line endings cannot be sourced anymore when using bash, as bash supports only LF line-endings in bash files (even when compiling bash for windows)

Pierstoval:
For the makefile, I could not make it work on Windows with `\n` nor `\r\n`, because it uses unix-based commands (like `test` for example), so needs Linux (WSL or at least unix commands in Path).

I'm 👍 for this change for consistency anyway, especially for `.env` file

fabpot:
Thank you @dunglas.

PR #127, opened 2017-07-11
Remove unused var

dunglas:


src/Downloader.php
```diff
@@ -79,7 +79,6 @@ public function setRepositories(array $repos)
      */
     public function getRecipes(array $operations): array
     {
-        $max = 1000;
```
fabpot: I would prefer to keep the var and replace the hardcoded `1000` in the code instead.

Pierstoval: But this var is unused 🤔

chalasr: > replace the hardcoded 1000 in the code instead

by `$max` so that it would be (used)

stof: We could even use a private constant rather than a var

dunglas: Switched to a private const

fabpot:
Thank you @dunglas.

Issue #54, opened 2017-04-28
Docker support


dunglas:
Hi,

I've created an alternative Symfony skeleton with Docker support: https://github.com/dunglas/symfony-docker-skeleton
It's basically the API Platform Docker setup tuned to support Flex.

It works, but some minimal modifications can be done to some recipes to ease the install (expect some PRs soon).

Main differences with the standard skeleton are:

* It doesn't use the Dotenv component (Docker has a builtin support of `.env` files)
* In `dev` mode, the front controller can be accessed from the host

It's probably possible to go one step further and to automatically add required services with a `Configurator` (e.g: add a Redis image to `docker-compose.yml` when installing `predis/predis`).

Is it something you may include into Flex?

dkarlovi:
I have a rather complex Symfony app (WIP, also ApiPlatform) in Docker: development with Docker Compose, deployment to Openshift / Kubernetes, support for environments, a lot of tweaks per each image for example, my app image allows for XDebug on development but it's disabled on production (using the exact same image, different runtime environment) or uses the same image for web app server (PHP-FPM) and async CLI-workers (Console + [Enqueue bundle](https://github.com/php-enqueue/enqueue-bundle)) which, if in development, monitor the `src/` folder for changes and auto-restart the workers.

From that experience, it shouldn't be too hard to get a pretty good all-around config, but the problem would be how to allow for extendability as you currently can't extend Dockerfiles (you can extend Docker Compose file, that's how I get the per-environment settings), it would be a run-once deal.

teohhanhui:
I don't believe Symfony Flex should create services in the Docker Compose file (or its equivalent in other container orchestration systems).

But we really need a "developer story" to understand what's the experience like using Symfony Flex with Docker. I believe that for Symfony Flex to be successful, it really needs to treat Docker as a first-class citizen (and this is the perfect opportunity to ensure that is the case). I've just started trying it out...

andrerom:
First, @dkarlovi nice work on https://github.com/dkarlovi/symfony-flex-docker

As for extensibility there are a few things we can potentially use here:
- Recipes can be allowed to specify either php extensions to enable _(among the once in php source)_, or to install via pecl
  - _(One blocker here, pecl is not able to pick latests version of for instance memcached on PHP 5.6, you'll have to specify version for that. Not saying we care about php 5.6, but the same issue will come back at any point there is sufficient breaks in php again for extensions)_
[...]

PR #128, opened 2017-07-12
Docker support (Docker Compose and Dockerfile)

dunglas:
/!\ This piece of software is experimental and may break your app. Don't use it without a proper versioning strategy.

This is a part of my global plan to add a first-class Docker support in Symfony. It requires https://github.com/dunglas/symfony-docker-skeleton/pull/5.

This PR allows recipes to updates `docker-compose.yaml` (to add services and volumes for instance) and the `Dockerfile` (to add PHP extensions for instance).

I'll blog about this when I'll be back from vacation.

### DIFF
--- a/src/Configurator/DockerComposeConfigurator.php
+++ b/src/Configurator/DockerComposeConfigurator.php
@@ -1,0 +1,128 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Flex\Configurator;
+
+use Symfony\Flex\Lock;
+use Symfony\Flex\Recipe;
+
+/**
+ * Adds services and volumes to docker-compose.yml file.
+ *
+ * @author Kévin Dunglas <dunglas@gmail.com>
+ */
+class DockerComposeConfigurator extends AbstractConfigurator
+{
+    public function configure(Recipe $recipe, $config, Lock $lock, array $options = [])
+    {
+        $installDocker = $this->composer->getPackage()->getExtra()['symfony']['docker'] ?? false;
+        if (!$installDocker) {
+            return;
+        }
+
+        $rootDir = $this->options->get('root-dir');
+        if (
+            (
+                !file_exists($dockerComposeFile = $rootDir.'/docker-compose.yml') &&
+                !file_exists($dockerComposeFile = $rootDir.'/docker-compose.yaml')
+            ) || $this->isFileMarked($recipe, $dockerComposeFile)
+        ) {
+            return;
+        }
+
+        $this->write('Adding Docker Compose entries');
+
+        $offset = 8;
+        $node = null;
+        $endAt = [];
+        $lines = [];
+        foreach (file($dockerComposeFile) as $i => $line) {
+            $lines[] = $line;
+            $ltrimedLine = ltrim($line, ' ');
+
+            // Skip blank lines and comments
+            if (('' !== $ltrimedLine && 0 === strpos($ltrimedLine, '#')) || '' === trim($line)) {
+                continue;
+            }
+
+            // Extract Docker Compose keys (usually "services" and "volumes")
+            if (!preg_match('/^[\'"]?([a-zA-Z0-9]+)[\'"]?:\s*$/', $line, $matches)) {
+                // Detect indentation to use
+                $offestLine = \strlen($line) - \strlen($ltrimedLine);
+                if ($offset > $offestLine && 0 !== $offestLine) {
+                    $offset = $offestLine;
+                }
+                continue;
+            }
+
+            // Keep end in memory (check break line on previous line)
+            $endAt[$node] = '' !== trim($lines[$i - 1]) ? $i : $i - 1;
+            $node = $matches[1];
+        }
+        $endAt[$node] = \count($lines) + 1;
+
+        foreach ($config as $key => $value) {
+            if (isset($endAt[$key])) {
+                array_splice($lines, $endAt[$key], 0, $this->markData($recipe, $this->parse(1, $offset, $value)));
+                continue;
+            }
+
+            $lines[] = sprintf("\n%s:", $key);
+            $lines[] = $this->markData($recipe, $this->parse(1, $offset, $value));
+        }
+
+        file_put_contents($dockerComposeFile, implode('', $lines));
+    }
+
+    public function unconfigure(Recipe $recipe, $config, Lock $lock)
+    {
+        $rootDir = $this->options->get('root-dir');
+        if (!file_exists($dockerCompose = $rootDir.'/docker-compose.yml') &&
+            !file_exists($dockerCompose = $rootDir.'/docker-compose.yaml')
+        ) {
+            return;
+        }
+
+        $name = $recipe->getName();
+        // Remove recipe and add break line
+        $contents = preg_replace(sprintf('{%s+###> %s ###.*?###< %s ###%s+}s', "\n", $name, $name, "\n"), PHP_EOL.PHP_EOL, file_get_contents($dockerCompose), -1, $count);
+        if (!$count) {
+            return;
+        }
+
+        foreach ($config as $key => $value) {
+            if (0 === preg_match(sprintf('{^%s:[ \t\r\n]*([ \t]+\w)}m', $key), $contents, $matches)) {
+                $contents = preg_replace(sprintf('{\n?^%s:[ \t\r\n]*}sm', $key), '', $contents, -1, $count);
+            }
+        }
+
+        $this->write(sprintf('Removing Docker Compose entries from %s', $dockerCompose));
+        file_put_contents($dockerCompose, ltrim($contents, "\n"));
+    }
+
+    private function parse($level, $indent, $services): string
+    {
+        $line = '';
+        foreach ($services as $key => $value) {
+            $line .= str_repeat(' ', $indent * $level);
+            if (!\is_array($value)) {
+                if (\is_string($key)) {
+                    $line .= sprintf('%s:', $key);
+                }
+                $line .= sprintf("%s\n", $value);
+                continue;
+            }
+            $line .= sprintf("%s:\n", $key).$this->parse($level + 1, $indent, $value);
+        }
+
+        return $line;
+    }
+}
--- a/src/Configurator/DockerfileConfigurator.php
+++ b/src/Configurator/DockerfileConfigurator.php
@@ -1,0 +1,66 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Flex\Configurator;
+
+use Symfony\Flex\Lock;
+use Symfony\Flex\Recipe;
+
+/**
+ * Adds commands to a Dockerfile.
+ *
+ * @author Kévin Dunglas <dunglas@gmail.com>
+ */
+class DockerfileConfigurator extends AbstractConfigurator
+{
+    public function configure(Recipe $recipe, $config, Lock $lock, array $options = [])
+    {
+        $installDocker = $this->composer->getPackage()->getExtra()['symfony']['docker'] ?? false;
+        if (!$installDocker) {
+            return;
+        }
+
+        $dockerfile = $this->options->get('root-dir').'/Dockerfile';
+        if (!file_exists($dockerfile) || $this->isFileMarked($recipe, $dockerfile)) {
+            return;
+        }
+
+        $this->write('Adding Dockerfile entries');
+
+        $lines = [];
+        foreach (file($dockerfile) as $line) {
+            $lines[] = $line;
+            if (!preg_match('/^###> recipes ###$/', $line)) {
+                continue;
+            }
+
+            $lines[] = ltrim($this->markData($recipe, implode("\n", $config)), "\n");
+        }
+
+        file_put_contents($dockerfile, implode('', $lines));
+    }
+
+    public function unconfigure(Recipe $recipe, $config, Lock $lock)
+    {
+        if (!file_exists($dockerfile = $this->options->get('root-dir').'/Dockerfile')) {
+            return;
+        }
+
+        $name = $recipe->getName();
+        $contents = preg_replace(sprintf('{%s+###> %s ###.*?###< %s ###%s+}s', "\n", $name, $name, "\n"), "\n", file_get_contents($dockerfile), -1, $count);
+        if (!$count) {
+            return;
+        }
+
+        $this->write('Removing Dockerfile entries');
+        file_put_contents($dockerfile, ltrim($contents, "\n"));
+    }
+}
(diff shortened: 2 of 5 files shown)


francoispluchino:
With the proposal #173, this PR would become useless, but you could add this feature to Flex by creating a Composer plugin requiring Flex in her dependencies.

stof:
Well, creating a plugin as described in #173 would not help much AFAIU, as it would also require extending the flex server to support more fields in the recipes (your proposal does not only involve changing Flex, but also the recipes)

francoispluchino:
It is not necessary to add new server-side fields, if the plugin is able to retrieve information elsewhere, for example: in the composer extra section, or in local file.

dunglas:
For this PR for instance, it must retrieve fields from the recipe, so from the server (because the server parse and modify the recipe).

dunglas:
But regarding this PR, it can be made more generic as `Dockerfile`s are just text files, and `docker-compose.yaml` YAML files. I will probably rename the configurator, but I would like to test it as is before (it requires some tiny server-side modifications).

francoispluchino:
@dunglas Regarding the required new fields, see my [comment](https://github.com/symfony/flex/issues/173#issuecomment-334112575).

maxhelias:
#54

maxhelias:
#54

maxhelias:
#54

maxhelias:
#54

dunglas:
@fabpot has done the required changes on the Flex servicer side. I need to test if it works now. I'll do it as soon as possible.

maxhelias:
Do not forget to resolve conflit over the doctrine recipe [symfony/recipes#113](https://github.com/symfony/recipes/pull/113)

src/Configurator/DockerfileConfigurator.php
```diff
@@ -0,0 +1,60 @@
...
+        if (!file_exists($dockerfile) || $this->isFileMarked($recipe, $dockerfile)) {
+            return;
+        }
+
+        $this->write('Adding Dockerfile entries');
+
+        $lines = [];
+        foreach (file($dockerfile) as $line) {
+            $lines[] = $line;
+            if ("###> recipes ###\n" !== $line) {
```
covex-nn: Sometimes this will be always true under Windows, because `$line` can have `\r\n` at the end

covex-nn:
I can skip `DockerfileConfigurator` by removing `###> recipes ###` from my Dockerfile. But it won't be possible to skip `DockerComposerConfigurator` - it will add not validated `$config` to my `docker-compose.yaml` anyway =(

maxhelias:
Would not it be relevant to disable docker support for people who do not use it so that there is no loss of performance in case they already have a `docker-composer.yaml` file at the root of their project ?

I see this with an options in the `composer.json` but why not in another way if someone has a better idea

maxhelias:
Also, it would be necessary to be able to manage more provider to be able to use either mysql, postgres, mongodb ...

src/Configurator/DockerComposeConfigurator.php
```diff
@@ -0,0 +1,72 @@
...
+        if (!file_exists($dockerCompose) || $this->isFileMarked($recipe, $dockerCompose)) {
+            return;
+        }
+
+        $this->write('Adding Docker Compose entries');
+
+        $lines = [];
+        foreach (file($dockerCompose) as $line) {
+            $lines[] = $line;
+            if (!preg_match('/^[\'"]?([a-zA-Z0-9]+)[\'"]?:\s*$/', $line, $matches) || !isset($config[$matches[1]])) {
```
covex-nn: If i will add a comment at the end of a line with `"services":` in my `docker-compose.yml`, there will be two `services` sections there after applying `$config`

src/Configurator/DockerfileConfigurator.php
dunglas: No because this line is generated by Flex, and Flex always uses Unix EOLs.

dunglas:
This one is ready to be merged.

maxhelias:
If anyone wants to try, just following these instruction

## Clone : 
```
git clone https://github.com/dunglas/symfony-docker
```
Run the docker :
```
docker-compose up -d
```

## Switch branch :
Define the fork repo on repositories section like this :
```
"repositories": [
 {
   "type": "vcs",
   "url": "https://github.com/dunglas/flex"
 }
],
```
Define the branch `dev-docker` for `symfony/flex` dependencies on `composer.json`

## Test a recipe : 
```
SYMFONY_ENDPOINT=https://symfony.sh/r/github.com/symfony/recipes/{PR} composer req {recipe}
```
### Recipe available :
Mercure : https://github.com/symfony/recipes/pull/583
Messenger : https://github.com/symfony/recipes/pull/413
Doctrine : https://github.com/symfony/recipes/pull/113

teohhanhui:
Would love to review this before it gets merged. Hope we're not in a hurry, after 2 years.

dunglas:
Hi there, can we merge this one?

src/Configurator/DockerComposeConfigurator.php
```diff
@@ -0,0 +1,120 @@
...
+class DockerComposeConfigurator extends AbstractConfigurator
+{
+    public function configure(Recipe $recipe, $config, Lock $lock, array $options = [])
+    {
+        $installDocker = $this->composer->getPackage()->getExtra()['symfony']['docker'] ?? false;
+        if (!$installDocker) {
+            return;
+        }
+
+        $dockerCompose = $this->options->get('root-dir').'/docker-compose.yaml';
```
fabpot: We need to support `docker-compose.yml` as well, right?

fabpot: We could also support the `COMPOSE_PROJECT_NAME` env var: https://docs.docker.com/compose/reference/envvars/#compose_project_name

dunglas: How could we use `COMPOSE_PROJECT_NAME`?

dunglas: To distinguish containers of multiple projects running at the same time? You would set this env var to the name defined in `composer.json`? Actually it must be set before running `docker-compose`, so it should be in Symfony CLI I guess.

dunglas: `docker-compose.yml` support added

teohhanhui: Docker Compose always reads `.env`. Special env vars like `COMPOSE_PROJECT_NAME` can be set in there. But also, any env var defined in `.env` can be used for interpolation in `docker-compose.yml`

src/Configurator/DockerComposeConfigurator.php
```diff
@@ -0,0 +1,128 @@
...
+            if (0 === preg_match(sprintf('{^%s:[ \t\r\n]*([ \t]+\w)}m', $key), $contents, $matches)) {
+                $contents = preg_replace(sprintf('{\n?^%s:[ \t\r\n]*}sm', $key), '', $contents, -1, $count);
+            }
+        }
+
+        $this->write(sprintf('Removing Docker Compose entries from %s', $dockerCompose));
+        file_put_contents($dockerCompose, ltrim($contents, "\n"));
+    }
+
+    private function parse($level, $indente, $services): string
```
fabpot: typo? should be `$indent`?

fabpot:
@dunglas Can you cleanup the commit history? That's the last thing to do before merging!

src/Configurator/DockerComposeConfigurator.php
```diff
@@ -0,0 +1,128 @@
...
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Flex\Configurator;
+
+use Symfony\Flex\Lock;
+use Symfony\Flex\Recipe;
+
+/**
+ * Adds services and volumes to docker-compose.yaml file.
```
teohhanhui: I know I've made this point before, but upstream documentation refers to `docker-compose.yml` everywhere. Sure, `.yaml` works as well, but `docker-compose.yml` is what users would be looking for when using upstream docs. I don't see the need to create additional friction.

src/Configurator/DockerComposeConfigurator.php
```diff
@@ -0,0 +1,128 @@
...
+        foreach (file($dockerCompose) as $i => $line) {
+            $lines[] = $line;
+            $ltrimedLine = ltrim($line, ' ');
+
+            // Skip blank lines and comments
+            if (('' !== $ltrimedLine && 0 === strpos($ltrimedLine, '#')) || '' === trim($line)) {
+                continue;
+            }
+
+            // Extract Docker Compose keys (usually "services" and "volumes")
```
teohhanhui: Shouldn't it use the Symfony YAML component to properly parse the YAML file? Or is it really undesirable to add an additional dependency?

src/Configurator/DockerComposeConfigurator.php
```diff
@@ -0,0 +1,128 @@
...
+    {
+        $installDocker = $this->composer->getPackage()->getExtra()['symfony']['docker'] ?? false;
+        if (!$installDocker) {
+            return;
+        }
+
+        $rootDir = $this->options->get('root-dir');
+        if (
+            (
+                !file_exists($dockerCompose = $rootDir.'/docker-compose.yml') &&
```
teohhanhui: `$dockerCompose` -> `$dockerComposeFile`?

src/Configurator/DockerfileConfigurator.php
teohhanhui: > No because this line is generated by Flex, and Flex always uses Unix EOLs.

`git` (or something else) might do EOL conversion... So that assumption is dangerous.

teohhanhui: Maybe use the same markers as Docker official images?

https://github.com/docker-library/php/blob/04e67dc6bb76783d2d1b64513b8f42961e578645/Dockerfile-debian.template#L115

tests/Configurator/DockerComposeConfiguratorTest.php
```diff
@@ -0,0 +1,325 @@
...
+class DockerComposeConfiguratorTest extends TestCase
+{
+    const ORIGINAL_CONTENT = <<<'YAML'
+version: '3.4'
+
+services:
+  app:
+    build:
+      context: .
+      target: symfony_docker_php
```
teohhanhui: The `docker` part of the name is really redundant... :smile:

tests/Configurator/DockerComposeConfiguratorTest.php
```diff
@@ -0,0 +1,325 @@
...
+      args:
+        SYMFONY_VERSION: ${SYMFONY_VERSION:-}
+        STABILITY: ${STABILITY:-stable}
+    volumes:
+      # Comment out the next line in production
+      - ./:/srv/app:rw,cached
+      # If you develop on Linux, comment out the following volumes to just use bind-mounted project directory from host
```
teohhanhui: Can be simplified.

https://github.com/api-platform/api-platform/blob/6a2ff40552d7d7b008c698cd120eea7f01fb1247/docker-compose.yml#L22-L23

Something like `var/jwt` should not be a problem, as a more specific path would take precedence. (But we need to verify that `setfacl` works...)

tests/Configurator/DockerfileConfiguratorTest.php
```diff
@@ -0,0 +1,187 @@
...
+RUN rm -Rf docker/
+
+# Download the Symfony skeleton
+ENV SKELETON_COMPOSER_JSON https://raw.githubusercontent.com/symfony/skeleton/v3.3.2/composer.json
+RUN [ -f composer.json ] || php -r "copy('$SKELETON_COMPOSER_JSON', 'composer.json');"
+
+RUN mkdir -p var/cache var/logs var/sessions \
+    && composer install --prefer-dist --no-dev --no-progress --no-suggest --optimize-autoloader --classmap-authoritative --no-interaction \
+	&& composer clear-cache \
+# Permissions hack because setfacl does not work on Mac and Windows
```
teohhanhui: This comment is wrong. As you are aware, we use `setfacl` just fine in api-platform/api-platform

Perhaps this needs syncing with the progress we've made there?

tests/Configurator/DockerfileConfiguratorTest.php
```diff
@@ -0,0 +1,187 @@
...
+RUN chmod +x /usr/local/bin/docker-app-install-composer
+
+RUN set -xe \
+	&& docker-app-install-composer \
+	&& mv composer.phar /usr/local/bin/composer
+
+# https://getcomposer.org/doc/03-cli.md#composer-allow-superuser
+ENV COMPOSER_ALLOW_SUPERUSER 1
+
+RUN composer global require "hirak/prestissimo:^0.3" --prefer-dist --no-progress --no-suggest --optimize-autoloader --classmap-authoritative \
```
teohhanhui: Shouldn't this be installing `symfony/flex` instead? Flexception... :rofl:

src/Configurator/DockerComposeConfigurator.php
dunglas: AFAIK, it's not desirable to add such dependency in a Composer plugin.

tests/Configurator/DockerComposeConfiguratorTest.php
dunglas: It's just a fixture file, its content doesn't matter.

tests/Configurator/DockerfileConfiguratorTest.php
dunglas: It's just a fixture file, its content doesn't matter.

tests/Configurator/DockerComposeConfiguratorTest.php
dunglas: It's just a fixture file, its content doesn't matter.

src/Configurator/DockerComposeConfigurator.php
lyrixx: But we could use it if it's available?!

tests/Configurator/DockerComposeConfiguratorTest.php
teohhanhui: So, where will the actual file live? In symfony/recipes? We should fix it there then...

teohhanhui: Or is it the plan that https://github.com/dunglas/symfony-docker will be moved under the symfony org?

src/Configurator/DockerComposeConfigurator.php
dunglas: Having different behaviors regarding the installed dependencies will be a pain to debug, isn't it? Maybe could we use @theofidry's PHP Scoper to ship our custom version of the component, but it could we handle that in another PR? Existing YAML manipulation is already done this way, if we change, we must change everywhere.

dunglas:
@fabpot last comments fixed, and history cleaned.

@teohhanhui 

> git (or something else) might do EOL conversion... So that assumption is dangerous.

Makes sense, changed to use a regex then.

> Maybe use the same markers as Docker official images?

Flex uses the same markers everywhere. I'm not sure if it's a good idea to change just for Docker.

teohhanhui:
> Flex uses the same markers everywhere. I'm not sure if it's a good idea to change just for Docker.

Indeed it's not important, as long as it's appropriate for the target file's syntax. :smile:

fabpot:
Thank you @dunglas.

Plopix:
question here regarding Docker Integration: https://github.com/symfony/symfony/issues/38259

Issue #116, opened 2017-06-25
Resolve directory paths when displaying messages


javiereguiluz:
As you can see here, some messages in Flex output say `%BIN_DIR%` and `%WEB_DIR%` instead of showing the real path to those dirs. Is this on purpose? I find it a bit confusing:

![unresolved-dir-paths](https://user-images.githubusercontent.com/73419/27518410-d3a9a2a4-59df-11e7-851c-e8366a969fe4.png)

fabpot:
That's on purpose, but expanding those directories is also possible. I don't have a strong opinion on this. Maybe expanding is a bit better. PR welcome.

fabpot:
Anyone willing to work on this one?

ricardocino:
If nobody else before that, willing to have a look after work today / tomorrow

PR #131, opened 2017-07-22
Display folder name in console output instead of variable.

ricardocino:
Fix for requested change in #116 , in console output it now shows the folder instead of %WEB_DIR% etc.

290a3d9f701907e466358ae3c4af493a

### DIFF
--- a/src/ScriptExecutor.php
+++ b/src/ScriptExecutor.php
@@ -43,7 +43,8 @@ class ScriptExecutor
      */
     public function execute(string $type, string $cmd): void
     {
-        if (null === $expandedCmd = $this->expandCmd($type, $this->options->expandTargetDir($cmd))) {
+        $translatedCmd = $this->options->expandTargetDir($cmd);
+        if (null === $expandedCmd = $this->expandCmd($type, $translatedCmd)) {
             return;
         }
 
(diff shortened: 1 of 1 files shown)


src/ScriptExecutor.php
```diff
@@ -43,7 +43,8 @@ public function __construct(Composer $composer, IOInterface $io, Options $option
      */
     public function execute(string $type, string $cmd): void
     {
-        if (null === $expandedCmd = $this->expandCmd($type, $this->options->expandTargetDir($cmd))) {
+        $translatedCmd = $this->options->expandTargetDir($cmd);
```
javiereguiluz: @ricardocino thanks for providing this feature. My only (minor) concern is the `$translatedCmd` variable name. "Translated" has a strong meaning unrelated to this feature, so it may be confusing. Sadly I cannot propose a better name ($parsedCm, $resolvedCmd, $realCmd, etc. don't look good either to me).

ricardocino: Agreed, happy to change to another variable name if someone has a better idea.

fabpot: `$expandedCmd` sounds the best one to me as the method is name `expandCmd`.

hhamon: `$absoluteCmd`, `$fullCmd` maybe

ricardocino: @fabpot The method name is actually called "expandTargetDir", the method "expandCmd" is a different method which is actually being stored in $expandedCmd.

fabpot: Right, but still, `$expandedCmd` looks good to me.

fabpot:
Thank you @ricardocino.

PR #139, opened 2017-08-22
do not resolve php or extensions

kriswallsmith:
The Composer plugin is preventing the following commands from executing successfully:

```
composer require php
composer require ext-mongodb # (etc)
```

### DIFF
--- a/src/PackageResolver.php
+++ b/src/PackageResolver.php
@@ -12,6 +12,7 @@
 namespace Symfony\Flex;
 
 use Composer\Package\Version\VersionParser;
+use Composer\Repository\PlatformRepository;
 
 /**
  * @author Fabien Potencier <fabien@symfony.com>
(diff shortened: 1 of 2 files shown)


src/PackageResolver.php
```diff
@@ -46,7 +46,7 @@ public function resolve(iterable $arguments = []): iterable
         // second pass to resolve package names
         $packages = [];
         foreach ($explodedArguments as $i => $argument) {
-            if (false === strpos($argument, '/')) {
+            if ('php' !== $argument && 0 !== strpos($argument, 'ext-') && false === strpos($argument, '/')) {
```
jvasseur: We should probably use [PLATFORM_PACKAGE_REGEX](https://github.com/composer/composer/blob/master/src/Composer/Repository/PlatformRepository.php#L27) to handle all cases and stay in sync to possible future platform packages.

fabpot: 👍

kriswallsmith: Thanks @jvasseur! I just pushed a fix.

fabpot:
Thank you @kriswallsmith.

PR #143, opened 2017-08-30
BundlesConfigurator: use ::class

gharlan:
Before:

screenshot 2017-08-30 23 22 19

After:

screenshot 2017-08-30 23 23 53

In my opinion the second one looks better.


jrobeson:
I did the same in my bundles.php file

ro0NL:
What about alpha order :) would be nice.

fabpot:
Thank you @gharlan.

PR #145, opened 2017-09-08
Support for composer update nothing

simPod:
Command `composer update nothing` was giving 
```
[UnexpectedValueException]
 "nothing" is not a valid alias.
```

This fixes it. Hope I have put it in the right place.


stof:
just use `composer update --lock` instead of the old "hack" of using a package name which would never match anything.

stof:
`nothing` is not a special package in composer at all (you are making it special now).

`composer update yolo` or `composer update fake` or `composer update lock` may have been used exactly the same than `composer update nothing` to emulate the `--lock` flag before it was implemented.

simPod:
Just googled It, lol. Thanks and nevermind :)

PR #148, opened 2017-09-11
Add support for env vars in phpunit.xml.*

fabpot:


tests/Configurator/EnvConfiguratorTest.php
```diff
@@ -0,0 +1,128 @@
...
+         colors="true"
+         bootstrap="vendor/autoload.php"
+>
+    <php>
+        <ini name="error_reporting" value="-1" />
+        <server name="KERNEL_CLASS" value="App\Kernel" />
+
+<!-- ###+ FooBundle ### -->
+<server name="APP_ENV" value="test" />
+<server name="APP_DEBUG" value="0" />
```
yceruto: Could be aligned with default indent 8, regarding to default file structure.

fabpot: done

src/Configurator/EnvConfigurator.php
```diff
@@ -47,15 +59,44 @@ public function configure(Recipe $recipe, $vars): void
...
+            $data = '';
+            foreach ($vars as $key => $value) {
+                if ('%generate(secret)%' === $value) {
+                    $value = bin2hex(random_bytes(16));
+                }
+                if ('#' === $key[0]) {
+                    $data .= '        <!-- '.$value." -->\n";
+                } else {
+                    $value = $this->options->expandTargetDir($value);
+                    $data .= "        <server name=\"$key\" value=\"$value\" />\n";
```
yceruto: According to https://github.com/symfony/recipes/pull/170 shouldn't be `<env ... />`?

fabpot: indeed, fixed. Thanks.

PR #153, opened 2017-09-13
Change the way we install contrib packages

fabpot:
When using Flex "extensively", installing contrib repos is a bit frustrating. If you opt-out (the default), you have a warning that the recipe won't be installed. The way to fix it is to 1/ remove the package 2/ enable contrib 3/ reinstall the package.

Instead, what about asking the question to the user and act accordingly (that means less to learn as enabling contrib is done automatically when selecting `[p]`)?

### DIFF
--- a/src/Flex.php
+++ b/src/Flex.php
@@ -62,7 +62,6 @@ class Flex implements PluginInterface, EventSubscriberInterface
         $this->downloader = new Downloader($composer, $io);
         $this->downloader->setFlexId($this->getFlexId());
         $extra = $composer->getPackage()->getExtra();
-        $this->downloader->allowContrib($extra['symfony']['allow-contrib'] ?? false);
         $this->downloader->setRepositories($extra['symfony']['repositories'] ?? []);
 
         // useful when symfony/flex is not yet installed (composer install without vendor/ for instance)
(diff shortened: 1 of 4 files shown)


weaverryan:
I like this approach. And we would still have several months to try it out. This is the first time that Flex is interactive, correct? That's the only weird part... but I still think it's best.

src/Flex.php
```diff
@@ -161,11 +160,44 @@ public function update(Event $event): void
...
         $this->io->writeError(sprintf('<info>Symfony operations: %d recipe%s</>', count($recipes), count($recipes) > 1 ? 's' : ''));
+        $installContribs = $this->composer->getPackage()->getExtra()['symfony']['allow-contrib'] ?? false;
         foreach ($recipes as $recipe) {
-            if ($recipe->isNotInstallable() && $recipe->isEmpty()) {
-                $this->io->writeError(sprintf('  - <warning>Ignoring</> recipe %s', $this->formatOrigin($recipe->getOrigin())));
-                $this->io->writeError('    Enable via composer config extra.symfony.allow-contrib true');
-                continue;
+            if ('install' === $recipe->getJob() && !$installContribs && $recipe->isContrib()) {
+                $this->io->writeError(sprintf('  - <warning>WARNING</> %s', $this->formatOrigin($recipe->getOrigin())));
+                $question = '    This package wants to update your project for better integration.
```
weaverryan: This doesn't make it look like it's clear that the issue is that this comes from the contrib repository. It's tough, we want to explain what's going on ("there are 2 repos, one is super protected, the other is not as protected, and this comes from that repo")... without all those words :).

fabpot: Indeed, the current sentence is far from perfect... I was expecting some help from you :)

src/Flex.php
```diff
@@ -161,11 +160,44 @@ public function update(Event $event): void
...
+    Allow?
+    [<comment>y</>] Yes
+    [<comment>n</>] No
+    [<comment>a</>] Yes for all packages, only for the current installation session
+    [<comment>p</>] Yes permanently, never ask again for this project
+    (defaults to <comment>n</>): ';
```
xabbuh: Should we print a different text if the command is run in non-interactive mode?

src/Flex.php
weaverryan: Indeed :). I don't remember if I was short on time... or just being lazy... Let's just pretend it was the first

>    The recipe for this package comes from the contrib repository, which is open to community contributions.
>    Allow this recipe?

Later, we could even add a short link to some docs (e.g. `http://symfony.com/contrib-recipes`)

fabpot: I've used:

```
The recipe for this package comes from the "contrib" repository, which is open to community contributions.
    Do you want to execute this recipe?
```

src/Flex.php
fabpot: The question is never displayed if the command is non-interactive.

fabpot:
This PR is ready

src/Flex.php
fabpot: I've made one small tweak. If non-interactive mode, it displays 'IGNORING' like before this PR and 'WARNING' in interactive mode. So, in non-interactive mode, the display is the same as before.

PR #156, opened 2017-09-15
Add help command and set it by default

Grafikart:
This is a very opinionated PR so be free to close it if you don't like it.

When `make` is used without argument it will print an help showing available targets for the current project (`make help` can also be used). The output is based on this [post](https://marmelab.com/blog/2016/02/29/auto-documented-makefile.html) but can be simplified.

csgrxlf

To use this feature "recipes" will have to comment public targets, for instance :

```makefile
cache-clear: ## Clear the cache
ifdef CONSOLE
	@$(CONSOLE) cache:clear --no-warmup
else
	@rm -rf var/cache/*
endif
```

Targets that are not commented are not visible (usefull to "hide" target used internaly like `serve_as_php` or `serve_as_symfony`)

### DIFF
--- a/src/Configurator/MakefileConfigurator.php
+++ b/src/Configurator/MakefileConfigurator.php
@@ -36,7 +36,7 @@ ifndef APP_ENV
 endif
 
 .DEFAULT_GOAL := help
-.PHONY : help
+.PHONY: help
 help:
 	@grep -E '^[a-zA-Z_-]+:.*?## .*$$' Makefile | sort | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[36m%-15s\033[0m %s\n", $$1, $$2}'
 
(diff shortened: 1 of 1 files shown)


sstok:
Target with underscores should be ignored as these are considered internal.

Grafikart:
If you don't comment the target it won't appear, but I fixed the regexp to ignore _

Nicofuma:
With `@grep -E '(^[a-zA-Z_-]+:.*?##.*$$)|(^##)' $(MAKEFILE_LIST) | awk 'BEGIN {FS = ":.*?## "}; {printf "\033[32m%-30s\033[0m %s\n", $$1, $$2}' | sed -e 's/\[32m##/[33m/'` you can add sections and you will also have colors close to the bin/console ones.

Ex with sections:
```
#
## Section 1
##---------------------------------------------------------------------------

target1:          ## target one

#
## Section 2
##---------------------------------------------------------------------------

target2:          ## target two
target2: target1
```

fabpot:
Thank you @Grafikart.

fabpot:
see symfony/recipes#207

PR #157, opened 2017-09-18
Do not rely on a hard coded config-dir in BundlesConfigurator

jakzal:
Currently, when `config-dir` is changed in `composer.json`, the `bundles.php` file is the only one generated in a hard coded location (`config/bundles.php`). Everything else is nicely put in the directory configured in `composer.json`. 

This is an attempt to fix the issue so that `config-dir` is respected by flex consistently.


fabpot:
Thank you @jakzal.

PR #162, opened 2017-09-23
Make Flex work on PHP 7.0

nicolas-grekas:
IMHO, enforcing 7.1 is only slowing down the adoption of Flex.
Unlike other existing projects that can tell "use the previous existing version",
we're starting with no community at all, and we want to spread adoption.
This is a strategic move to me.

Symfony 3.4 does work with 7.0, so PHP 7.0 users/companies that want to start new projects on it will have the choice to start with an already outdated standard edition, or nothing. This may hurt the Symfony project.

We have to provide a better alternative to them: Flex on PHP 7.0.
Here it is.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,10 +4,10 @@ sudo: false
 
 matrix:
   include:
-    - php: 7.1
+    - php: 7.0
     - php: 7.1
       env: COMPOSER_FLAGS='--prefer-lowest'
-    - php: nightly
+    - php: 7.2
   fast_finish: true
 
 cache:
--- a/composer.json
+++ b/composer.json
@@ -9,7 +9,7 @@
         }
     ],
     "require": {
-        "php": "^7.1",
+        "php": "^7.0",
         "composer-plugin-api": "^1.1"
     },
     "require-dev": {
--- a/src/Command/RemoveCommand.php
+++ b/src/Command/RemoveCommand.php
@@ -27,7 +27,7 @@ class RemoveCommand extends BaseRemoveCommand
         parent::__construct();
     }
 
-    protected function execute(InputInterface $input, OutputInterface $output): ?int
+    protected function execute(InputInterface $input, OutputInterface $output)
     {
         $input->setArgument('packages', $this->resolver->resolve($input->getArgument('packages')));
 
--- a/src/Command/RequireCommand.php
+++ b/src/Command/RequireCommand.php
@@ -27,7 +27,7 @@ class RequireCommand extends BaseRequireCommand
         parent::__construct();
     }
 
-    protected function execute(InputInterface $input, OutputInterface $output): ?int
+    protected function execute(InputInterface $input, OutputInterface $output)
     {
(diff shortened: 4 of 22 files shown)


B-Galati:
I think it's a good idea but @fabpot did a poll on Twitter on this and 7.1 won. So perhaps this change is not gonna be welcome by the community.
Here is the tweet -> https://twitter.com/fabpot/status/851558576770252800
Also, the active support of php 7.0 ends in December.

nicolas-grekas:
@B-Galati the pool was about Symfony 4, it doesn't apply to Flex IMHO. I see no reasons why it should.

B-Galati:
Good point ! You are entirely right 👍

stof:
Well, PHP 7.0 will reach end of active support on December 3rd.

Flex is expected to be released syable at the same time than Symfony 3.4 and 4.0, which means on November 30th.

Do we actually need to reintroduce support for PHP 7.0 for 3 days of active support ?

stof:
btw, if you want to add PHP 7.0 support, it does not only involves changing Flex itself, but also reviewing all recipes to ensure they are compatible with PHP 7.0 (hint: currently, they are not)

stof:
And currently, we don't have a way to make Flex select an older version of a recipe based on the targeted PHP version, so we would never be able to use PHP 7.1 features in recipe codes unless the targeted version of the composer package requires PHP 7.1+, meaning we would have to start a new version of the recipe for Symfony 4.0 packages.

And the first package recipe I know of requiring PHP 7.1 is `symfony/framework-bundle`

fabpot:
@nicolas-grekas That's for you :)

nicolas-grekas:
Where in FrameworkBundle? I parses all PHP files, they all are ok with 7.0 now.

nicolas-grekas:
We don't have a way to select a recipe based in a PHP version, but I'd expect a recipe to match the min version of the package version they configure. So that's not a real issue to me.

stof:
@nicolas-grekas whihc is not the case currently, as they were written with the assumption that Flex itself targets PHP 7.1, and so `max(PHP 7.1, min-package-target)` was used as recipe supported version. So **all** recipes must be reviewed.

nicolas-grekas:
No pb, there are only very few PHP files in the repos, I'll take care of it.

stof:
@nicolas-grekas don't forget the contrib ones

nicolas-grekas:
See https://github.com/symfony/recipes/pull/195

nicolas-grekas:
> PHP 7.0 will reach end of active support on December 3rd

PHP 7.0 will reach end of life on 3 Dec 2018, and even later if we consider Ubuntu Xenial, which will maintain it until April 2021. So definitely yes, we want to support 7.0.

fabpot:
@nicolas-grekas You should probably change the min version in symfony/skeleton as well, right?

fabpot:
Thank you @nicolas-grekas.

PR #169, opened 2017-09-29
Fix PHP 7.0 compatibility

yceruto:
Negative string offsets (`$string[-1]`) only work since `7.1`:
 * https://wiki.php.net/rfc/negative-string-offsets
 * https://3v4l.org/kOlne

Other solution could be `substr($source, -1)` if your prefer.


fabpot:
`substr($source, -1)` looks better to me.

xabbuh:
I would prefer `substr()` with a negative offset too.

yceruto:
Done.

fabpot:
Thank you @yceruto.

Issue #173, opened 2017-10-04
[Proposal] Allow to extend Flex


francoispluchino:
It would be very interesting to be able to extend the features of Flex with other plugin Composer. This would allow for example to add a new configuration specific to a project, or to set up recipes for private repositories (see #71).

You will find in this [comment](https://github.com/symfony/flex/issues/70#issuecomment-299915048) of the issue #70, the reasons and arguments to make Symfony Flex more flexible, in report with this proposal.

For this, it would be quick and simple to implement the Composer event dispatcher in Flex, enabling:

- dispatch the event `flex.fetch_recipe` before the construction of the new recipe (for each installed/updated package)
- dispatch the event `flex.init` (to replace the below components)
- to be able to replace the component `Configurator` (to extend features, like Composer)
- to be able to replace the component `Downloader` (to extend features, like Composer)

It would also be necessary to add:

- a method `Configurator::update($recipe)`, and use it in the switch case `install`/`update`/`uninstall` (because, the switch case `update` is empty currently).
- a method `Flex::setConfigurator($configurator)` to replace the configurator
- a method `Flex::setDownloader($downloader)` to replace the downloader

> **Note:** The event `flex.fetch_recipe` allows you to create an extended class of `Recipe`, and returns `null` to let Flex create a `Recipe` instance.

Are you interested by this feature? If so, I can quickly offer you a pull request.

stof:
What kind of extension point would you need ? Just adding a way to replace everything only opens the door for weird breakages (especially given that we are still making changes to Flex).

Regarding `flex.fetch_recipe`, this would not fit well in the current architecture which fetches recipes in batch.

stof:
For configurations specific to a project, I don't see the need to customize recipes for that: recipes will give you the initial configuration of the package in your project. But then, you are free to update the configuration as you want. These files become part of your project.

And regarding private repositories, the best way to support them would probably be to support multiple repositories in the plugin (currently, the repository is configurable but it replaces the existing one).
I would say that this makes sense in the roadmap, but once we settle on the Flex architecture, as this would require starting to provide BC on the protocol between the Flex client and the server. For now, we are still changing this part regularly to improve it before the Symfony 4 release. For instance, we changed the fetching of recipes to load them in a single HTTP request in June, and we change the way contrib recipes are handled 3 weeks ago. both of them would have been much harder if the communication with the server had to be covered by BC already.

francoispluchino:
[...]

Issue #71, opened 2017-05-07
[Proposal] Recipe for private repositories


francoispluchino:
Using a server (https://symfony.sh) to define the Recipes is an excellent initiative for open-source packages, allowing to create an automatic configuration for Symfony even if the author of the package didn't include / doesn't wish to include this configuration. It is a practice more and more used, in particular in the world of javascript, like [TypeScript type definitions](https://github.com/DefinitelyTyped/DefinitelyTyped) for example.

However, the repositories [symfony/recipes](https://github.com/symfony/recipes) and [symfony/recipes-contrib](https://github.com/symfony/recipes-contrib) are public, and require approval to add or modify their manifests and their configurations (manual and strict for `official`, semi-automatic if I understood correctly for `contrib`). This approach is therefore not suitable for private packages.

Questions are beginning to be asked about this on the Medium [Symfony 4: Contributing Recipes](https://medium.com/@fabpot/symfony-4-contributing-recipes-2f489dc6f645), With an approach to create a private server, And asking the question about opening the source code of symfony.sh to create this private server.

I'm not against this approach quite the opposite, it sounds like creating a Satis server for Composer. However, it would be simpler/quicker to add the manifest and configurations to a dedicated directory in the repository. Just like Composer, that we can add repositories that are not referenced in Packagist or a Satis private server by using the VCS Repository. But unlike the VCS Repositories of Composer, the performance are not impacted, quite the opposite (no HTTP request).

The path to the Recipe directory might will have a default value (`./recipe/` or other), with a possibility to change this value in the `composer.json` file with the `extra.symfony.recipe-path` key (or another name). If this directory exists in the repository with the `manifest.json` file, in this case, no request is done on the `symfony.sh` server.

This approach can be easily implemented given that the automatic configuration is performed just after the installation of the dependency and added just before the search of recipe on the symfony.sh server (in `Symfony\Flex\Flex::filterPackageNames()`.

**To summarize the workflow after the installation of package:**

1. Check the presence of the key `extra.symfony-recipe-path` in `composer.json` file of package
2. Get the real path of package recipe
3. Check if the `manifest.json` file is present and valid
4. If package recipe is valid, install the automatic configuration from package source code
5. Else, search recipe in symfony.sh server and install the automatic configuration from remote recipe if recipe is present on the server (the current behavior)

What do you think?

dkarlovi:
Check #70 for their reasons why not.

francoispluchino:
[...]

Issue #70, opened 2017-05-07
Allow Composer packages to ship their own recipes


dkarlovi:
If you're a package developer, you might want to opt-in to being Flex compatible. This way, you're shipping the recipe with the package, maintaining it there (each version might be different, but as they're together, they're in sync). Without this, you maintain your package in your own repo, but need to maintain a part of it somewhere else (in the recipes* repos). If you push a major release, you might have to wait for someone else to approve the PR, etc.

You could allow for packages to ship their own Symfony recipes, maybe directly inside `composer.json`? It could work like this:

1. developer Dave decides to create a new Symfony bundle
2. he develops it completely and adds `composer.json` to it
3. he wishes to be compatible with Symfony Flex too so he adds a block to `composer.json` with the recipe
4. packagist.org sees that the package has additional information and notifies Flex API server about it, this could abstracted with tags in `composer.json` and supporting tag subscription notifications in packagist.org
5. user installs the bundle, gets the benefits of Flex without Dave pushing his recipes to recipes* repos

This might be already possible somehow, but I didn't find a reference to it. This would also mean recipes-contrib would become (mostly) obsolete because maintainers would be able to ship their own recipes, the only use case for contrib would be adding recipes to bundles where you might not be able to add recipes directly, for example with unmaintained or uncooperative projects. Recipes repo would stay in place mostly for curation purposes (as in, enumerating quality packages and providing aliases for them).

javiereguiluz:
@dkarlovi I'm afraid this idea has been rejected before because Symfony Flex needs a central repository in order to implement some of the features that are planned for the coming months.

I understand this may be a bummer for you, but it's common for Symfony developers to use "central repositories". They use GitHub.com as the central repository of their public code and Packagist.org as the central repository to define Composer packages.

dkarlovi:
@javiereguiluz the difference is, they control their own additions to the central repository, it doesn't go through some 3rd party.

dkarlovi:
Also, you still have the central repo, as stated in 4), you just don't need to go through the recipes repo.

javiereguiluz:
I disagree with you about the 3rd party comment. For your code you rely on a private 3rd party called GitHub.com and for the Composer packages you rely on a private 3rd party called Packagist.org (this one has open sourced some parts of it). You can get banned  by them and they have Terms of Use, so your code/package can be rejected by them.

The idea is to apply the same rules for symfony/recipes-contrib ... **everything will be approved**, except some clearly illegal things, etc.

[...]

PR #174, opened 2017-10-05
Allowing to extend configurators and recipes

francoispluchino:
| Q             | A
| ------------- | ---
| Branch?       | 1.0
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | no
| Tests pass?   | yes
| Fixed tickets | #173
| License       | MIT
| Doc PR        | ~

Symfony Flex as a great automation tool to standardize the project configuration system, but for the moment, it is limited to the specific configurators in her source code.

It would be very interesting to be able to extend the features of Flex with other plugin Composer. This would allow for example to add a new specific configurator to a project, or to set up recipes for private repositories (see #71).

All recipes added by a Composer plugin are considered like recipes of contributor, with a specific origin. So, the process of the contributor's recipes is retained (wizard, skip, or accept the recipe).

To add a new configurator, it's simple:

```php
<?php

namespace Foo\FlexExtension;

use Composer\Composer;
use Composer\EventDispatcher\EventSubscriberInterface;
use Composer\IO\IOInterface;
use Composer\Plugin\PluginInterface;
use Symfony\Flex\FetchRecipesEvent;
use Symfony\Flex\FlexEvents;

class FlexExtension implements PluginInterface, EventSubscriberInterface
{
    /**
     * {@inheritdoc}
     */
    public static function getSubscribedEvents()
    {
        return array(
            FlexEvents::FETCH_RECIPES => array(
                array('fetchRecipes', 0),
            ),
        );
    }

    /**
     * {@inheritdoc}
     */
    public function activate(Composer $composer, IOInterface $io)
    {
        foreach($composer->getPluginManager()->getPlugins() as $plugin) {
            if ($plugin instanceof \Symfony\Flex\Flex) {
                $plugin->add('foo/custom-configurator', \CustomConfigurator::class);
                break;
            }
        }
    }
    
    public function fetchRecipes(FetchRecipesEvent $event)
    {
        $composerOperations = $event->getOperations();
        // loop logic in operations
        $event->addManifest('bar/foo', [
            'foo/custom-configurator' => [
                'custom' => 'config',
            ],
        ]);
    }
}
```

> **Note:** The configurator name must have a vendor prefix.

### DIFF
--- a/src/Configurator.php
+++ b/src/Configurator.php
@@ -64,6 +64,23 @@ class Configurator
         }
     }
 
+    public function add(string $name, string $class)
+    {
+        if (false === strpos($name, '/')) {
+            throw new \InvalidArgumentException(sprintf('Flex configurator name "%s" must be prefixed with the vendor name, ex. "foo/custom-configurator".', $name));
+        }
+
+        if (isset($this->configurators[$name])) {
+            throw new \InvalidArgumentException(sprintf('Flex configurator with the name "%s" already exists.', $name));
+        }
+
+        if (!is_subclass_of($class, AbstractConfigurator::class)) {
+            throw new \InvalidArgumentException(sprintf('Flex configurator class "%s" must extend the class "%s".', $class, AbstractConfigurator::class));
+        }
+
+        $this->configurators[$name] = $class;
+    }
+
     private function get($key): AbstractConfigurator
     {
         if (!isset($this->configurators[$key])) {
--- a/src/FetchRecipesEvent.php
+++ b/src/FetchRecipesEvent.php
@@ -1,0 +1,71 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Flex;
+
+use Composer\DependencyResolver\Operation\OperationInterface;
+use Composer\EventDispatcher\Event;
+use Composer\Package\PackageInterface;
+
+/**
+ * @author François Pluchino <francois.pluchino@gmail.com>
+ */
+class FetchRecipesEvent extends Event
+{
+    private $operations;
+
+    private $manifests = [];
+
+    /**
+     * @param string               $name       The event name
+     * @param OperationInterface[] $operations The operations of Composer
+     * @param array                $manifests  The manifests
+     */
+    public function __construct(string $name, array $operations, array $manifests)
+    {
+        parent::__construct($name);
+        $this->operations = $operations;
+        $this->manifests = $manifests;
+    }
+
+    /**
+     * @return OperationInterface[]
+     */
+    public function getOperations(): array
+    {
+        return $this->operations;
+    }
+
+    public function addManifest(PackageInterface $package, array $manifest)
+    {
+        $name = $package->getName();
+        $manifest['is_contrib'] = true;
+        if (!isset($manifest['origin'])) {
+            $manifest['origin'] = sprintf('%s:%s@composer-plugin recipe', $name, $package->getPrettyVersion());
+        }
+        $this->manifests[$name] = $manifest;
+    }
+
+    public function hasManifest(string $name): bool
+    {
+        return isset($this->manifests[$name]);
+    }
+
+    public function getManifest(string $name): array
+    {
+        return $this->manifests[$name] ?? [];
+    }
+
+    public function getManifests(): array
+    {
+        return $this->manifests;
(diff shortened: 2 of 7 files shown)


src/Configurator.php
```diff
@@ -64,6 +64,23 @@ public function unconfigure(Recipe $recipe)
...
+    {
+        if (false === strpos($name, '/')) {
+            throw new \InvalidArgumentException(sprintf('Flex configurator name "%s" must be prefixed with the vendor name, ex. "foo/custom-configurator".', $name));
+        }
+
+        if (isset($this->configurators[$name])) {
+            throw new \InvalidArgumentException(sprintf('Flex configurator with the name "%s" already exists.', $name));
+        }
+
+        if (AbstractConfigurator::class !== get_parent_class($class)) {
```
stof: use `is_subclass_of` instead, to allow deeper inheritance.

src/FetchRecipesEvent.php
```diff
@@ -0,0 +1,71 @@
...
+    }
+
+    public function addManifest(PackageInterface $package, array $manifest): void
+    {
+        $name = $package->getName();
+        $manifest['is_contrib'] = true;
+        if (!isset($manifest['origin'])) {
+            $manifest['origin'] = sprintf('%s:%s@composer-plugin recipe', $name, $package->getPrettyVersion());
+        }
+        $this->manifests[$name] = $manifest;
```
stof: should we allow overriding a manifest which already exists and comes from an official recipe ?

nicolas-grekas:
Wow you really want it :)

I think the biggest blocker to merge any "hook-enabling" code into Flex in the added maintenance burden this will create. @stof already mentioned this in #173, and I fully agree with him. We won't be able to provide a decent BC policy regarding the internal interface, nor the hook themselves, because we just want to provide a developer experience, not a base library that ppl could to extend internally.

Pushing this to us is asking for the Symfony project to endorse the cost of the maintenance.
But this is OSS and looking at comments #173 I think we don't want to buy that technical debt.

Personally, would thus respectfully recommend to not merge this PR or any similar approach in the short/mid term (nobody knows about the future :)).

src/FetchRecipesEvent.php
francoispluchino: Yes to allowing to add configurators. Note that in this case, the manifest is considered like a contributor's recipe.

francoispluchino:
@nicolas-grekas, Yes! I especially want to you show that it was very simple to implement with very little code. I understand the argument about the cost of the maintenance, even if it is extremely weak with a very light implementation.

> We won't be able to provide a decent BC policy regarding the internal interface, nor the hook themselves, nor the server protocol, because we just want to provide a developer experience, not a base library that ppl could extend internally.

With this implementation, you can keep only the argument about the internal interfaces. For the `AbstractConfigurator` class, ok, but for the `Event` class, it should not necessarily change. Regarding the Server protocol, it is independent of this proposal.

> But this is OSS and looking at comments #173 I think we don't want to buy that technical debt.

The comments of the issue #173 were mainly on the implementation proposal, which is no longer the same as this one, given that I have retained only the most important, and greatly simplified the implementation, while maintaining maximum flexibility.

Finally, the only argument that cannot have a counter-argument, is that you do not want Flex to be a great tool for the automation of the configuration. It is a choice, and I can only accept it. Too bad, but at least you had the courage to tell me that this is an arbitrary choice on the orientation of this library, and I thank you. Maybe we will rediscuss it in 5/6 years, but as you said, we are in the long term :-).

src/Configurator.php
francoispluchino: @stof Fixed.

francoispluchino:
@stof

The requested changes were made.

@nicolas-grekas

> But this is OSS and looking at comments #173 I think we don't want to buy that technical debt.

I forgot to mention that I am well aware that Symfony Flex is an Open Source Software, and that there is no need to add code that does not serve the community. However, my proposal is not the wish of a single person, but it is also the wish of other Symfony users. You have only to see the reactions on this [comment](https://github.com/symfony/flex/issues/70#issuecomment-299915048), or this [one](https://github.com/symfony/flex/issues/70#issuecomment-299717790), or that [one](https://github.com/symfony/flex/issues/71#issue-226862863).

It should not be forgotten that a very very large part of the Symfony community has not yet switched to Flex (and logically, you are very well aware of this point), and as a result, they have not yet manifested themselves, and that we are very few to participate of the issues compared to the totality of developers using Symfony (even if in absolute, the number is big). I think this point should not be overlooked, and that this type of functionality would serve a lot to all Symfony developers.

nicolas-grekas:
Closing as this won't make it. Feel free to fork if you need extra features of course.

PR #182, opened 2017-10-13
Fix xml doublequotes escaping

### LINKED ISSUE #180: Unescaped xml special chars environment variables breaks phpunit.xml.dist
Symfony Flex version: 1.0.29
PHP version: 7.1.10
OS: Ubuntu 16.04 from WSL

After having installed the orm-pack with `composer req orm`, my phpunit config is broken.
Flex adds the lines defined in the manifest.json. The `DATABASE_URL` is wrapped in double quotes.
But the double quotes breaks the attribute value. It should handle environment variables with double-quotes.

Generated lines :

```
<!-- ###+ doctrine/doctrine-bundle ### -->
<!-- Format described at http://docs.doctrine-project.org/projects/doctrine-dbal/en/latest/reference/configuration.html#connecting-using-a-url -->
<!-- For an SQLite database, use: "sqlite:///%kernel.project_dir%/var/data.db" -->
<!-- Set "serverVersion" to your server version to avoid edge-case exceptions and extra database calls -->
<env name="DATABASE_URL" value=""mysql://root@127.0.0.1:3306/symfony?charset=utf8mb4&serverVersion=5.7"" />
<!-- ###- doctrine/doctrine-bundle ### -->
```

EDIT: I've renamed to issue. I didn't noticed the unescaped `&` on the value.

toretto460:
Fix #180

### DIFF
--- a/src/Configurator/EnvConfigurator.php
+++ b/src/Configurator/EnvConfigurator.php
@@ -80,6 +80,7 @@ class EnvConfigurator extends AbstractConfigurator
                     $data .= '        <!-- '.$value." -->\n";
                 } else {
                     $value = $this->options->expandTargetDir($value);
+                    $value = htmlspecialchars($value);
                     $data .= "        <env name=\"$key\" value=\"$value\" />\n";
                 }
             }
--- a/tests/Configurator/EnvConfiguratorTest.php
+++ b/tests/Configurator/EnvConfiguratorTest.php
@@ -45,6 +45,7 @@ class EnvConfiguratorTest extends TestCase
             '#1' => 'Comment 1',
             '#2' => 'Comment 3',
             'APP_SECRET' => 's3cretf0rt3st',
+            'DATABASE_URL' => '"mysql://test"',
         ]);
 
         $envContents = <<<EOF
@@ -55,6 +56,7 @@ APP_DEBUG=0
 # Comment 1
 # Comment 3
 APP_SECRET=s3cretf0rt3st
+DATABASE_URL="mysql://test"
 ###< FooBundle ###
 
 EOF;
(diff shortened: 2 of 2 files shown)


src/Configurator/EnvConfigurator.php
```diff
@@ -80,6 +80,7 @@ private function configurePhpUnit(Recipe $recipe, $vars)
                     $data .= '        <!-- '.$value." -->\n";
                 } else {
                     $value = $this->options->expandTargetDir($value);
+                    $value = trim($value, '"');
```
xabbuh: What would I do now if I wanted to pass a string that was enclosed by double quotes?

toretto460: I agree, according to the xml specs it would be better to use `htmlspecialchars` 

> To allow attribute values to contain both single and double quotes, the apostrophe or single-quote character (') may be represented as "&apos;", and the double-quote character (") as "&quot;".
https://www.w3.org/TR/xml11/#syntax

Just pushed the fix

tests/Configurator/EnvConfiguratorTest.php
```diff
@@ -45,6 +45,7 @@ public function testConfigure()
             '#1' => 'Comment 1',
             '#2' => 'Comment 3',
             'APP_SECRET' => 's3cretf0rt3st',
+            'DATABASE_URL' => '"mysql://test"',
```
clement-michelet: I noticed (and updated my issue) that `&` was also not escaped and breaks the phpunit config.
A test environment variable could be something like this `"mysql://test/symfony?foo=<bar>&version=1.0"` to ensure that all characters are escaped.

What do you think ?

fabpot:
Closing in favor of #200

PR #183, opened 2017-10-13
fix xml escape special chars for phpunit.xml.dist

### LINKED ISSUE #180: Unescaped xml special chars environment variables breaks phpunit.xml.dist

clement-michelet:
Fix #180

### DIFF
--- a/src/Configurator/EnvConfigurator.php
+++ b/src/Configurator/EnvConfigurator.php
@@ -80,6 +80,7 @@ class EnvConfigurator extends AbstractConfigurator
                     $data .= '        <!-- '.$value." -->\n";
                 } else {
                     $value = $this->options->expandTargetDir($value);
+                    $value = filter_var($this->options->expandTargetDir($value), FILTER_SANITIZE_FULL_SPECIAL_CHARS);
                     $data .= "        <env name=\"$key\" value=\"$value\" />\n";
                 }
             }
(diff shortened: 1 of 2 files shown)


xabbuh:
looks like a duplicate of #182

clement-michelet:
Indeed. I opened the original issue and prepared a patch for it. 
I'll close this one and add a review on the other one.

PR #195, opened 2017-10-27
Fix allow-contrib inconsistency before asking

Pierstoval:
This is a problem, as when you already have `extra.symfony.allow-contrib` set to either `true` or `"true"`, the question is still always asked, where it shouldn't.


fabpot:
Thank you @Pierstoval.

PR #197, opened 2017-10-29
Install fix with lock

### LINKED ISSUE #192: Configuring recipes executed twice
Configuring recipes executed again for all recipes after `composer require` when `composer install` was not executed before

I copied [composer.json](https://raw.githubusercontent.com/symfony/skeleton/3.4/composer.json) from symfony/skeleton:3.4, then I executed following commands: `composer install`, `composer require twig/extensions`, `rm config/packages/twig_extensions.yaml` (twig extensions config is very weird, i do not need it).

Then I commit and push files to git repository and pull them to other directory, or just delete all gitignored files. And finally i run `composer require phpunit/phpunit --dev`

As a result, all recipes are unexpectedly configured again.

```
Symfony operations: 6 recipes (0e173ee762890f6f05b1f472942b65db)
  - Configuring symfony/flex (1.0): From github.com/symfony/recipes:master
  - Configuring phpunit/phpunit (4.7): From github.com/symfony/recipes:master
  - Configuring symfony/console (3.3): From github.com/symfony/recipes:master
  - Configuring symfony/routing (3.3): From github.com/symfony/recipes:master
  - Configuring symfony/framework-bundle (3.3): From github.com/symfony/recipes:master
  - Configuring twig/extensions (1.0): From github.com/symfony/recipes:master
Executing script cache:clear [OK]
Executing script assets:install --symlink --relative public [OK]
```

As a second result, deleted `config/packages/twig_extensions.yaml` comes back again with `.env` and `/public/bundles/`

But when I run `composer install` before `composer require phpunit/phpunit --dev`, only `.env` file comes back with it's default values (and this is not right too, i think). After `composer require phpunit/phpunit --dev` only `phpunit` was confugured (as expected):

```
Symfony operations: 1 recipe (b0660f8af662872cebd825d7d1b0dc0e)
  - Unconfiguring phpunit/phpunit (4.7): From github.com/symfony/recipes:master
Executing script cache:clear [OK]
Executing script assets:install --symlink --relative public [OK]
```

fabpot:
To fix #192 and some more edge cases, we need a proper lock file for Symfony recipes instead of trying to guess via the state of Composer. The lock stores more than needed for now, but that opens up some more opportunities for the future.

fixes #192

### DIFF
--- a/src/Flex.php
+++ b/src/Flex.php
@@ -42,7 +42,6 @@ class Flex implements PluginInterface, EventSubscriberInterface
     private $downloader;
     private $postInstallOutput = [''];
     private $operations = [];
-    private $originalLockHash;
     private static $activated = true;
 
     public function activate(Composer $composer, IOInterface $io)
@@ -68,9 +67,6 @@ class Flex implements PluginInterface, EventSubscriberInterface
         $this->configurator = new Configurator($composer, $io, $this->options);
         $this->downloader = new Downloader($composer, $io);
         $this->downloader->setFlexId($this->getFlexId());
-
-        // useful when symfony/flex is not yet installed (composer install without vendor/ for instance)
-        $this->updateOriginalLockHash();
 
         $search = 3;
         foreach (debug_backtrace(DEBUG_BACKTRACE_PROVIDE_OBJECT) as $trace) {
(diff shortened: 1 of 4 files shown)


src/Flex.php
```diff
@@ -337,6 +331,27 @@ private function formatOrigin(string $origin): string
...
+    private function shouldRecordOperation(OperationInterface $operation): bool
+    {
+        if ($operation instanceof UpdateOperation) {
+            $package = $operation->getTargetPackage();
+        } else {
+            $package = $operation->getPackage();
+        }
+
+        // FIXME: getNames() can return n names
+        $name = $package->getNames()[0];
```
Pierstoval: Is there a particular reason for this method to return more than one name? 🤔

stof: yes: `replace` and `provide` are giving more names to the package.

PR #200, opened 2017-11-02
Fix escaping in .env and phpunit.xml.dist

fabpot:
fixes symfony/recipes#233, #180, and #182


fabpot:
symfony/recipes#234 must be merged as well to fix the Doctrine issue.

PR #201, opened 2017-11-08
Fix line breaks for error messages

michaelperrin:
The arguments provided to the `StreamOutput` instance were not correctly set, leading to error messages with double line breaks like in this screenshot:

app_ _michael_michaelsmbp2017_ _-zsh_ _160x45


fabpot:
Thank you @michaelperrin.

PR #203, opened 2017-11-09
Fix container parameters configurator

### LINKED ISSUE #202: [ContainerConfigurator] Duplicating parameters
Duplicating parameters on `config/services.yaml` when the parameter already exists in the file.

Example:

```yaml
parameters:
  locale: pt # original parameter
  locale: en # parameter added after install
```

yceruto:
Fixes #202

- Configure and add `parameters` key if it does not exists.
- Configure without duplicated.
- Complex configuration with linebreak and comments.


fabpot:
Thank you @yceruto.

Issue #196, opened 2017-10-27
Make the path of `App\` namespace configurable


kolibri:
Hello.

I found this closed issue: https://github.com/symfony/flex/issues/138
And I want to wide the scope of this and suggest to configure the path to the `App\` namespace configurable, or in other words, the directory that is under control of flex and its manifests.

As we already configure the path to the `App\` namespace inside the `composer.json` it would be nice if that could also be the configuration of "Here Flex, this is your directory. If you copy things from manifests, place them there."

The use case could be, to have the possibility to separate business logic from framework integration inside the `src` directory.

Pierstoval:
If you're just talking about the `App` namespace, I think this will never be possible. The issue #138 you linked has enough discussions about this.

For `src/` you can change it by customizing the `src-dir` option in your `composer-json`:

```json
{
    "extra": {
        "src-dir": "lib/"
    }
}
```

This will not change the namespace though.

If you want to change the namespace, change it for everything that's installed by Flex manually (entities, etc.), and change the `psr-4` rules in your `composer.json` file.

xabbuh:
@Pierstoval The issue is that when making these adjustments in the `composer.json` file, they will not be taken into account insides recipes. For example, the DoctrineBundle recipe uses the hardcoded `src` path for its mapping configuration: https://github.com/symfony/recipes/blob/master/doctrine/doctrine-bundle/1.6/config/packages/doctrine.yaml#L12

A possible approach @kolibri and me discussed last week was to be able to refer to `%SRC_DIR` and so inside files of your recipe that will be copied by Flex.

fabpot:
Closing this issue as I'm not interested in supporting this use case. If somebody wants to make it work, I would happily review pull requests.

PR #204, opened 2017-11-10
[WIP][Proposal] Replace dir vars in copied manifest files

kolibri:
| Q             | A
| ------------- | ---
| Branch?       | 1.0 (1.1)
| Bug fix?      | no
| New feature?  | yes
| BC breaks?    | no
| Deprecations? | yes
| Tests pass?   | there is no test (yet) (so "no")
| Fixed tickets | #196
| License       | MIT
| Doc PR        | ---

With this PR we can replace the directory variables, that are already available in the `manifest.json` files (`%BIN_DIR%`, `%CONF_DIR%`, `%CONFIG_DIR%`, `%SRC_DIR%`, `%VAR_DIR%`, `%PUBLIC_DIR%`) inside all files, that are copied from manifests.
So, this can makes manifests more usable with a modified directory structure.

(Yes, this is very WIP, and I know, that we have to update all recipies to take advantage of this change. Use case for this PR is, to have something to discuss in the related ticket #196 )

I'm not sure, if it makes sense to limit this feature to `yaml` (or configuration) files only.

### DIFF
--- a/src/Configurator/CopyFromRecipeConfigurator.php
+++ b/src/Configurator/CopyFromRecipeConfigurator.php
@@ -62,7 +62,14 @@ class CopyFromRecipeConfigurator extends AbstractConfigurator
             mkdir(dirname($to), 0777, true);
         }
 
-        file_put_contents($to, $contents);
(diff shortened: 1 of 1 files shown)


MetalArend:
@kolibri Awesome. Could this also be used to move the files to the updated path? So, files that are referenced as to be placed in the src directory would be moved to the actual mapped directory?

https://github.com/symfony/flex/issues/374

PR #211, opened 2017-11-23
do not try to copy .env.dist if it doesn't exist

garak:
See https://github.com/symfony/recipes/issues/270


fabpot:
Thank you @garak.

PR #216, opened 2017-11-27
Adding a message & avoiding the composer.json revert on script failure

### LINKED ISSUE #142: Failure during install fails to update composer.json file
### steps to reproduce

`composer create symfony/skeleton foo`
`cd foo`
`composer config minimum-stability dev`
`composer req api`

### buggy outcome 1/2
```
Script make cache-warmup returned with error code 2
!!  
!!    [Symfony\Component\DependencyInjection\Exception\ServiceNotFoundException]  
!!    You have requested a non-existent service "".                               
```
how do you quickly debug that btw? the error message is just useless...

### buggy outcome 2/2

composer.json file doesn't contain anything about api-platform, despite the fact the package is installed and listed by `composer show`.
Do `composer up` and they are just removed, leaving Flex in an inconsistent state.

weaverryan:
Fixes #142 

This depends on #composer/composer#6842

An example of what a failure looks like:

screen shot 2017-11-27 at 1 18 12 pm

We now do NOT revert the `composer.json`, and instead are able to show a nice message at the bottom to tell the user that "it's ok!".

### DIFF
--- a/src/Flex.php
+++ b/src/Flex.php
@@ -18,9 +18,11 @@ use Composer\DependencyResolver\Operation\InstallOperation;
 use Composer\DependencyResolver\Operation\OperationInterface;
 use Composer\DependencyResolver\Operation\UpdateOperation;
 use Composer\DependencyResolver\Operation\UninstallOperation;
+use Composer\EventDispatcher\ScriptExecutionException;
 use Composer\Factory;
 use Composer\EventDispatcher\EventSubscriberInterface;
 use Composer\Installer;
+use Composer\Installer\InstallerErrorEvent;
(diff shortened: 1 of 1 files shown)


Pierstoval:
Hehe, I made a demo a few hours ago with some students and it was a perfect failure because of the failing script.

However I'm not sure about the `!!` part: I think it's not reassuring enough for the end-user about the composer script 🤔

weaverryan:
@Pierstoval Good point :). I've replaced with 2 spaces - so no extra `!` on the last two lines, but they'll still be "indented" to line up with the text above from Composer.

PR #220, opened 2017-12-01
Add >= to "configuring" message

nicolas-grekas:
It's surprising to see "3.3" when installing a 4.0.0 component. This should fix it.


nicolas-grekas:
Or should we use `≥`?

fabpot:
Thank you @nicolas-grekas.

Issue #225, opened 2017-12-04
FrameworkBundle must be configured before all other packages


covex-nn:
I've copied composer.json from [symfony/skeleton](https://github.com/symfony/skeleton/blob/3.4/composer.json), added `"sonata-project/notification-bundle": "^3.2"` into `require` section. And after `composer install` i've got an error message:

```
Symfony operations: 11 recipes (073fc8f0285c8a9a7e50f725a90a200a)
  - Configuring symfony/flex (>=1.0): From github.com/symfony/recipes:master
  - Configuring sonata-project/datagrid-bundle (>=2.3.0): From auto-generated recipe
  - Configuring twig/extensions (>=1.0): From github.com/symfony/recipes:master
  - Configuring symfony/translation (>=3.3): From github.com/symfony/recipes:master


  [ErrorException]
  file(C:\Users\Mindubaev\Docker\3\test/config/services.yaml): failed to open stream: No such file or directory
```

This error was occurred because Flex tried to add container parameter for `symfony/translation` before `symfony/framework-bundle` was configured.

I think that `symfony/framework-bundle` must be configured before all other packages (except `symfony/flex`) to avoid this error.

Pierstoval:
Did you add `sonata-project/notification-bundle` _before_ or _after_ the FrameworkBundle in the composer.json `require` section? I don't know how Composer behaves in this kind of case, if we can add a priority just by changing the order or something 😕 Because in the real end, priority should be Flex > FrameworkBundle > everything else

xabbuh:
Is there a specific reason you did not use the `composer create-project` command?

covex-nn:
@Pierstoval i tried both, result is the same: Composer installed and Flex tried to configure `symfony/translation` before `symfony/framework-bundle`.

To fix this we could sort recipes in [Symfony\Flex\Flex::update](https://github.com/symfony/flex/blob/8917eb9e7d490088b85ea2a4cd7f2ad52bb0d9ee/src/Flex.php#L130) after `list($recipes, $vulnerabilities) = $this->fetchRecipes()` to put Flex and FrameworkBundle in the beginning of the list for `install` job and to the end for `uninstall`.

@xabbuh i tried to find a way how to develop and test recipes localy to show a result someone before PR to recipes-contrib. The idea was to store only recipe's files + composer.json in git repository, to gitignore everything else and to let composer reinstall application again with `composer install` in other place.

Original composer.json, where Flex failed first:

```json
{
    "autoload": {
[...]

PR #230, opened 2017-12-05
Sort recipes on update before configuration

covex-nn:
Recipe for `symfony/framework-bundle` must be applied after recipe for `symfony/flex` and before all other recipes on install/update.

See #225

### DIFF
--- a/src/Flex.php
+++ b/src/Flex.php
@@ -118,7 +118,11 @@ class Flex implements PluginInterface, EventSubscriberInterface
     {
         $operation = $event->getOperation();
         if ($this->shouldRecordOperation($operation)) {
-            $this->operations[] = $operation;
+            if ($operation instanceof InstallOperation && 'symfony/framework-bundle' === $operation->getPackage()->getName()) {
+                array_unshift($this->operations, $operation);
+            } else {
+                $this->operations[] = $operation;
+            }
         }
     }
 
--- a/tests/FlexTest.php
+++ b/tests/FlexTest.php
@@ -9,7 +9,7 @@
  * file that was distributed with this source code.
  */
 
-namespace Symfony\Flex\Tests\Configurator;
+namespace Symfony\Flex\Tests;
 
 use Composer\Composer;
 use Composer\Config;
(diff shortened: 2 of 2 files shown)


src/Recipe.php
```diff
@@ -70,6 +70,32 @@ public function isContrib(): bool
         return $this->data['is_contrib'] ?? false;
     }
 
+    public static function queueCompare(Recipe $first, Recipe $second): int
+    {
+        static $order = ['symfony/flex' => -2, 'symfony/framework-bundle' => -1];
```
Pierstoval: Why not put them in a static class var?

src/Recipe.php
```diff
@@ -70,6 +70,32 @@ public function isContrib(): bool
...
+                $weight[$i] = $order[$name];
+            }
+        }
+        if ($weight[0] == $weight[1]) {
+            return 0;
+        } elseif ($weight[0] <= $weight[1]) {
+            return -1;
+        } else {
+            return 1;
+        }
```
Pierstoval: Isn't this a good use-case for the spaceship operator? 😄

covex-nn: I haven't used it before and forgot about it! =) I'll use it

src/Recipe.php
covex-nn: I'll put array in a `private static $queueOrder`

src/Recipe.php
```diff
@@ -70,6 +71,24 @@ public function isContrib(): bool
         return $this->data['is_contrib'] ?? false;
     }
 
+    public static function queueCompare(Recipe $first, Recipe $second): int
```
nicolas-grekas: - instead of a generic mechanism, I think a special case for framework-bundle only could be enough
- this means we should also remove the recipe for symfony/flex, which is useless anyway, since framework-bundle is a mandatory dep
- which means we could "just" deal with that case in `Flex::fetchRecipes()`, wdyt?

nicolas-grekas: see https://github.com/symfony/recipes/pull/295

covex-nn: @nicolas-grekas if there will be only `symfony/framework-bundle` recipe, then we could only prepend "install FrameworkBundle" operation to `$this->operations` instead of appending it in [`Symfony\Flex\Flex::record`](https://github.com/symfony/flex/blob/8917eb9e7d490088b85ea2a4cd7f2ad52bb0d9ee/src/Flex.php#L117)

covex-nn:
@nicolas-grekas i have updated PR, now it is waiting for merge flex recipe with framework-bundle one

src/Flex.php
```diff
@@ -118,7 +118,11 @@ public function record(PackageEvent $event)
     {
         $operation = $event->getOperation();
         if ($this->shouldRecordOperation($operation)) {
-            $this->operations[] = $operation;
+            if ($operation instanceof InstallOperation && $operation->getPackage()->getName() == 'symfony/framework-bundle') {
```
nicolas-grekas: yoda + strict comparison:
`'symfony/framework-bundle' === $operation->getPackage()->getName()`

fabpot:
@covex-nn You can probably remove the `WIP` in the PR title.

src/Flex.php
covex-nn: @nicolas-grekas done

fabpot:
Thank you @covex-nn.

PR #231, opened 2017-12-06
Stick to framework-bundle's version by default

nicolas-grekas:
Works around https://github.com/composer/composer/issues/6869

### DIFF
--- a/src/PackageResolver.php
+++ b/src/PackageResolver.php
@@ -11,6 +11,7 @@
 
 namespace Symfony\Flex;
 
+use Composer\Factory;
 use Composer\Package\Version\VersionParser;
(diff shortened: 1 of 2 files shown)


src/PackageResolver.php
```diff
@@ -29,7 +30,7 @@ public function __construct(Downloader $downloader)
         $this->downloader = $downloader;
     }
 
-    public function resolve(array $arguments = []): array
+    public function resolve(array $arguments = [], bool $isRequire = false): array
```
Pierstoval: Shouldn't be more like `$isRequireCommand`?

fabpot:
Thank you @nicolas-grekas.

Issue #238, opened 2017-12-11
Flex fails to modify .env if .env.dist is missing


weaverryan:
The description sounds like an edge-case, but the situation occurs when upgrading an existing project to a Symfony Flex project (at one stage, I add Flex, then run `composer install` to install recipes for all my existing packages).

To repeat:

```
composer create-project symfony/skeleton missing_env
cd missing_env
rm .env.dist
composer require orm
```

`doctrine/doctrine-bundle` will report that its recipe was installed, but there are no changes to the `.env` file.

weaverryan:
See #239 :)

PR #239, opened 2017-12-11
Guaranteeing that Flex is installed THEN fwbundle

weaverryan:
See #238.

I talked with @nicolas-grekas about this. By merging the Flex recipe into FWBundle, we created a possibility (which I hit while creating our upgrade tutorial) that the `.env.dist` file is not present. And then, recipes may start installing incorrectly because this file is needed (so that recipes can update it and `.env`).

The fix is this PR + to revert https://github.com/symfony/recipes/pull/295

### DIFF
--- a/src/Flex.php
+++ b/src/Flex.php
@@ -118,8 +118,20 @@ class Flex implements PluginInterface, EventSubscriberInterface
     {
         $operation = $event->getOperation();
         if ($this->shouldRecordOperation($operation)) {
-            if ($operation instanceof InstallOperation && 'symfony/framework-bundle' === $operation->getPackage()->getName()) {
-                array_unshift($this->operations, $operation);
+            if ($operation instanceof InstallOperation && in_array($packageName = $operation->getPackage()->getName(), ['symfony/framework-bundle', 'symfony/flex'])) {
+                if ('symfony/flex' === $packageName) {
+                    array_unshift($this->operations, $operation);
+                } else {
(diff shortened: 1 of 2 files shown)


src/Flex.php
```diff
@@ -118,8 +118,20 @@ public function record(PackageEvent $event)
     {
         $operation = $event->getOperation();
         if ($this->shouldRecordOperation($operation)) {
-            if ($operation instanceof InstallOperation && 'symfony/framework-bundle' === $operation->getPackage()->getName()) {
-                array_unshift($this->operations, $operation);
+            if ($operation instanceof InstallOperation && in_array($packageName = $operation->getPackage()->getName(), ['symfony/framework-bundle', 'symfony/flex'])) {
+                if ($packageName === 'symfony/flex') {
```
fabpot: Yoda condition?

src/Flex.php
```diff
@@ -118,8 +118,20 @@ public function record(PackageEvent $event)
...
+                    array_unshift($this->operations, $operation);
+                } else {
+                    if ($this->operations && $this->operations[0] instanceof InstallOperation && $this->operations[0]->getPackage()->getName() === 'symfony/flex') {
```
fabpot: Yoda here as well

weaverryan: Right you are ;)

fabpot:
Thank you @weaverryan.

PR #241, opened 2017-12-12
Add support for unpacking Composer packages

fabpot:
This PR is short but packed (pun intended) with great features.

Basically, it helps fix an issue people have with Flex: starting a new project.

### DIFF
--- a/src/Command/RequireCommand.php
+++ b/src/Command/RequireCommand.php
@@ -12,13 +12,23 @@
 namespace Symfony\Flex\Command;
 
 use Composer\Command\RequireCommand as BaseRequireCommand;
+use Composer\DependencyResolver\Pool;
+use Composer\Factory;
+use Composer\Json\JsonFile;
+use Composer\Json\JsonManipulator;
+use Composer\Package\Link;
+use Composer\Package\Package;
+use Composer\Package\Version\VersionParser;
 use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Flex\PackageResolver;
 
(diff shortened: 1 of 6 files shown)


Pierstoval:
About the post, I wonder why it's only reserved to the `symfony-pack` type? Why can't it be associated with the already existing `metapackage`, which is, by the way, [documented](https://getcomposer.org/doc/04-schema.md#type)?

nicolas-grekas:
The reason is that when you require several packages at once, only some of them should be unpacked.
`composer require a/b a/c b/a b/c --unpack`
only those that were designed to be unpacked should be unpacked.

Pierstoval:
Yeah, but a metapackage is precisely the kind of package that _can_ be unpacked: if a `metapackage` contains requirements, they should be unpackable. If the `metapackage` is only here to create conflicts (like `symfony/lts` or `roave/security-advisories`), even if you set `--unpack`, it will not unpack anything, so it can be considered safe.

I'm just commenting on this because it will reinforce the idea for some people that Flex is only for Symfony packages, which is not the case. Anyone willing to create a pack should be able to unpack it even if it's not a Symfony-related package.

nicolas-grekas:
can != should - that's the point.

> it will reinforce the idea for some people that Flex is only for Symfony packages

Flex is only for Symfony apps yes, framework bundle is a, implicit requirement of all recipes.
Let's first try improving the Symfony-based experience, and if some ideas are more broad then fine, that's OSS, ppl can contribute to Composer. We're cannot fix the world (who are we to even think we could).

fabpot:
Symfony Flex is currently definitely for Symfony, no questions about it.

fabpot:
Based on the feedback, unpack never happens recursively. It gives more control. There is now also a new `unpack` command that allows to unpack already installed packs.

PR #242, opened 2017-12-12
Make ContainerConfigurator use config-dir option for file paths

### LINKED ISSUE #236: Use %CONFIG_DIR% in ContainerConfigurator?
Hi,

I am organizing my backend (Symfony) and frontend (JS/SCSS/etc) files in two separate folders, with `composer.json` and other stuff at the root of the project. (Note: `backend` and `frontend` are only for file organization. Symfony will still be rendering most of the frontend, so it's not like its two separate apps.)

So, I've got `backend/bin/, backend/config/, backend/src/`, etc. I configured `composer.json` to set `"config-dir": "backend/config"`. That has worked fine for installing several bundles.

However, trying to `composer req validator`, I get the error:

    file(E:\_projects\foo/config/services.yaml): failed to open stream: No such file or directory

Looking at the [ContainerConfigurator](https://github.com/symfony/flex/blob/master/src/Configurator/ContainerConfigurator.php#L45), it's using `getcwd().'/config/services.yaml'`.

Compare that to the [BundleConfigurator](https://github.com/symfony/flex/blob/master/src/Configurator/BundlesConfigurator.php#L103) which uses `$this->options->expandTargetDir('%CONFIG_DIR%/bundles.php')`.

I think the ContainerConfigurator should also use the config-dir param. Right? I'll open a PR if so.

simshaun:
Fixes #236

### DIFF
--- a/src/Configurator/ContainerConfigurator.php
+++ b/src/Configurator/ContainerConfigurator.php
@@ -27,7 +27,7 @@ class ContainerConfigurator extends AbstractConfigurator
     public function unconfigure(Recipe $recipe, $parameters)
     {
         $this->write('Unsetting parameters');
-        $target = getcwd().'/config/services.yaml';
+        $target = $this->getConfDir().'/services.yaml';
         $lines = [];
         foreach (file($target) as $line) {
             foreach (array_keys($parameters) as $key) {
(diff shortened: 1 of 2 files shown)


fabpot:
Thank you @simshaun.

PR #250, opened 2017-12-26
Don't remove the main Git directory

javiereguiluz:
This is an attempt to fix https://github.com/symfony/recipes/issues/324  but I haven't tested it properly yet!

### DIFF
--- a/src/Configurator/CopyFromRecipeConfigurator.php
+++ b/src/Configurator/CopyFromRecipeConfigurator.php
@@ -72,6 +72,12 @@ class CopyFromRecipeConfigurator extends AbstractConfigurator
     {
         foreach ($manifest as $source => $target) {
             $target = $this->options->expandTargetDir($target);
+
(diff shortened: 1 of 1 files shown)


xabbuh:
What about doing this one level higher by just not allowing to remove the recipe applied for `symfony/flex`?

nicolas-grekas:
LGTM for @xabbuh's proposal.

nicolas-grekas:
But then some of FrameworkBundle's gitignore should be moved to the flex recipe (.env and vendor), isn't it?

weaverryan:
I prefer the *current* patch over @xabbuh's proposal, actually. It makes more sense to me to day "`.git` is special, don't remove it" versus "the entire symfony/flex recipe is special, don't remove it". 

👍 for the PR as it is currently

fabpot:
Thank you @javiereguiluz.

PR #256, opened 2017-12-29
Add symfony/thanks to requirements

nicolas-grekas:
I propose to move symfony/thanks from the skeleton to Flex. Here is the first part of it.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -10,7 +10,8 @@
     ],
     "require": {
         "php": "^7.0",
-        "composer-plugin-api": "^1.1"
+        "composer-plugin-api": "^1.1",
+        "symfony/thanks": "^1"
     },
     "require-dev": {
         "composer/composer": "^1.4",


nicolas-grekas:
Just to conterbalance down votes, here are some people that like the plugin, and know it does NOTHING unless YOU explicitly call "composer thanks"
https://github.com/symfony/thanks/issues/24
592 stars on the repo also means ~592 users used the command.

We should help people sending good vibes to OSS authors.

docteurklein:
Maybe this would better be put in the "require-dev"   portion of the target project?

nicolas-grekas:
In fact, that's already provided by the skeleton. My call here is that Flex provides an experience, and that thanks should be part of that experience. :)

lyrixx:
:+1:

weaverryan:
Just curious: what’s the advantage/difference of having it here versus the skeleton?

Majkl578:
@weaverryan To force it to even more people in all projects that migrate to flex. Very nasty practice.

dunglas:
👍, the main benefit I see to move `thanks` as a Flex dependency is it help keeping the Skeleton's `composer.json` minimal.

@Majkl578 to force people to store those huge 265 LOC on their hard drive, that will never be executed if the user doesn't run `composer thanks`? So nasty...

nicolas-grekas:
> Another solution would be to remove the symfony/thanks repository and move these 2 classes in Flex directly. So it will benefit only to Symfony, not to the PHP ecosystem at a whole. Less nasty?

More nasty IMHO :)

> what’s the advantage/difference of having it here versus the skeleton?

what Kévin said: Thanks should come as part of the Flex experience, it's not a (dev) dependency of yours.

mickaelandrieu:
@nicolas-grekas I don't see the point of this drama, just merge! This is not the first, and wont be the last time something is merged without any review, and in this *specific case* I don't see the real value of the community opinion.

Haters gonna hate, keep doing a good job Nicolas!

Majkl578:
> the main benefit I see to move thanks to a Flex dependency is it helps keeping the Skeleton's composer.json minimal.

Where exactly is the benefit for Flex here?

> @Majkl578 to force people to store those huge 265 LOC on their hard drive, that will never be executed if the user doesn't run composer thanks?

To force people to get dependency they don't need and never willingly requested, which will end up in their production environment.

> Another solution would be to remove the symfony/thanks repository and move these 2 classes in Flex directly. So it will benefit only to Symfony, not to the PHP ecosystem at a whole. Less nasty?

You contradict here. This suggestion doesn't make any sense. The thanks package is vendor neutral, flex is not. But that doesn't mean it should be either merged into flex or be depended upon. This is not discussion about symfony/thanks which is undoubtedly good for OSS, but about strong requirement in Flex - please keep the talk on topic.

jschroed91:
👍

theofidry:
@Majkl578 FWIW I really don't get what's the fuss about, neither here nor on Slack... It's just a package frigging small harmless package to say thanks when you want to. Put that in suggest? Meh, you _really_ think people will bother?

>To force people to get dependency they don't need and never willingly requested, which will end up in their production environment.

I also think there is more worrying dependency and it won't end up in production environments unless you install dev dependencies (in which case you probably don't care anyway).

Now please guys chill out a bit, I really don't get how we arrive to debates like this for this kind of thing.

Pierstoval:
Then get these classes and put them in Composer itself, then nobody will cry, and more PHP packages can benefit from it.

slootjes:
Great idea @Pierstoval to make it a part of Composer instead, it does sound like the best solution to me too. This could be a global PHP community thing rather than something for Symfony Flex users (and users who manually require it).

javiereguiluz:
@slootjes problem is nobody has volunteered to make that happen and also, Composer leader doesn't like the idea (at least yet). See https://twitter.com/seldaek/status/947100791814443008

slootjes:
Ah... Creating the PR doesn't seem like a big issue as the code is already there but if Jordi doesn't want to have it added it's pretty much a dead end.

Besides that I don't see an issue in having this package to be a requirement from flex, it's a great addition and someone who doesn't want it can remove the few files manually anyway.

nicolas-grekas:
Closing in favor of https://github.com/symfony/flex/pull/262
I just wish we could learn to be a bit less harsh...
Anyway, thank you all for the discussion and happy new year.

PR #260, opened 2018-01-02
Help people understand generated files are *theirs*

nicolas-grekas:
image

A response to https://twitter.com/fabpot/status/938074790023139329

### DIFF
--- a/src/Flex.php
+++ b/src/Flex.php
@@ -170,6 +170,7 @@ class Flex implements PluginInterface, EventSubscriberInterface
 
         $this->io->writeError(sprintf('<info>Symfony operations: %d recipe%s (%s)</>', count($recipes), count($recipes) > 1 ? 's' : '', $this->downloader->getSessionId()));
         $installContribs = $this->composer->getPackage()->getExtra()['symfony']['allow-contrib'] ?? false;
+        $manifest = null;
         foreach ($recipes as $recipe) {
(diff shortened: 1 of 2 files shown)


src/Flex.php
```diff
@@ -231,6 +232,15 @@ function ($value) {
             }
         }
 
+        if (null !== $manifest) {
+            array_unshift(
+                $this->postInstallOutput,
+                '',
+                '<info>A few files may have been created to configure newly installed packages.</>',
+                'Don\'t hesitate to <comment>review</>, <comment>edit</> and <comment>commit</> them: these files are <comment>*yours*</>.'
```
stof: I don't think you should put stars in it. If you want to make it bold, make it bold for real (I think the Console component supports it)

Pierstoval: The console supports it, but not all terminals 😉

nicolas-grekas: stars removed (bold+yellow didn't look nice)

src/Flex.php
```diff
@@ -231,6 +232,15 @@ function ($value) {
             }
         }
 
+        if (null !== $manifest) {
+            array_unshift(
+                $this->postInstallOutput,
+                '',
+                '<info>A few files may have been created to configure newly installed packages.</>',
```
fabpot: Some files may have been created or updated to configure your newly installed packages.

fabpot:
Thank you @nicolas-grekas.

PR #261, opened 2018-01-02
Speedup Composer by downloading packages in parallel

nicolas-grekas:
Purpose
-------

This PR is heavily inspired by the great [hirak/prestissimo](https://github.com/hirak/prestissimo) by Hiraku NAKANO.

Its purpose is to provide a portable alternative to `hirak/prestissimo`, so that it can be bundled by default into Flex: unlike `hirak/prestissimo`, it doesn't require the `curl` extension and has zero extra dependencies. Instead, it reuses the existing code infrastructure in Composer to download packages.

If `hirak/prestissimo` is installed, this behavior disables itself automatically.

Limitations
-----------

- By reusing Composer's downloader, it is not possible to create server connections concurrently.
  Instead, the second download is started when the first one receives its first bytes from the server, and so on.
- JSON files downloaded from package repositories (e.g. from packagist.org) are still loaded sequentially (could be improved, PR welcome.)
- Packages installed by cloning source code repositories are not parallelized (neither with `hirak/prestissimo`.)

### DIFF
--- a/src/Flex.php
+++ b/src/Flex.php
@@ -21,6 +21,8 @@ use Composer\DependencyResolver\Operation\UninstallOperation;
 use Composer\Factory;
 use Composer\EventDispatcher\EventSubscriberInterface;
 use Composer\Installer;
+use Composer\Installer\InstallerEvent;
+use Composer\Installer\InstallerEvents;
 use Composer\Installer\PackageEvent;
 use Composer\Installer\PackageEvents;
 use Composer\Installer\SuggestedPackagesReporter;
@@ -32,6 +34,7 @@ use Composer\Json\JsonManipulator;
 use Composer\Plugin\PluginInterface;
 use Composer\Script\Event;
 use Composer\Script\ScriptEvents;
+use Hirak\Prestissimo\Plugin as Prestissimo;
 
 /**
  * @author Fabien Potencier <fabien@symfony.com>
@@ -46,6 +49,7 @@ class Flex implements PluginInterface, EventSubscriberInterface
     private $postInstallOutput = [''];
     private $operations = [];
     private $lock;
+    private $cacheDirPopulated = false;
     private static $activated = true;
 
     public function activate(Composer $composer, IOInterface $io)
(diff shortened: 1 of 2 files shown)


nicolas-grekas:
> composer create-project symfony/skeleton demo

- 30s native, 16s with this PR, 9s with prestissimo - that's the curl advantage we cannot rely on

> installing 134 packages on Blackfire

- 3m45s native, 2min with this PR, prestissimo 35s (but needs cURL)

src/Flex.php
```diff
@@ -250,6 +253,21 @@ public function executeAutoScripts(Event $event)
         $this->io->write($this->postInstallOutput);
     }
 
+    public function populateCacheDir(InstallerEvent $event)
+    {
+        if (extension_loaded('curl') && class_exists(Prestissimo::class, false)) {
```
Pierstoval: Are you sure `Prestissimo` is the right class? I can't find it on https://github.com/hirak/prestissimo

bitgandtter:
Can we include a validation for the cURL ext and if its enabled use that approach to gain the full speed?

src/Flex.php
nicolas-grekas: missing alias, fixed thanks

nicolas-grekas:
@bitgandtter not sure what you mean, but using cURL is far from trivial - basically, that's what https://github.com/hirak/prestissimo provides, so just use it if you have curl.

bitgandtter:
yes my bad @nicolas-grekas i miss the above check https://github.com/symfony/flex/pull/261#discussion-diff-159292888R258 that in fact was the intent of my comment

fabpot:
Flex is part of the `require` section, so always available. It's then the responsibility of Flex to do nothing in case of an install (for deployment), which is exactly what it does.

LKDevelopment:
@nicolas-grekas why you doesn`t contribute this to Composer so every PHP Developer would profit from IT Not only Symfony Devs.

nicolas-grekas:
@GromNaN @LKDevelopment I'm focusing on improving the Symfony + Flex based experience that's correct. That's already enough work for one (passionate) person.
The good news is that this is going to be released under a very permissive license so that anyone else could take over.
For myself, it serves my current motivation and goals (and available time for balanced a life.)

nicolas-grekas:
The technical answer is that things cannot be "just" added to Composer, nor released standalone: projects have policies, stability to ensure, maintenance to accept, etc. It could happen, but that's a significant effort. Better start with some pragmatic first step.

fabpot:
For the record, I think @nicolas-grekas even submitted a pull request on Composer to add such parallelization a long time ago... yep, 2.5 years ago, still open: composer/composer#3951

staabm:
with https://github.com/amphp/artax you can get a zero dependency (php-ext wise) http parallel (non-blocking) client, maybe this is an alternative to what you have right now (and I could provide better performance).

fabpot:
Thank you @nicolas-grekas.

PR #262, opened 2018-01-02
invite to install and use Thanks

nicolas-grekas:
alternative to https://github.com/symfony/flex/pull/256
sibling to https://github.com/symfony/thanks/pull/28

### DIFF
--- a/src/Flex.php
+++ b/src/Flex.php
@@ -31,10 +31,12 @@ use Composer\IO\ConsoleIO;
 use Composer\IO\NullIO;
 use Composer\Json\JsonFile;
 use Composer\Json\JsonManipulator;
+use Composer\Plugin\CommandEvent;
 use Composer\Plugin\PluginInterface;
 use Composer\Script\Event;
 use Composer\Script\ScriptEvents;
 use Hirak\Prestissimo\Plugin as Prestissimo;
+use Symfony\Thanks\Thanks;
 
 /**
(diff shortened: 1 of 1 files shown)


fabpot:
Thank you @nicolas-grekas.

src/Flex.php
```diff
@@ -416,6 +436,7 @@ public static function getSubscribedEvents(): array
             ScriptEvents::POST_CREATE_PROJECT_CMD => 'configureProject',
             ScriptEvents::POST_INSTALL_CMD => 'install',
             ScriptEvents::POST_UPDATE_CMD => 'update',
+            PluginEvents::COMMAND => 'inspectCommand',
```
pyrech: @nicolas-grekas looks like v1.0.55 is broken as the `PluginEvents` class does not exist. Did you mean `CommandEvent`?

javiereguiluz:
We could display it randomly one out of 100 times.

nicolas-grekas:
@ciaranmcnulty my proposal: you install the plugin as suggested, then someone fixes https://github.com/symfony/thanks/issues/29 (or in swapped order)

nicolas-grekas:
> displaying every time I run composer

every time composer *update* runs - that's not the same :)

ciaranmcnulty:
It'd be nice if it didn't nag _when I decide not to star the repos_

PR #265, opened 2018-01-03
Fix flex v1.0.55 using wrong class name

pyrech:
Ref https://github.com/symfony/flex/pull/262#pullrequestreview-86327921


fabpot:
fixed in 1.0.56

PR #266, opened 2018-01-03
Fix display of thanks invite

nicolas-grekas:
my bad, didn't notice this early "return"


fabpot:
Thank you @nicolas-grekas.

PR #267, opened 2018-01-03
Dont do parallel download for one package

nicolas-grekas:


fabpot:
Thank you @nicolas-grekas.

PR #270, opened 2018-01-06
Added console output for copied and removed files

szymach:
Result of the discussion from [#25651](https://github.com/symfony/symfony/issues/25651).
  
  TODO:
- [ ] check for verbosity
- [x] make sure adding/deleting both directories and single files is logged,
- [x] more test cases for adding / removing directories, as well as for invalid data.

### DIFF
--- a/src/Configurator/CopyFromPackageConfigurator.php
+++ b/src/Configurator/CopyFromPackageConfigurator.php
@@ -37,25 +37,16 @@ class CopyFromPackageConfigurator extends AbstractConfigurator
         foreach ($manifest as $source => $target) {
             $target = $this->options->expandTargetDir($target);
             if ('/' === substr($source, -1)) {
-                $this->copyDir(
-                    $this->concatenatePathParts([$from, $source]),
-                    $this->concatenatePathParts([$to, $target])
-                );
+                $this->copyDir($this->concatenatePathParts([$from, $source]), $this->concatenatePathParts([$to, $target]));
             } else {
                 $targetPath = $this->concatenatePathParts([$to, $target]);
                 if (!is_dir(dirname($targetPath))) {
                     mkdir(dirname($targetPath), 0777, true);
-                    $this->write(sprintf(
-                        'Created directory <fg=green>"%s"</>',
-                        $this->relativizePath($targetPath)
-                    ));
+                    $this->write(sprintf('Created directory <fg=green>"%s"</>', $this->relativizePath($targetPath)));
                 }
 
                 if (!file_exists($targetPath)) {
-                    $this->copyFile(
-                        $this->concatenatePathParts([$from, $source]),
-                        $targetPath
-                    );
+                    $this->copyFile($this->concatenatePathParts([$from, $source]), $targetPath);
                 }
             }
         }
@@ -66,16 +57,10 @@ class CopyFromPackageConfigurator extends AbstractConfigurator
         foreach ($manifest as $source => $target) {
             $targetPath = $this->concatenatePathParts([$to, $target]);
             if ('/' === substr($source, -1)) {
-                $this->removeFilesFromDir(
-                    $this->concatenatePathParts([$from, $source]),
-                    $this->concatenatePathParts([$to, $target])
-                );
+                $this->removeFilesFromDir($this->concatenatePathParts([$from, $source]), $this->concatenatePathParts([$to, $target]));
             } else {
                 @unlink($targetPath);
-                $this->write(sprintf(
-                    'Removed file <fg=green>"%s"</>',
-                    $this->relativizePath($targetPath)
-                ));
+                $this->write(sprintf('Removed file <fg=green>"%s"</>', $this->relativizePath($targetPath)));
             }
         }
     }
@@ -86,19 +71,13 @@ class CopyFromPackageConfigurator extends AbstractConfigurator
             mkdir($target, 0777, true);
         }
 
-        $iterator = new \RecursiveIteratorIterator(
-            new \RecursiveDirectoryIterator($source, \RecursiveDirectoryIterator::SKIP_DOTS),
-            \RecursiveIteratorIterator::SELF_FIRST
-        );
+        $iterator = new \RecursiveIteratorIterator(new \RecursiveDirectoryIterator($source, \RecursiveDirectoryIterator::SKIP_DOTS), \RecursiveIteratorIterator::SELF_FIRST);
         foreach ($iterator as $item) {
             $targetPath = $this->concatenatePathParts([$target, $iterator->getSubPathName()]);
             if ($item->isDir()) {
                 if (!is_dir($targetPath)) {
                     mkdir($targetPath);
-                    $this->write(sprintf(
-                        'Created directory <fg=green>"%s"</>',
-                        $this->relativizePath($targetPath)
-                    ));
+                    $this->write(sprintf('Created directory <fg=green>"%s"</>', $this->relativizePath($targetPath)));
                 }
             } elseif (!file_exists($targetPath)) {
                 $this->copyFile($item, $targetPath);
@@ -114,33 +93,21 @@ class CopyFromPackageConfigurator extends AbstractConfigurator
 
         copy($source, $target);
         @chmod($target, fileperms($target) | (fileperms($source) & 0111));
-        $this->write(sprintf(
-            'Created file <fg=green>"%s"</>',
-            $this->relativizePath($target)
-        ));
+        $this->write(sprintf('Created file <fg=green>"%s"</>', $this->relativizePath($target)));
     }
 
     private function removeFilesFromDir(string $source, string $target)
     {
-        $iterator = new \RecursiveIteratorIterator(
-            new \RecursiveDirectoryIterator($source, \RecursiveDirectoryIterator::SKIP_DOTS),
-            \RecursiveIteratorIterator::CHILD_FIRST
-        );
+        $iterator = new \RecursiveIteratorIterator(new \RecursiveDirectoryIterator($source, \RecursiveDirectoryIterator::SKIP_DOTS), \RecursiveIteratorIterator::CHILD_FIRST);
         foreach ($iterator as $item) {
             $targetPath = $this->concatenatePathParts([$target, $iterator->getSubPathName()]);
             if ($item->isDir()) {
                 // that removes the dir only if it is empty
                 @rmdir($targetPath);
-                $this->write(sprintf(
-                    'Removed directory <fg=green>"%s"</>',
-                    $this->relativizePath($targetPath)
-                ));
+                $this->write(sprintf('Removed directory <fg=green>"%s"</>', $this->relativizePath($targetPath)));
             } else {
                 @unlink($targetPath);
-                $this->write(sprintf(
-                    'Removed file <fg=green>"%s"</>',
-                    $this->relativizePath($targetPath)
-                ));
+                $this->write(sprintf('Removed file <fg=green>"%s"</>', $this->relativizePath($targetPath)));
             }
         }
     }
--- a/src/Configurator/CopyFromPackageConfigurator.php
+++ b/src/Configurator/CopyFromPackageConfigurator.php
@@ -37,14 +37,25 @@ class CopyFromPackageConfigurator extends AbstractConfigurator
         foreach ($manifest as $source => $target) {
             $target = $this->options->expandTargetDir($target);
             if ('/' === substr($source, -1)) {
-                $this->copyDir($from.'/'.$source, $to.'/'.$target);
+                $this->copyDir(
+                    $this->concatenatePathParts([$from, $source]),
+                    $this->concatenatePathParts([$to, $target])
+                );
             } else {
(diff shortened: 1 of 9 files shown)


szymach:
@fabpot let me know if this is OK or should I change something, I will still be working on this anyway.

nicolas-grekas:
Could you post a screenshot please?

src/Configurator/CopyFromRecipeConfigurator.php
```diff
@@ -66,6 +66,8 @@ private function copyFile(string $to, string $contents, bool $executable)
         if ($executable) {
             @chmod($to, fileperms($to) | 0111);
         }
+
+        $this->write('Copying file "'. basename($to) .'" to "'. $to.'"');
```
fabpot: I would reuse the work done in https://github.com/sensiolabs/SensioGeneratorBundle/pull/397 (probably to be adapted) for the way we display the information. See https://github.com/sensiolabs/SensioGeneratorBundle/pull/397/files#diff-0e445b5f16f472d999100a19e18c4e33R73 specifically.

szymach:
How about this:

install
remove

Not sure about the double slash on the file path - I do not think it is a result of my changes.

szymach:
@nicolas-grekas @fabpot I have performed the requested changes and posted a screenshot of two packages being added and removed (see above). Please tell me if this is OK and if it is, I will continue the work on this.

src/Configurator/CopyFromRecipeConfigurator.php
```diff
@@ -66,6 +66,8 @@ private function copyFile(string $to, string $contents, bool $executable)
         if ($executable) {
             @chmod($to, fileperms($to) | 0111);
         }
+
+        $this->write(sprintf('Created file <fg=green>"%s"</>', $this->relativizePath($to)));
```
fabpot: I would remove "file".

src/Configurator/CopyFromRecipeConfigurator.php
```diff
@@ -93,9 +95,17 @@ private function removeFiles(array $manifest, array $files, string $to)
     private function removeFile(string $to)
     {
         @unlink($to);
+        $this->write(sprintf('Removed file <fg=green>"%s"</>', $this->relativizePath($to)));
```
fabpot: I would remove "file".

fabpot:
Fixing the double-slash would be a nice bonus :)

I cannot see the "Removing entries in ..." in your PR, should probably be "Removed entries in ..." to be consistent with the other messages.

szymach:
You mean the [.gitignore](https://github.com/symfony/flex/blob/master/src/Configurator/GitignoreConfigurator.php#L49) messages? Was not added by me, but I can adjust my messages to say "Creating / Removing" if we are aiming for consistency. I will also fix that double slash.

Anyways, I did the minimum set of changes to see if I am going in the right direction, so I have only added these two messages. I still have more to go, which is removing directory message and the whole `CopyFromPackageConfigurator`.

fabpot:
The removing message indeed already exists. It shows up in your screenshot. And indeed, I think changing it is required for more consistency.

I think you are indeed in the right direction.

src/Configurator/CopyFromPackageConfigurator.php
```diff
@@ -53,52 +64,98 @@ private function copyFiles(array $manifest, string $from, string $to)
...
-            } elseif (!file_exists($target.'/'.$iterator->getSubPathName())) {
-                $this->copyFile($item, $target.'/'.$iterator->getSubPathName());
+            } elseif (!file_exists($targetPath)) {
+                $this->copyFile($item, $targetPath);
             }
         }
     }
 
-    public function copyFile(string $source, string $target)
+    public function copyFile(string $source, string $target): void
```
szymach: Should this method really be public? I have found no usage for it outside of the this class.

nicolas-grekas: it is now, so better not break it - but "void" must be removed: Flex is 7.0-compatible

src/Configurator/CopyFromPackageConfigurator.php
```diff
@@ -85,20 +111,51 @@ public function copyFile(string $source, string $target)
         if (file_exists($target)) {
             return;
         }
+
         copy($source, $target);
         @chmod($target, fileperms($target) | (fileperms($source) & 0111));
+        $this->write(sprintf(
+            'Created file <fg=green>"%s"</>',
+            $this->relativizePath($target)
+        ));
```
fabpot: Could you move this on one line? We rarely split a line (that is not important) on several lines.

src/Configurator/CopyFromPackageConfigurator.php
```diff
@@ -85,20 +111,51 @@ public function copyFile(string $source, string $target)
...
     }
 
     private function removeFilesFromDir(string $source, string $target)
     {
-        $iterator = new \RecursiveIteratorIterator(new \RecursiveDirectoryIterator($source, \RecursiveDirectoryIterator::SKIP_DOTS), \RecursiveIteratorIterator::CHILD_FIRST);
+        $iterator = new \RecursiveIteratorIterator(
+            new \RecursiveDirectoryIterator($source, \RecursiveDirectoryIterator::SKIP_DOTS),
+            \RecursiveIteratorIterator::CHILD_FIRST
+        );
```
fabpot: Same here, one line is enough.

src/Configurator/CopyFromPackageConfigurator.php
```diff
@@ -85,20 +111,51 @@ public function copyFile(string $source, string $target)
...
         foreach ($iterator as $item) {
+            $targetPath = $this->concatenatePathParts([$target, $iterator->getSubPathName()]);
             if ($item->isDir()) {
                 // that removes the dir only if it is empty
-                @rmdir($target.'/'.$iterator->getSubPathName());
+                @rmdir($targetPath);
+                $this->write(sprintf(
+                    'Removed directory <fg=green>"%s"</>',
+                    $this->relativizePath($targetPath)
+                ));
```
fabpot: Here as well.

src/Configurator/CopyFromPackageConfigurator.php
```diff
@@ -85,20 +111,51 @@ public function copyFile(string $source, string $target)
...
             } else {
-                @unlink($target.'/'.$iterator->getSubPathName());
+                @unlink($targetPath);
+                $this->write(sprintf(
+                    'Removed file <fg=green>"%s"</>',
+                    $this->relativizePath($targetPath)
+                ));
```
fabpot: and here

src/Configurator/CopyFromRecipeConfigurator.php
```diff
@@ -35,9 +35,13 @@ private function copyFiles(array $manifest, array $files, string $to)
...
             if ('/' === substr($source, -1)) {
-                $this->copyDir($source, $to.'/'.$target, $files);
+                $this->copyDir($source, $this->concatenatePathParts([$to, $target]), $files);
             } else {
-                $this->copyFile($to.'/'.$target, $files[$source]['contents'], $files[$source]['executable']);
+                $this->copyFile(
+                    $this->concatenatePathParts([$to, $target]),
+                    $files[$source]['contents'],
+                    $files[$source]['executable']
+                );
```
fabpot: and here

src/Configurator/CopyFromRecipeConfigurator.php
```diff
@@ -81,21 +87,42 @@ private function removeFiles(array $manifest, array $files, string $to)
...
                 foreach (array_keys($files) as $file) {
                     if (0 === strpos($file, $source)) {
-                        $this->removeFile($to.'/'.$target.'/'.substr($file, strlen($source)));
+                        $this->removeFile(
+                            $this->concatenatePathParts([
+                                $to,
+                                $target,
+                                substr($file, strlen($source))
+                            ])
+                        );
```
fabpot: And here

tests/Configurator/CopyFromRecipeConfiguratorTest.php
```diff
@@ -0,0 +1,101 @@
...
+        @unlink($this->targetFile);
+    }
+
+    private function createConfigurator(): CopyFromRecipeConfigurator
+    {
+        return new CopyFromRecipeConfigurator(
+            $this->getMockBuilder(Composer::class)->getMock(),
+            $this->io,
+            new Options()
+        );
```
fabpot: And here.

src/Configurator/CopyFromPackageConfigurator.php
szymach: Yes I have already removed it, this is just an old diff.

src/Configurator/CopyFromPackageConfigurator.php
```diff
@@ -85,20 +111,51 @@ public function copyFile(string $source, string $target)
...
             }
         }
     }
+
+    private function relativizePath(string $absolutePath): string
+    {
+        $relativePath = str_replace(getcwd(), '.', $absolutePath);
+
+        return is_dir($absolutePath) ? rtrim($relativePath, '/').'/' : $relativePath;
```
stof: this requires being tested on Windows (which has a different path separator). I'm almost sure your implementation is broken currently.

src/Configurator/CopyFromPackageConfigurator.php
```diff
@@ -85,20 +111,51 @@ public function copyFile(string $source, string $target)
...
+    }
+
+    private function concatenatePathParts(array $parts): string
```
stof: Can we make them shared utilities instead of duplicating them in each configurator ? It would avoid having to maintain them multiple times (especially once we make them more complex to support Windows paths too)

src/Configurator/CopyFromPackageConfigurator.php
szymach: It is a direct copy from [SensioGeneratorBundle](https://github.com/sensiolabs/SensioGeneratorBundle/pull/397/files#diff-0e445b5f16f472d999100a19e18c4e33R102). Did it not work on Windows?

src/Configurator/CopyFromPackageConfigurator.php
```diff
@@ -55,13 +55,14 @@ private function copyFiles(array $manifest, string $from, string $to)
...
+            $targetPath = $this->filePathUtil->concatenatePathParts([$to, $target]);
             if ('/' === substr($source, -1)) {
-                $this->removeFilesFromDir($this->concatenatePathParts([$from, $source]), $this->concatenatePathParts([$to, $target]));
+                $this->removeFilesFromDir($this->filePathUtil->concatenatePathParts([$from, $source]), $this->filePathUtil->concatenatePathParts([$to, $target]));
             } else {
                 @unlink($targetPath);
-                $this->write(sprintf('Removed file <fg=green>"%s"</>', $this->relativizePath($targetPath)));
+                $this->write(sprintf('Removed <fg=green>"%s"</>', $this->filePathUtil->relativizePath($targetPath)));
             }
+            $this->removeDirectoryIfEmpty(dirname($targetPath));
```
szymach: I have added this because as far as I can tell, you can have empty directories left over. I can remove this if need be.

szymach:
@stof @fabpot I have added requested changes. If you find it OK, I will add more tests and the verbosity check.

src/FilePathUtility.php
```diff
@@ -0,0 +1,36 @@
...
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Flex;
+
+class FilePathUtility
```
fabpot: A suggestion: rename `FilePathUtility` to `Path`. Then rename `relativizePath` to `relativize` and `concatenatePathParts` to `concatenate`. I think the code would be easier to read: `$path->relativize(...)`. WDYT?

fabpot: This class should be tagged as `@internal`

szymach: Sounds good.

szymach:
@stof [this bit here](https://github.com/symfony/flex/blob/master/src/Configurator/AbstractConfigurator.php#L47) only outputs the messages when the `verbose` flag is on, do you want me to check different levels of verbosity as well?
@fabpot

src/Configurator/CopyFromPackageConfigurator.php
```diff
@@ -85,20 +91,34 @@ public function copyFile(string $source, string $target)
         if (file_exists($target)) {
             return;
         }
+
+        if (!file_exists($source)) {
+            throw new Exception(sprintf('File "%s" does not exist!', $source));
```
szymach: I have added this bit here, since it seemed sensible to throw an exception if a source file is missing.

fabpot: I would have used `\LogicException` here as it should never happen.

szymach:
ping @fabpot @stof

szymach:
ping @fabpot @stof

fabpot:
Thank you @szymach.

nicolas-grekas:
Breaks in #25817?

PR #271, opened 2018-01-07
require ext-ctype

### LINKED ISSUE #269: composer should check for required extension `ctype`
Hi! 
Nothing terrible, but I thought it would worth it to add an issue to not forget.

Currently getting this error: 

```
Fatal error: Uncaught Error: Call to undefined function Symfony\Flex\Configurator\ctype_digit() in /usr/src/app/example/sample-app/vendor/symfony/flex/src/Configurator/EnvConfigurator.php:47
Stack trace:
#0 /usr/src/app/example/sample-app/vendor/symfony/flex/src/Configurator/EnvConfigurator.php(25): Symfony\Flex\Configurator\EnvConfigurator->configureEnvDist(Object(Symfony\Flex\Recipe), Array)
#1 /usr/src/app/example/sample-app/vendor/symfony/flex/src/Configurator.php(52): Symfony\Flex\Configurator\EnvConfigurator->configure(Object(Symfony\Flex\Recipe), Array)
#2 /usr/src/app/example/sample-app/vendor/symfony/flex/src/Flex.php(235): Symfony\Flex\Configurator->install(Object(Symfony\Flex\Recipe))
#3 [internal function]: Symfony\Flex\Flex->update(Object(Composer\Script\Event))
#4 phar:///usr/local/bin/composer/src/Composer/EventDispatcher/EventDispatcher.php(171): call_user_func(Array, Object(Composer\Script\Event))
#5 phar:///usr/local/bin/composer/src/Composer/EventDispatcher/EventDispatcher.php(96): Composer\EventDispatc in /usr/src/app/example/sample-app/vendor/symfony/flex/src/Configurator/EnvConfigurator.php on line 47
/usr/src/app/example/sample-app #
```

docteurklein:
fixes #269

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -10,7 +10,8 @@
     ],
     "require": {
         "php": "^7.0",
-        "composer-plugin-api": "^1.1"
+        "composer-plugin-api": "^1.1",
+        "ext-ctype": "^7.0"
     },
     "require-dev": {
         "composer/composer": "^1.4",


nicolas-grekas:
Looking at the code base, ctype_digit is used once.
Shouldn't we replace it by a regexp instead?

docteurklein:
That might be even better indeed. 

Now I think most of the users have `ctype` installed, but depending on the distro you're using, it might not (alpine f.e).

stof:
The requirement-checker defines ctype as required to run Symfony, so is it actually an issue if Flex relies on it too ?

Btw, we might want to reference more extensions directly in composer.json for Symfony components (currently, the requirement checker defines ctype, json, iconv and a few others as requirements of a Symfony-compatible system)

fabpot:
What about #282 instead?

PR #272, opened 2018-01-08
Dispatch PluginEvents::PRE_FILE_DOWNLOAD when calling endpoint

nicolas-grekas:
By leveraging this hook + https://github.com/composer/composer/pull/6985, I'm able to share one connection+SSL context, reducing round-trip latency (next PR.)


fabpot:
Thank you @nicolas-grekas.

PR #276, opened 2018-01-08
Fix authentication issue on prefetch

### LINKED ISSUE #273: Authentication failed since 1.0.54 (Private Satis with Basic-Auth )
Since the 1.0.54 release the install/requere command ask for authentication. We use a private Satis instance secured by Basic-Auth.
screen-001

The data ist stored in the auth.json an works well with Flex < 1.0.53. Entering the username and password in the CLI fails also.

I think the symfony/thanks component try to connect the Satis.

nicolas-grekas:
Fixes #273

### DIFF
--- a/src/ParallelDownloader.php
+++ b/src/ParallelDownloader.php
@@ -59,19 +59,19 @@ class ParallelDownloader extends RemoteFilesystem
                 continue;
             }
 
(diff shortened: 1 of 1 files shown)


fabpot:
Thank you @nicolas-grekas.

PR #280, opened 2018-01-09
Workaround: Avoid emojis which cause terminal failures in Docker

DHager:
Sending certain unicode values to STDOUT may cause Docker terminals to abruptly quit and disconnect for some users, as noted in docker/toolbox#695  and moby/moby#22345. In this particular case the issue is triggered by the emoji star (U+2B50) which was added in #262 / 0a45ec98fe91340001746baa1336c5d9d827f938 . 

While this is **not a bug in Flex**, I still recommend skipping the cosmetic emojis, since their appeal is probably outweighed by the impact and users and the potential that some users will incorrectly blame Flex when `composer update` disconnects their terminal halfway through the "thanks reminder."

### DIFF
--- a/src/Flex.php
+++ b/src/Flex.php
@@ -171,12 +171,9 @@ class Flex implements PluginInterface, EventSubscriberInterface
         }
 
         if ($this->displayThanksReminder) {
-            $love = '\\' === DIRECTORY_SEPARATOR ? 'love' : '💖 ';
-            $star = '\\' === DIRECTORY_SEPARATOR ? 'star' : '⭐ ';
-
             $this->io->writeError('');
             $this->io->writeError('What about running <comment>composer global require symfony/thanks && composer thanks</> now?');
-            $this->io->writeError(sprintf('This will spread some %s by sending a %s to the GitHub repositories of your fellow package maintainers.', $love, $star));
+            $this->io->writeError('This will spread some love by sending a star to the GitHub repositories of your fellow package maintainers.');
             $this->io->writeError('');
         }
 


nicolas-grekas:
I would prefer not doing this. The world is enough boring to work harder to make it fun :)
Can't we detect this case where we cannot use the emoji instead?

javiereguiluz:
I agree with @DHager here. Why don't we remove the emojis and always display the same chars as on Windows? We can't break something as important as Docker with something that it's optional.

nicolas-grekas:
@javiereguiluz we already do this on Windows. This patch will remove the emoji on ALL platforms, just because Docker on Windows breaks (note that "Docker on Windows" is not detected as "on Windows" by the current logic.)

DHager:
@nicolas-grekas In this scenario, the PHP code is running on a Linux container, sectioned off inside a Linux virtual machine, which in turn is running within a Windows physical host. 

Even if you could detect the situation, that's probably the kind of stuff which should exist in a separate library or utility.

bitgandtter:
I agree with @nicolas-grekas on this one, the issue on this matter is on docker and they have the way to fix it, we should not make flex boring IMHO

DHager:
@bitgandtter When I look at `composer update`, I see a scriptable command-line cross-plaform package-management tool... From my perspective, all of that stuff is *supposed* to be safely-boring (perhaps with stodgy phrases like "POSIX compliant.") 

Ultimately, it means everybody has *more* time for fun, because they aren't debugging a mysterious crash or getting paged after-hours when something breaks.

fabpot:
Thank you @DHager.

fabpot:
@DHager Can you submit the same PR here? https://github.com/symfony/thanks/blob/b637913a92664571afd1067fc645ecb30c9cf5b2/src/Thanks.php

nicolas-grekas:
There is another emoji in the ParallelDownloader class also. @DHager I would also really appreciate if you could help us find a workaround that would allow reusing emojis on platform where it works. Eg detecting the faulty terminal.

nicolas-grekas:
@DHager no need, see https://github.com/symfony/thanks/pull/34

nicolas-grekas:
Reverted in @DHager, which alternative emoji used.

PR #281, opened 2018-01-09
Fix backtrace-related initialization logic

nicolas-grekas:
The previous logic base on the `$search = 3;` counter doesn't work because eg `Application` is found several times in the trace, leading to several run of the same logic.

Getting the `ArgvInput` instance is also easier when looking at trace's arguments.

The remaining are CS changes found by php-cs-fixer.


fabpot:
Thank you @nicolas-grekas.

PR #286, opened 2018-01-10
Leverage ext-curl when possible to achieve high speed downloads

nicolas-grekas:
This PR brings hirak/prestissimo-like download performance to Flex, but with greater compatibility with Composer's native behavior, thanks to a new extension point that is going to be part of Composer 1.7 (https://github.com/composer/composer/pull/6985).

Meanwhile, it already provides concurrent connections, and auto-enables the curl-based downloader conditionally.

In short, it should be totally seamless.

### DIFF
--- a/src/ParallelDownloader.php
+++ b/src/ParallelDownloader.php
@@ -12,11 +12,8 @@
 namespace Symfony\Flex;
 
 use Composer\Config;
-use Composer\Downloader\FileDownloader;
 use Composer\Downloader\TransportException;
-use Composer\Factory;
 use Composer\IO\IOInterface;
-use Composer\Package\PackageInterface;
 use Composer\Util\RemoteFilesystem;
 
 /**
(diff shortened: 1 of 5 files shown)


src/ParallelDownloader.php
```diff
@@ -12,104 +12,94 @@
...
+
+    public function setNextOptions(array $options)
+    {
+        $this->nextOptions = parent::getOptions() !== $options ? $options : [];
+
+        return $this;
     }
 
-    protected function callbackGet($notificationCode, $severity, $message, $messageCode, $bytesTransferred, $bytesMax)
+    public function callbackGet($notificationCode, $severity, $message, $messageCode, $bytesTransferred, $bytesMax)
```
fabpot: Should we tag it as being `@internal`?

nicolas-grekas:
PR read, works GREAT!

nicolas-grekas:
PR ready for another try. All good on my side. Now able to do concurrent downloads on more files than prestissimo. As such, prestissimo is now disabled if found.

fabpot:
Thank you @nicolas-grekas.

PR #288, opened 2018-01-12
Reword post-install comment

thewilkybarkid:
"Don't hesitate to" is common but has a bit of a negative vibe.

(Could also just start with "Review".)


fabpot:
Thank you @thewilkybarkid.

PR #293, opened 2018-01-14
Sort bundles.php alphabetically

mhujer:
Generated `bundles.php` should be sorted alphabetically to prevent unnecessary merge-conflicts. 
`symfony.lock` and `composer.json` are already sorted automatically.

### DIFF
--- a/src/Configurator/BundlesConfigurator.php
+++ b/src/Configurator/BundlesConfigurator.php
@@ -78,6 +78,10 @@ class BundlesConfigurator extends AbstractConfigurator
     private function dump(string $file, array $bundles)
     {
         $contents = "<?php\n\nreturn [\n";
+
+        // sort bundles alphabetically to prevent unnecessary merge-conflicts
+        ksort($bundles);
+
         foreach ($bundles as $class => $envs) {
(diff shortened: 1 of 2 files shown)


nicolas-grekas:
Hello @mhujer and thanks for this PR.
This has already been proposed in the past, and rejected (eg https://github.com/symfony/flex/pull/123)
The reason is that sometimes, the order in which bundles are loaded is of importance, so we cannot do this without risking breaking apps.

mhujer:
Thanks for the link to the PR, I didn't find it before!

tomasfejfar:
Wouldn't it be best to add dependencies to flex? While the order is meaningful, it's currently arbitrary. Making it explicit would be another step towards better DX, don't you think?

stof:
Well, for now, Flex adds the new bundle at the end when installing it, and leaves everything else untouched. This allows you to put bundles in the order you want (by editing the file to change the order).
Applying the sorting breaks this, as any manual edition would become lost the next time Flex registers a bundle.

tomasfejfar:
Oh, I forgot that some bundles may not be installed by flex and the growing complexity regarding that. Thanks.

PR #303, opened 2018-01-19
Add compatibility with composer-plugin-api ^1.0

nicolas-grekas:
Let's remove this friction point, especially when Ubuntu's Composer is v1.0.2. Will save a few blocking steps for new comers.


nicolas-grekas:
PR ready - all features except "unpack" work on 1.0. A warning is displayed when an old version of composer is used.

image

or 

image

javiereguiluz:
Could we display that warning message only when the user tries to execute an unsupported feature?

nicolas-grekas:
We cannot : if we were able to hook there, we would be able to implement the unsupported features :)

fabpot:
Thank you @nicolas-grekas.

PR #304, opened 2018-01-19
Display thanks reminder only when a package is actually updated

nicolas-grekas:
Right now, the reminder is displayed when running composer update with no composer.lock file.
This was not intended. Here is what was to me :)

Sibling to https://github.com/symfony/thanks/pull/38


fabpot:
Thank you @nicolas-grekas.

PR #305, opened 2018-01-19
Fix BC with Composer 1.0

nicolas-grekas:
Now fixes 3 issues:
- BC with Composer 1.0
- Flex aliases with Composer 1.1
- thanks reminder should need package AND command update

Best reviewed compared to v1.0.63: https://github.com/symfony/flex/compare/v1.0.63...54d6a5dc4a3e6cf9338d46ed51fdf7ea4efef78d

### DIFF
--- a/src/Command/RemoveCommand.php
+++ b/src/Command/RemoveCommand.php
@@ -1,0 +1,36 @@
+<?php
+
+/*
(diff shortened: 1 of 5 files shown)


fabpot:
Thank you @nicolas-grekas.

PR #308, opened 2018-01-21
Fix command registration

### LINKED ISSUE #307: Command "unpack" is not defined
After running `composer require debug`, when I run `composer unpack debug` I get:

```
[Symfony\Component\Console\Exception\CommandNotFoundException]
Command "unpack" is not defined.
```

nicolas-grekas:
Fix #307


fabpot:
Thank you @nicolas-grekas.

PR #309, opened 2018-01-25
Fix Composer 1.0 incompatibility

nicolas-grekas:
Fix https://github.com/symfony/skeleton/issues/55


fabpot:
Thank you @nicolas-grekas.

PR #312, opened 2018-01-31
Check max host connections option is available

PohlmanUSA:
This option was added in libcurl 7.30 which is not available by default on RHEL distros. 

Added a check before setting this option.


fabpot:
Thank you @PohlmanUSA.

PR #314, opened 2018-02-01
Fix cURL error handling

nicolas-grekas:
Fix https://github.com/symfony/thanks/issues/42


fabpot:
Thank you @nicolas-grekas.

PR #319, opened 2018-02-17
simplify matching for showing makefile targets

SimonHeimberg:
does not use grep anymore, awk can do this as well


fabpot:
Thank you @SimonHeimberg.

Issue #325, opened 2018-02-21
CopyFromPackage unconfigure not removing files


lopezator:
CopyFromPackageConfigurator isn't removing files when using special values like %PUBLIC_DIR%, %BIN_DIR%, etc.

Example in symfony-requirements recipe:

https://github.com/symfony/recipes/blob/master/symfony/requirements-checker/1.0/manifest.json

Steps to reproduce, in a fresh Symfony4 install:

$ composer require requirements-checker

The recipe copies the file check.php into public folder as expected.

$ composer remove requirements-checker

The file check.php into public folder isn't removed.

The target dir isn't being expanded in the removeFiles function, so it checks the existence of file using the the path, "%PUBLIC_DIR%/check.php" literally.

PR #326, opened 2018-02-21
bug symfony/flex#325 CopyFromPackage unconfigure is not removing files

lopezator:
bug symfony/flex#325 CopyFromPackage unconfigure is not removing files when using special directories like %PUBLIC_DIR%


lopezator:
Why the checks have failed? any clue? @fabpot

fabpot:
Thank you @lopezator.

PR #327, opened 2018-03-02
support for custom symfony.lock file path

hjanuschka:
we are using 2 different composer.json files. to  handle beta and master builds.

e.g:

installing beta dependencies:
`COMPOSER=composer.beta.json composer install`

`symfony.lock` is not really capable of handling a situation like this.
this PR adds a env `SYMFONY_LOCKFILE_PATH`

this way one can set the name/path of the symfony.lock file

for us it results in:
`COMPOSER=composer.beta.json SYMFONY_LOCKFILE_PATH=symfony.beta.lock composer install`

and for master/prod
`composer install`

let me know if you want me to change anything in this PR.
keep up the good work ❤️

### DIFF
--- a/src/Flex.php
+++ b/src/Flex.php
@@ -101,11 +101,7 @@ class Flex implements PluginInterface, EventSubscriberInterface
         $this->configurator = new Configurator($composer, $io, $this->options);
         $this->downloader = new Downloader($composer, $io, $this->rfs);
         $this->downloader->setFlexId($this->getFlexId());
-        $lock_file_path = str_replace(Factory::getComposerFile(), 'composer.json', 'symfony.lock');
-        if (getenv("SYMFONY_LOCKFILE_PATH")) {
-            $lock_file_path  = getenv("SYMFONY_LOCKFILE_PATH");
-        }
-        $this->lock = new Lock($lock_file_path);
+        $this->lock = new Lock($this->getLockFilePath()->symfony);
 
         $populateRepoCacheDir = __CLASS__ === self::class;
         if ($composer->getPluginManager()) {
@@ -186,14 +182,24 @@ class Flex implements PluginInterface, EventSubscriberInterface
         }
     }
 
-    public function composerLockMissing()
+    public function getLockFilePath()
     {
-        if (getenv("COMPOSER")) {
-            $composer_file = getenv("COMPOSER");
-            $composer_lock = str_replace(".json", ".lock", $composer_file);
-            return file_exists($composer_file) && !file_exists($composer_lock);
+        $composerFile = Factory::getComposerFile();
+        $lockFilePath = "json" === pathinfo($composerFile, PATHINFO_EXTENSION)
+                ? substr($composerFile, 0, -4).'lock'
+                :  $composerFile .  '.lock';
+        $symfonyLock = str_replace("composer.", "symfony.", $lockFilePath);
+        if (getenv("SYMFONY_LOCKFILE_PATH")) {
+            $symfonyLock  = getenv("SYMFONY_LOCKFILE_PATH");
         }
-        return file_exists('composer.json') && !file_exists('composer.lock');
+        return (object)[
+            "composer" => $lockFilePath,
+            "symfony" => $symfonyLock
+        ];
+    }
+    public function isComposerLockMissing()
+    {
+        return file_exists(Factor::getComposerFile()) && !file_exists($this->getLockFilePath()->composer);
     }
     public function configureProject(Event $event)
     {
--- a/src/Flex.php
+++ b/src/Flex.php
@@ -182,7 +182,7 @@ class Flex implements PluginInterface, EventSubscriberInterface
         }
     }
 
-    public function getLockFilePath()
+    private function getLockFilePath()
     {
         $composerFile = Factory::getComposerFile();
         $lockFilePath = "json" === pathinfo($composerFile, PATHINFO_EXTENSION)
@@ -197,7 +197,7 @@ class Flex implements PluginInterface, EventSubscriberInterface
             "symfony" => $symfonyLock
         ];
     }
-    public function isComposerLockMissing()
+    private function isComposerLockMissing()
     {
         return file_exists(Factor::getComposerFile()) && !file_exists($this->getLockFilePath()->composer);
     }
--- a/src/Flex.php
+++ b/src/Flex.php
@@ -101,7 +101,7 @@ class Flex implements PluginInterface, EventSubscriberInterface
         $this->configurator = new Configurator($composer, $io, $this->options);
         $this->downloader = new Downloader($composer, $io, $this->rfs);
         $this->downloader->setFlexId($this->getFlexId());
-        $this->lock = new Lock($this->getLockFilePath()->symfony);
+        $this->lock = new Lock($this->getSymfonyLockFilePath());
 
         $populateRepoCacheDir = __CLASS__ === self::class;
         if ($composer->getPluginManager()) {
@@ -182,25 +182,28 @@ class Flex implements PluginInterface, EventSubscriberInterface
         }
     }
 
-    private function getLockFilePath()
+    private function getComposerLockFilePath()
     {
         $composerFile = Factory::getComposerFile();
         $lockFilePath = "json" === pathinfo($composerFile, PATHINFO_EXTENSION)
                 ? substr($composerFile, 0, -4).'lock'
                 :  $composerFile .  '.lock';
+        return $lockFilePath;
+    }
+
+    private function getSymfonyLockFilePath()
+    {
+        $lockFilePath = $this->getComposerLockFilePath();
         $symfonyLock = str_replace("composer.", "symfony.", $lockFilePath);
         if (getenv("SYMFONY_LOCKFILE_PATH")) {
             $symfonyLock  = getenv("SYMFONY_LOCKFILE_PATH");
         }
-        return (object)[
-            "composer" => $lockFilePath,
-            "symfony" => $symfonyLock
-        ];
+        return $symfonyLock;
     }
 
     private function isComposerLockMissing()
     {
-        return file_exists(Factory::getComposerFile()) && !file_exists($this->getLockFilePath()->composer);
+        return file_exists(Factory::getComposerFile()) && !file_exists($this->getComposerLockFilePath());
     }
 
     public function configureProject(Event $event)
--- a/src/Flex.php
+++ b/src/Flex.php
@@ -101,7 +101,11 @@ class Flex implements PluginInterface, EventSubscriberInterface
         $this->configurator = new Configurator($composer, $io, $this->options);
         $this->downloader = new Downloader($composer, $io, $this->rfs);
         $this->downloader->setFlexId($this->getFlexId());
-        $this->lock = new Lock(str_replace(Factory::getComposerFile(), 'composer.json', 'symfony.lock'));
+        $lock_file_path = str_replace(Factory::getComposerFile(), 'composer.json', 'symfony.lock');
+        if(getenv("SYMFONY_LOCKFILE_PATH")) {
+          $lock_file_path  = getenv("SYMFONY_LOCKFILE_PATH");
+        }
+        $this->lock = new Lock($lock_file_path);
 
         $populateRepoCacheDir = __CLASS__ === self::class;
         if ($composer->getPluginManager()) {
--- a/src/Flex.php
+++ b/src/Flex.php
@@ -169,7 +169,7 @@ class Flex implements PluginInterface, EventSubscriberInterface
                 }
             }
 
-            if ($populateRepoCacheDir && isset(self::$repoReadingCommands[$command]) && ('install' !== $command || ($this->composerLockMissing()))) {
+            if ($populateRepoCacheDir && isset(self::$repoReadingCommands[$command]) && ('install' !== $command || ($this->isComposerLockMissing()))) {
                 $this->populateRepoCacheDir();
             }
 
--- a/src/Flex.php
+++ b/src/Flex.php
@@ -197,10 +197,12 @@ class Flex implements PluginInterface, EventSubscriberInterface
             "symfony" => $symfonyLock
         ];
     }
+
     private function isComposerLockMissing()
     {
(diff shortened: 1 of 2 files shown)


src/Flex.php
```diff
@@ -101,7 +101,11 @@ public function activate(Composer $composer, IOInterface $io)
         $this->configurator = new Configurator($composer, $io, $this->options);
         $this->downloader = new Downloader($composer, $io, $this->rfs);
         $this->downloader->setFlexId($this->getFlexId());
-        $this->lock = new Lock(str_replace(Factory::getComposerFile(), 'composer.json', 'symfony.lock'));
+        $lock_file_path = str_replace(Factory::getComposerFile(), 'composer.json', 'symfony.lock');
```
stof: please use camelCased variable names

src/Flex.php
```diff
@@ -182,6 +186,15 @@ public function activate(Composer $composer, IOInterface $io)
         }
     }
 
+    public function composerLockMissing()
+    {
+        if (getenv("COMPOSER")) {
```
stof: use `Factory::getComposerFile()` instead of duplicating its logic

src/Flex.php
```diff
@@ -182,6 +186,15 @@ public function activate(Composer $composer, IOInterface $io)
         }
     }
 
+    public function composerLockMissing()
```
Pierstoval: I'd suggest `isComposerLockMissing()` as method name instead

src/Flex.php
```diff
@@ -182,6 +186,15 @@ public function activate(Composer $composer, IOInterface $io)
...
+    {
+        if (getenv("COMPOSER")) {
+            $composer_file = getenv("COMPOSER");
+            $composer_lock = str_replace(".json", ".lock", $composer_file);
```
Pierstoval: Use camelCase var names here too

hjanuschka:
@stof @Pierstoval  many thx for taking your time reviewing this.
i addressed your feedback and also added a test to verify it works.

src/Flex.php
```diff
@@ -182,6 +182,25 @@ public function activate(Composer $composer, IOInterface $io)
         }
     }
 
+    public function getLockFilePath()
+    {
+        $composerFile = Factory::getComposerFile();
+        $lockFilePath = "json" === pathinfo($composerFile, PATHINFO_EXTENSION)
```
hjanuschka: looks magic? taken from https://github.com/composer/composer/blob/ea9b7ecbb05c4b4265b1e15d72b2695439fdc12a/src/Composer/Factory.php#L389

src/Flex.php
```diff
@@ -182,6 +182,25 @@ public function activate(Composer $composer, IOInterface $io)
         }
     }
 
+    public function getLockFilePath()
```
xabbuh: `private`?

src/Flex.php
```diff
@@ -182,6 +182,25 @@ public function activate(Composer $composer, IOInterface $io)
...
+        $symfonyLock = str_replace("composer.", "symfony.", $lockFilePath);
+        if (getenv("SYMFONY_LOCKFILE_PATH")) {
+            $symfonyLock  = getenv("SYMFONY_LOCKFILE_PATH");
+        }
+        return (object)[
+            "composer" => $lockFilePath,
+            "symfony" => $symfonyLock
+        ];
+    }
+    public function isComposerLockMissing()
```
xabbuh: same here

hjanuschka:
thx for the time @xabbuh  -  feedback addressed 👍

src/Flex.php
```diff
@@ -182,6 +182,25 @@ public function activate(Composer $composer, IOInterface $io)
...
+        $symfonyLock = str_replace("composer.", "symfony.", $lockFilePath);
+        if (getenv("SYMFONY_LOCKFILE_PATH")) {
+            $symfonyLock  = getenv("SYMFONY_LOCKFILE_PATH");
+        }
+        return (object)[
+            "composer" => $lockFilePath,
+            "symfony" => $symfonyLock
+        ];
+    }
+    private function isComposerLockMissing()
```
Pierstoval: Missing newline above this one

src/Flex.php
```diff
@@ -182,6 +182,25 @@ public function activate(Composer $composer, IOInterface $io)
...
+    {
+        return file_exists(Factor::getComposerFile()) && !file_exists($this->getLockFilePath()->composer);
+    }
     public function configureProject(Event $event)
```
Pierstoval: Missing newline above this one

tests/FlexTest.php
```diff
@@ -32,6 +32,26 @@
 
 class FlexTest extends TestCase
 {
+
```
Pierstoval: Unnecessary newline here

tests/FlexTest.php
```diff
@@ -78,6 +98,30 @@ public function getRecordTests()
...
+     * dataset1: nothing overriden, just verify default names
+     * dataset2: composer = default, custom symfony.lock path
+     * @testWith ["composer.beta.json","", "composer.beta.lock", "symfony.beta.lock"]
+     *           ["","", "./composer.lock", "./symfony.lock"]
+     *           ["","symfony.beta.lock", "./composer.lock", "symfony.beta.lock"]
+     */
+    public function testSymfonyLock($composerEnv, $symfonyEnv, $expectedComposer, $expectedSymfony)
+    {
+
+
```
Pierstoval: Unnecessary newlines here

tests/FlexTest.php
```diff
@@ -78,6 +98,30 @@ public function getRecordTests()
...
+        putenv("COMPOSER=" . $composerEnv);
+        putenv("SYMFONY_LOCKFILE_PATH=" . $symfonyEnv);
+
+        $flex = new Flex();
+        $r = $this->invokeMethod($flex, 'getLockFilePath', []);
+        $this->assertEquals($expectedComposer, $r->composer);
+        $this->assertEquals($expectedSymfony, $r->symfony);
+
+
```
Pierstoval: Unnecessary newline here

hjanuschka:
@Pierstoval feedback addressed

tests/FlexTest.php
```diff
@@ -78,6 +97,27 @@ public function getRecordTests()
...
+        putenv("SYMFONY_LOCKFILE_PATH=" . $symfonyEnv);
+
+        $flex = new Flex();
+        $r = $this->invokeMethod($flex, 'getLockFilePath', []);
+        $this->assertEquals($expectedComposer, $r->composer);
+        $this->assertEquals($expectedSymfony, $r->symfony);
+
+        putenv("COMPOSER=''");
+        putenv("SYMFONY_LOCKFILE_PATH=''");
+    }
```
Pierstoval: Missing newline after this one (I'm sorry for such reviews :rofl: )

hjanuschka: hahaha no worries, fixed! i am sorry for such laziness :D

hjanuschka:
@Pierstoval feedback adressed, i really hope it gets declined in the last minute 🤒

src/Flex.php
```diff
@@ -182,6 +182,27 @@ public function activate(Composer $composer, IOInterface $io)
...
+    private function getLockFilePath()
+    {
+        $composerFile = Factory::getComposerFile();
+        $lockFilePath = "json" === pathinfo($composerFile, PATHINFO_EXTENSION)
+                ? substr($composerFile, 0, -4).'lock'
+                :  $composerFile .  '.lock';
+        $symfonyLock = str_replace("composer.", "symfony.", $lockFilePath);
+        if (getenv("SYMFONY_LOCKFILE_PATH")) {
+            $symfonyLock  = getenv("SYMFONY_LOCKFILE_PATH");
+        }
```
Pierstoval: IMO if the env var is set you shouldn't even parse `$composerFile` , it saves perfs

tests/FlexTest.php
```diff
@@ -78,6 +97,28 @@ public function getRecordTests()
         ];
     }
 
+    /**
+     * dataset0: custom COMPOSER env is set, generates correct symfony lock
+     * dataset1: nothing overriden, just verify default names
+     * dataset2: composer = default, custom symfony.lock path
+     * @testWith ["composer.beta.json","", "composer.beta.lock", "symfony.beta.lock"]
```
Pierstoval: Why not use a classic data provider?

src/Flex.php
hjanuschka: i am not sure if i get this one a 100%.
computing the "right" lock file based on the `getComposerFile()`  - is needed anyway - for later usage in `isComposerLockMissing` - otherwise - it needs to be computed there, as it was prior to the PR.

maybe i am wrong?

tests/FlexTest.php
```diff
@@ -32,6 +32,25 @@
...
+    * taken from Juan Tremino's fantastic tutorial
+    * https://jtreminio.com/2013/03/unit-testing-tutorial-part-3-testing-protected-private-methods-coverage-reports-and-crap/
+    *
+    * @param object &$object    Instantiated object that we will run method on.
+    * @param string $methodName Method name to call
+    * @param array  $parameters Array of parameters to pass into method.
+    *
+    * @return mixed Method return.
+    */
+    public function invokeMethod(&$object, $methodName, array $parameters = array())
```
stof: why a reference ? This does not make any sense here. We are not in a PHP 4 world anymore. There is no need to pass objects by reference.

stof: this helper method should be private

tests/FlexTest.php
hjanuschka: i prefer annotations over dataprovider, i think its lesser, code and easier to maintain/read
you require me to change it? - is it a ko-criteria?

tests/FlexTest.php
hjanuschka: fixed. (sorry was due to copy & paste)

hjanuschka: fixed

hjanuschka:
@stof feedback addressed
@Pierstoval comments added

tests/FlexTest.php
stof: I don't think it is less code (because your annotation contains code here). And I find maintenance easier when the code is written in actual code rather than instead a comment, as tooling will be able to recognize it and help me maintain it (I will get code highlighting, autocompletion, etc...)

and regarding readability, I find data providers quite readable when using a generator to provide the different datasets (the syntax using a nested array is indeed harder to read)

src/Flex.php
```diff
@@ -182,6 +182,27 @@ public function activate(Composer $composer, IOInterface $io)
...
+        return (object)[
+            "composer" => $lockFilePath,
+            "symfony" => $symfonyLock
```
stof: instead of returning an object with 2 properties where each caller only need 1, I would create 2 methods returning strings instead.

src/Flex.php
```diff
@@ -182,6 +182,27 @@ public function activate(Composer $composer, IOInterface $io)
...
+        ];
+    }
+
+    private function isComposerLockMissing()
+    {
+        return file_exists(Factor::getComposerFile()) && !file_exists($this->getLockFilePath()->composer);
```
stof: `Factor` is the wrong class name so this is broken

hjanuschka: 🙄 OMG - thx

tests/FlexTest.php
hjanuschka: ok got you, your repo, your rules, and all valid points, changed to a dataprovider.

src/Flex.php
hjanuschka: valid point, thx! done!

tests/FlexTest.php
```diff
@@ -78,6 +97,36 @@ public function getRecordTests()
...
+     * dataset0: custom COMPOSER env is set, generates correct symfony lock
+     * dataset1: nothing overriden, just verify default names
+     * dataset2: composer = default, custom symfony.lock path
+     */
+    public function getLockFileTestCases()
+    {
+        return  [["composer.beta.json","", "composer.beta.lock", "symfony.beta.lock"],
+                 ["","", "./composer.lock", "./symfony.lock"],
+                 ["","symfony.beta.lock", "./composer.lock", "symfony.beta.lock"]];
+    }
```
Pierstoval: You can use a generator here and simply refactor this as:

```php
yield ["composer.beta.json","", "composer.beta.lock", "symfony.beta.lock"];
yield ["","", "./composer.lock", "./symfony.lock"];
yield ["","symfony.beta.lock", "./composer.lock", "symfony.beta.lock"];
```

Better readability, less memory usage :)

stof: I'm not sure memory usage will be that different. AFAIK, PHPUnit turns the generator into an array anyway.

But this indeed improves readability.

hjanuschka: fixed!

hjanuschka:
@Pierstoval @stof  feedback addressed, didn't know that phpunit is capable of generators.

Pierstoval:
> @Pierstoval @stof feedback addressed, didn't know that phpunit is capable of generators.

Just for the anecdote, it's not phpunit, it's anything that is capable of iterating over something that is _de facto_ capable of iterating over a Generator :wink:

hjanuschka:
@stof  anything from your side that needs to be adressed?

fabpot:
@hjanuschka Thanks for your contribution. I have submitted #334, which is a simplified version of your code.

src/Flex.php
fabpot: I haven't keep this code as I don't see why one would not end its `composer.json` file with `.json`. Using `composer.foobar` works (I've tested it), but it does not seem something worth supporting.

hjanuschka:
@fabpot thx, while i am happy to have the functionallity, i am a bit sad too, i put in a alot effort in many feedback rounds into this. anyway as maintainer of oss - saying no, is the toughest part - and an important part.


however i miss, tests, and don't really get the "simplified" version by seeing the code

```
  if ($populateRepoCacheDir && isset(self::$repoReadingCommands[$command]) && ('install' !== $command || (file_exists($composerFile) && !file_exists(substr($composerFile, 0, -4).'lock')))) {
```

anyway, thx for taking your time and implementing it.

PR #330, opened 2018-03-07
Update ScriptExecutor.php

andreybolonin:


slootjes:
"KO" isn't a typo, it's just the opposite of "OK" so I don't see a reason to change this.

stof:
And displaying OK in case of error would be wrong, as it gives the wrong info. So this would introduce a bug.

so this is a -1 from me as well.

andreybolonin:
@slootjes @stof ok ))

PR #332, opened 2018-03-10
Allow to define the amount of bytes for generated values

slootjes:
Currently it's only possible with _%generate(secret)%_ to generate a 16 byte random value but I also wanted to be able to generated a 32 byte one instead. So I've implemented to possibility to define the number of bytes: _%generate(32)%_. Of course it's possible to define any number there so other use cases can be supported to. To stay backwards compatible _%generate(secret)%_ will still work.

While adding the test I also found a tiny bug where unlink() in another test.

### DIFF
--- a/src/Configurator/EnvConfigurator.php
+++ b/src/Configurator/EnvConfigurator.php
@@ -41,9 +41,7 @@ class EnvConfigurator extends AbstractConfigurator
 
         $data = '';
         foreach ($vars as $key => $value) {
-            if ('%generate(secret)%' === $value) {
-                $value = bin2hex(random_bytes(16));
-            }
+            $value = $this->replaceGeneratedValue($value);
             if ('#' === $key[0] && is_numeric(substr($key, 1))) {
                 $data .= '# '.$value."\n";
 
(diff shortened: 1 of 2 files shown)


src/Configurator/EnvConfigurator.php
```diff
@@ -142,4 +138,21 @@ private function unconfigurePhpUnit(Recipe $recipe, $vars)
             file_put_contents($phpunit, $contents);
         }
     }
+
+    private function replaceGeneratedValue($value)
+    {
+        if ('%generate(secret)%' === $value) {
+            return $this->generateRandomBytes();
+        }
+        if (preg_match('~^%generate\(([0-9]+)\)%$~', $value, $matches)) {
```
fabpot: I would say we need `generate(secret, %d+)` here instead (as generate() might generate some else than secret at some point).

src/Configurator/EnvConfigurator.php
```diff
@@ -142,4 +138,21 @@ private function unconfigurePhpUnit(Recipe $recipe, $vars)
             file_put_contents($phpunit, $contents);
         }
     }
+
+    private function replaceGeneratedValue($value)
+    {
+        if ('%generate(secret)%' === $value) {
+            return $this->generateRandomBytes();
+        }
+        if (preg_match('~^%generate\(secret,\s?([0-9]+)\)%$~', $value, $matches)) {
```
fabpot: I would use `\s*` instead of `\s?`

PR #337, opened 2018-03-20
Expand targer dirs in .gitignore

fabpot:
Closes symfony/recipes#386

### DIFF
--- a/src/Configurator/GitignoreConfigurator.php
+++ b/src/Configurator/GitignoreConfigurator.php
@@ -29,6 +29,7 @@ class GitignoreConfigurator extends AbstractConfigurator
 
         $data = '';
         foreach ($vars as $value) {
+            $value = $this->options->expandTargetDir($value);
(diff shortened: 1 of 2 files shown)


ro0NL:
@fabpot @chalasr now the bundle needs an update to include the leading slash right? :sweat_smile: 

with flex v1.0.75 it gives

```
###> lexik/jwt-authentication-bundle ###
config/jwt/*.pem
###< lexik/jwt-authentication-bundle ###
```

IMHO this path should be explicitly ignored relative to the root,  not relative to any dir.

Which is what frameworkbundle & co also does:

https://github.com/symfony/recipes/blob/4ccc10853933983a88d492edf8ef72ebb9cb3f67/symfony/framework-bundle/3.3/manifest.json#L22-L24

fabpot:
@ro0NL Already done: https://github.com/symfony/recipes/pull/387

PR #338, opened 2018-03-20
Add a new command to install missing recipes

### LINKED ISSUE #297: create a command to run missing recipes
Add a command like `composer flex-install`, which checks if the recipes have run for all installed packages. (It checks which packages are in composer.lock but not in symfony.lock)
Alternative: `composer install` runs the missing recipes.

Use cases:
* When the connection is broken while running composer, the installation is done by composer (maybe from cache), but the recipes are not run.
* When [upgrading existing application to flex](http://symfony.com/doc/current/setup/flex.html#upgrading-existing-applications-to-flex).

Work-around:
Delete the dependencies in vendor/, then run `composer install`.

fabpot:
closes #297

### DIFF
--- a/src/Command/InstallRecipesCommand.php
+++ b/src/Command/InstallRecipesCommand.php
@@ -1,0 +1,86 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Flex\Command;
+
+use Composer\Command\BaseCommand;
+use Composer\DependencyResolver\Operation\InstallOperation;
+use Composer\Factory;
+use Composer\Script\Event;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Flex\Lock;
+
+class InstallRecipesCommand extends BaseCommand
+{
+    private $flex;
+
+    public function __construct(/* cannot be type-hinted */ $flex)
+    {
+        $this->flex = $flex;
+
+        parent::__construct();
+    }
+
+    protected function configure()
+    {
+        $this->setName('install-recipes')
+            ->setDescription('Install missing recipes.')
+        ;
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        $symfonyLock = new Lock(getenv('SYMFONY_LOCKFILE') ?: str_replace('composer.json', 'symfony.lock', Factory::getComposerFile()));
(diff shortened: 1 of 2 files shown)


fabpot:
@javiereguiluz @weaverryan What do you think of this command?

When running `fix-recipes` (name to be determined), we check any missing entry in `symfony.lock` and execute the recipe. That makes it easier to upgrade to Symfony 4 or to install a missing recipe (if the server was down or the composer req crashed for some reasons).

javiereguiluz:
I like this command a lot ... but I don't think the name is correct:

```php
$this->setName('fix-recipes')
    ->setDescription('Install missing recipes.')
```

"Fixing recipes" and "installing missing recipes" looks like a different thing to me. Maybe this could be named `update-recipes` (similar to `composer update`). But maybe "update" is wrong too ... then we could name it `install-recipes` which is consistent with the description.

fabpot:
You're right, name is not good right now. `update` feels wrong as we are not updating recipes, we are merely installing the missing ones. I think the correct one would be `install-missing-recipes`. Is it a bit too long?

javiereguiluz:
I don't dislike `install-missing-recipes` ... but `install-recipes` could be enough. I mean, when you do `composer install` it only installs the missing packages (and it even says *"Nothing to install or update"* if needed).

If we need some mode to force the installation of recipes even if they are installed, a `--force` option could be added.

src/Command/RecipeFixerCommand.php
```diff
@@ -0,0 +1,86 @@
...
+use Composer\DependencyResolver\Operation\InstallOperation;
+use Composer\Factory;
+use Composer\Script\Event;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Flex\Lock;
+
+class RecipeFixerCommand extends BaseCommand
+{
+    private $flex;
```
Pierstoval: As it can't be type-hinted, can you add phpdoc on this property at least?

Pierstoval:
I'm 👍 for a `--force`. A simple "changelog" is enough, because this fix might overwrite/create files, so IMO it's better to debug them first with a dry-run

xabbuh:
An optional argument to specify the package(s) for which you want recipes to be applied as well as the `--force` option would be a nice bonus on top. But they could be added in a future PR.

weaverryan:
Good conversation!

* I like `install-recipes`
* As Javier mentioned, if everything is installed, we can tell them. If we install 2 missing recipes, we can say "Installed 2 missing recipes" so it's clear this installs only missing recipes.
* Let's discuss `--force` in another PR, but it feels cool :)

fabpot:
Name changed, ready.

nicolas-grekas:
Other name proposed in #342

PR #341, opened 2018-03-23
Fixed wrong return type in the CopyFromPackageConfigurator class

gquemener:
Currently, it is not possible to use the "copy-from-package" configurator to copy a directory.

A typo in the return type prevents it (as shown in [the first commit build](https://travis-ci.org/symfony/flex/builds/357369646)).

This PR introduces a `CopyDirectoryFromPackageConfiguratorTest` as I was concerned about increasing the overall complexity of the `CopyFromPackageConfiguratorTest` class.


fabpot:
Thank you @gquemener.

PR #342, opened 2018-03-26
Rename install-recipes to setup-recipes

nicolas-grekas:
I'm used to running `composer ins`, but #338 breaks it.


javiereguiluz:
Native English speakers could help us here, but to me `setup` and `install` have different meanings 😕

fabpot:
`fix-recipes`?

Pierstoval:
Why not `missing-recipes` with a `--force` option?

fabpot:
#343

PR #343, opened 2018-03-27
Rename intall-recipes to fix-recipes

fabpot:
replaces #342


stof:
you haven't changed the command name, so this does not rename the command

fabpot:
Indeed :)

PR #355, opened 2018-04-27
Fix: Remove unused property

localheinz:
This PR

* [x] removes an unused property

### DIFF
--- a/src/Downloader.php
+++ b/src/Downloader.php
@@ -36,7 +36,6 @@ class Downloader
     private $endpoint;
     private $caFile;
     private $flexId;
-    private $eventDispatcher;
 
     public function __construct(Composer $composer, IoInterface $io, ParallelDownloader $rfs)
     {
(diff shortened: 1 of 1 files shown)


src/Downloader.php
```diff
@@ -51,7 +50,6 @@ public function __construct(Composer $composer, IoInterface $io, ParallelDownloa
         $this->endpoint = rtrim($endpoint, '/');
         $this->io = $io;
         $config = $composer->getConfig();
-        $this->eventDispatcher = $composer->getEventDispatcher();
```
localheinz: Property is assigned a value, but never used.

fabpot:
Next time, I would prefer one PR for "simple" changes, that's better for :)

fabpot:
Good catch, thanks @localheinz.

localheinz:
Thank you, @derrabus, @fabpot, and @xabbuh!

PR #356, opened 2018-04-27
Fix: Run php-cs-fixer

localheinz:
This PR

* [x] runs `php-cs-fixer`

💁‍♂️ Should https://fabbot.io be enabled for this repository?

### DIFF
--- a/tests/Configurator/CopyDirectoryFromPackageConfiguratorTest.php
+++ b/tests/Configurator/CopyDirectoryFromPackageConfiguratorTest.php
@@ -17,7 +17,6 @@ use Composer\Composer;
 use Composer\Installer\InstallationManager;
 use Composer\IO\IOInterface;
 use Composer\Package\PackageInterface;
-use LogicException;
 use PHPUnit\Framework\TestCase;
 use Symfony\Flex\Configurator\CopyFromPackageConfigurator;
 use Symfony\Flex\Options;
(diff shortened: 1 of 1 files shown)


tests/Configurator/CopyDirectoryFromPackageConfiguratorTest.php
```diff
@@ -115,17 +114,18 @@ private function cleanUpTargetFiles()
...
+    private function rrmdir($dir)
+    {
         if (is_dir($dir)) {
             $objects = scandir($dir);
             foreach ($objects as $object) {
-                if ($object != "." && $object != "..") {
-                    if (filetype($dir."/".$object) == "dir") {
-                        $this->rrmdir($dir."/".$object);
+                if ('.' != $object && '..' != $object) {
+                    if ('dir' == filetype($dir.'/'.$object)) {
```
Pierstoval: Isn't it possible to use `===` in these statements instead?

localheinz: Sure, but out of scope of this PR. This PR is only about applying changes from running `php-cs-fixer`.

fabpot:
Thank you @localheinz.

fabpot:
I've also enabled fabbot.io for the repository

localheinz:
Thank you, @fabpot and @Pierstoval!

PR #357, opened 2018-04-27
Fix: Case mismatch

localheinz:
This PR

* [x] fixes case mismatches when invoking [`Composer\Command\BaseCommand::getIO()`](https://github.com/composer/composer/blob/1.0.2/src/Composer/Command/BaseCommand.php#L82-L98)


fabpot:
Thank you @localheinz.

localheinz:
Thank you, @derrabus, @fabpot, and @xabbuh!

PR #358, opened 2018-04-27
Fix: Do not invoke static method dynamically

localheinz:
This PR

* [x] invokes a `static` method `statically`


Pierstoval:
If you do this kind of change here, maybe every `$this->assert(...)` should also be changed to `static::assert()` in Tests, so I'm not sure it's really useful to make this kind of change. Static functions can always be called with `$this`, it's not really an issue

fabpot:
Both calls are valid. Let's close this one.

PR #361, opened 2018-04-29
Fix: Use strict comparison

localheinz:
This PR

* [x] uses strict comparisons where possible

Follows https://github.com/symfony/flex/pull/356/files#r184845342.


fabpot:
Thank you @localheinz.

localheinz:
Thank you, @fabpot and @xabbuh!

PR #363, opened 2018-05-02
Switched project id to be opt-in

fabpot:
Right now, we generate a unique id for each new project (or if a project does not have one already). But as we are only using it server side for restricting private recipes, let's move to an opt-in way. I didn't do that at first as it means having one more command in the composer command + one thing to do.

### DIFF
--- a/src/Command/FixRecipesCommand.php
+++ b/src/Command/FixRecipesCommand.php
@@ -34,7 +34,7 @@ class FixRecipesCommand extends BaseCommand
     {
         $this->setName('symfony:fix-recipes')
             ->setAliases(['fix-recipes'])
-            ->setDescription('Install missing recipes.')
+            ->setDescription('Installs missing recipes.')
         ;
     }
 
--- a/src/Command/GenerateIdCommand.php
+++ b/src/Command/GenerateIdCommand.php
@@ -29,7 +29,7 @@ class GenerateIdCommand extends Command
     protected function configure()
     {
         $this->setName('symfony:generate-id')
-            ->setDescription('Generate a unique ID for this project.')
+            ->setDescription('Generates a unique ID for this project.')
         ;
     }
 
--- a/src/Command/UnpackCommand.php
+++ b/src/Command/UnpackCommand.php
@@ -39,7 +39,7 @@ class UnpackCommand extends BaseCommand
     {
         $this->setName('symfony:unpack')
             ->setAliases(['unpack'])
-            ->setDescription('Unpack a Symfony pack.')
+            ->setDescription('Unpacks a Symfony pack.')
             ->setDefinition([
                 new InputArgument('packages', InputArgument::IS_ARRAY | InputArgument::REQUIRED, 'Installed packages to unpack.'),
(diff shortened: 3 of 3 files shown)


src/Command/UnpackCommand.php
```diff
@@ -37,7 +37,8 @@ public function __construct(PackageResolver $resolver)
 
     protected function configure()
     {
-        $this->setName('unpack')
+        $this->setName('symfony:unpack')
+            ->setAliases(['unpack'])
             ->setDescription('Unpack a Symfony pack.')
```
stof: should be `Unpacks` for consistency with the Composer descriptions.

src/Command/GenerateIdCommand.php
```diff
@@ -0,0 +1,40 @@
...
+    {
+        $this->flex = $flex;
+
+        parent::__construct();
+    }
+
+    protected function configure()
+    {
+        $this->setName('symfony:generate-id')
+            ->setDescription('Generate a unique ID for this project.')
```
stof: `Generates`

src/Command/GenerateIdCommand.php
```diff
@@ -0,0 +1,40 @@
...
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Flex\Command;
+
+use Composer\Command\RequireCommand as BaseRequireCommand;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+
+class GenerateIdCommand extends BaseRequireCommand
```
stof: Why extending the RequireCommand ?

src/Command/FixRecipesCommand.php
```diff
@@ -32,7 +32,8 @@ public function __construct(/* cannot be type-hinted */ $flex)
 
     protected function configure()
     {
-        $this->setName('fix-recipes')
+        $this->setName('symfony:fix-recipes')
+            ->setAliases(['fix-recipes'])
             ->setDescription('Install missing recipes.')
```
stof: `Installs`

src/Command/FixRecipesCommand.php
```diff
@@ -32,7 +32,8 @@ public function __construct(/* cannot be type-hinted */ $flex)
 
     protected function configure()
     {
-        $this->setName('fix-recipes')
+        $this->setName('symfony:fix-recipes')
```
ElectricMaxxx: What about moving both into a `symfony:flex:` namespase?

src/Command/GenerateIdCommand.php
```diff
@@ -0,0 +1,40 @@
...
+
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+
+class GenerateIdCommand extends Command
+{
+    private $flex;
+
+    public function __construct(/* cannot be type-hinted */ $flex)
```
Wirone: Excuse me for asking (as it's already merged), but why it cannot be type-hinted? I'm curious.

PR #365, opened 2018-05-02
Fix: Provide missing description

localheinz:
This PR

* [x] runs `composer validate` in `before_install` section on Travis
* [x] provides an otherwise missing description in `composer.json`

💁‍♂️ Running

``` 
$ composer validate
```

yields

```
/composer.json is valid for simple usage with composer but has
strict errors that make it unable to be published as a package:
See https://getcomposer.org/doc/04-schema.md for details on the schema
description : The property description is required
```


fabpot:
Thank you @localheinz.

PR #367, opened 2018-05-02
Fix: Use class keyword instead of string literals

localheinz:
This PR

* [x] consistently uses the `class` keyword instead of string literals when referencing class names


fabpot:
Thank you @localheinz.

localheinz:
Thank you, @fabpot!

PR #369, opened 2018-05-07
use the composer command event to hide suggest messages

prisis:
If you set the composer min version to v1.2 you can use the command event to hide the suggest messages


Pierstoval:
What is the real reason to hide the suggest messages?

nicolas-grekas:
@Pierstoval the message is already hidden, nothing new on the topic here, that's part of the "Flex experience", to give space to recipes.

@prisis Flex works also with older composer versions, will break them?

prisis:
Yes, this isn't a breaking change.

This will only work if the user has a composer version bigger then v1.2

maxhelias:
So with your PR, is the condition here is always necessary ?
https://github.com/symfony/flex/blob/master/src/Flex.php#L170

prisis:
Thx, missed that one, I will change that before the pr will be merged

fabpot:
Thanks for the PR, but the current was written to explicitly support old versions of Composer.

PR #377, opened 2018-07-05
Selectively enable "remove-vcs" instead of disabling interactivity

### LINKED ISSUE #285: Symfony Flex breaks interactivity of Composer scripts
Hi,

I have a composer script defined:

```json
{
    "scripts": {
        "post-install-cmd": [
            "My\\Cool\\ScriptHandler::run"
        ],
        "post-update-cmd": [
            "My\\Cool\\ScriptHandler::run"
        ]
    }
}
```

which works, except for one difference: the given InputStream is non-interactive.

```php
<?php

use Composer\Script\Event;

class ScriptHandler
{
    public function handle (Event $event)
    {
        var_dump($event->getIO()->isInteractive());
    }
}
```

The problem is flex, as this short snipped shows:

```
$ composer run-script parameters
> My\Cool\ScriptHandler::handle
bool(false)


$ composer remove symfony/flex
Package operations: 0 installs, 0 updates, 1 removal
  - Removing symfony/flex (v1.0.60)
[...]


$ composer run-script parameters
> My\Cool\ScriptHandler::handle
bool(true)
```

This currently makes implementing something similar to `incenteev/composer-parameter-handler` in combination with running flex impossible.

nicolas-grekas:
Fixes #285, requires Composer v1.7.

### DIFF
--- a/src/Flex.php
+++ b/src/Flex.php
@@ -30,6 +30,7 @@ use Composer\IO\IOInterface;
 use Composer\IO\NullIO;
 use Composer\Json\JsonFile;
 use Composer\Json\JsonManipulator;
+use Composer\Package\Comparer\Comparer;
 use Composer\Package\Locker;
 use Composer\Package\PackageInterface;
 use Composer\Plugin\PluginEvents;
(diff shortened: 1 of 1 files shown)


src/Flex.php
```diff
@@ -156,7 +157,11 @@ public function activate(Composer $composer, IOInterface $io)
             }
 
             if ('create-project' === $command) {
-                $input->setInteractive(false);
+                if (class_exists(Comparer::class)) {
```
maxhelias: The condition on the presence of the class is to determine if the option is present ?
If so, is not it be better to use the Composer constant like this `$this->composer::VERSION` ?

nicolas-grekas: Because the constant is not populated for snapshot versions of Composer...

nicolas-grekas: (comment added)

apfelbox:
Thank you for taking this over @nicolas-grekas , patch looks good 👍

fabpot:
Thank you @nicolas-grekas.

PR #378, opened 2018-07-10
Removing legacy tags significantly reduces memory and CPU usage

nicolas-grekas:
Before:
> Memory usage: 272.59MB (peak: 606.6MB), time: 5.79s

After:
> Memory usage: 145.56MB (peak: 158.56MB), time: 1.86s

### DIFF
--- a/src/Cache.php
+++ b/src/Cache.php
@@ -1,0 +1,54 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Flex;
+
+use Composer\Cache as BaseCache;
+
+/**
+ * @author Nicolas Grekas <p@tchwork.com>
+ */
+class Cache extends BaseCache
+{
+    private static $lowestTags = [
+        'symfony/symfony' => 'v3.4.0',
+    ];
+
+    public function read($file)
+    {
+        $content = parent::read($file);
+
+        if (0 === strpos($file, 'provider-symfony$')) {
+            $content = json_encode($this->removeLegacyTags(json_decode($content, true)));
+        }
+
+        return $content;
+    }
+
+    public function removeLegacyTags(array $data): array
+    {
+        foreach (self::$lowestTags as $package => $lowestVersion) {
+            if (!isset($data['packages'][$package][$lowestVersion])) {
+                continue;
+            }
+            foreach ($data['packages'] as $package => $versions) {
+                foreach ($versions as $version => $composerJson) {
+                    if (version_compare($version, $lowestVersion, '<')) {
+                        unset($data['packages'][$package][$version]);
+                    }
+                }
+            }
+            break;
+        }
+
+        return $data;
+    }
+}
--- a/src/ComposerRepository.php
+++ b/src/ComposerRepository.php
@@ -11,7 +11,11 @@
 
 namespace Symfony\Flex;
 
+use Composer\Config;
+use Composer\EventDispatcher\EventDispatcher;
+use Composer\IO\IOInterface;
 use Composer\Repository\ComposerRepository as BaseComposerRepository;
+use Composer\Util\RemoteFilesystem;
 
 /**
  * @author Nicolas Grekas <p@tchwork.com>
@@ -19,6 +23,13 @@ use Composer\Repository\ComposerRepository as BaseComposerRepository;
 class ComposerRepository extends BaseComposerRepository
 {
     private $providerFiles;
+
+    public function __construct(array $repoConfig, IOInterface $io, Config $config, EventDispatcher $eventDispatcher = null, RemoteFilesystem $rfs = null)
+    {
+        parent::__construct($repoConfig, $io, $config, $eventDispatcher, $rfs);
+
+        $this->cache = new Cache($io, $config->get('cache-repo-dir').'/'.preg_replace('{[^a-z0-9.]}i', '-', $this->url), 'a-z0-9.$');
+    }
 
     protected function loadProviderListings($data)
     {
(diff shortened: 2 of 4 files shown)


nicolas-grekas:
Only symfony/* components have a significant list of tags. We can, but case by case IMHO, when useful.

src/Cache.php
```diff
@@ -0,0 +1,46 @@
...
+
+        return $content;
+    }
+
+    public function removeLegacyTags(array $data): array
+    {
+        if (isset($data['packages']['symfony/symfony'])) {
+            foreach ($data['packages'] as $name => $versions) {
+                foreach ($versions as $version => $package) {
+                    if ($version < 3.4) {
```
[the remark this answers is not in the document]
stof: this float comparison to compare versions looks wrong to me. Versions are strings (with multiple dots in them btw).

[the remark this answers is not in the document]
nicolas-grekas: "3.4" moved to a private static (private consts aren't allowed on PHP7.0)
float comparison replaced by `version_compare()` check.

src/ComposerRepository.php
```diff
@@ -53,6 +64,12 @@ protected function fetchFile($filename, $cacheKey = null, $sha256 = null, $store
             return [];
         }
 
-        return parent::fetchFile($filename, $cacheKey, $sha256, $storeLastModifiedTime);
+        $data = parent::fetchFile($filename, $cacheKey, $sha256, $storeLastModifiedTime);
+
+        if (0 === strpos($filename, 'http://packagist.org/p/symfony/')) {
```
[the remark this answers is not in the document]
nicolas-grekas: Not doing this as this future is unlikely to happen: the jsons are already secured by a sha256 so that enabling https would just make things slower for no benefit on the security side. Running on "http" is a design decision for the packagist infrastructure.

nicolas-grekas:
Comments addressed thanks.

src/Cache.php
```diff
@@ -0,0 +1,54 @@
...
+
+use Composer\Cache as BaseCache;
+
+/**
+ * @author Nicolas Grekas <p@tchwork.com>
+ */
+class Cache extends BaseCache
+{
+    private static $lowestTags = [
+        'symfony/symfony' => 'v3.4.0',
```
stof: what about all the components ?

src/ComposerRepository.php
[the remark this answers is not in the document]
stof: it is already possible to force composer to always use HTTPS for packagist (by overriding the packagist.org registration without allowing protocol downgrade), and some people are doing it when they have issues due to a misbehaving HTTP proxy.

src/Cache.php
nicolas-grekas: symfony/symfony is found in all components, thanks to the "replace" statement I suppose.

src/ComposerRepository.php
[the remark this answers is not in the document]
nicolas-grekas: good to know, updated

src/Cache.php
```diff
@@ -0,0 +1,46 @@
...
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Flex;
+
+use Composer\Cache as BaseCache;
```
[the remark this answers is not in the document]
nicolas-grekas: then in ComposerRepository, we'll have `as ComposerComposerRepository` :)

src/Cache.php
stof: sure, but this will remove the outdated tags from the main package. I don't think it will remove outdated tags of the component packages. Meaning you only remove half of what you could be removing here

nicolas-grekas: It does for sure. Actually removing all versions of symfony/symfony (since it's not used) frees very little memory. The memory+CPU are reduced a lot *because* lowest tags for components are dropped.

src/Cache.php
```diff
@@ -0,0 +1,54 @@
...
+
+/**
+ * @author Nicolas Grekas <p@tchwork.com>
+ */
+class Cache extends BaseCache
+{
+    private static $lowestTags = [
```
Tobion: you can make this a constant. since php 5.6 array constants are supported.

src/ComposerRepository.php
```diff
@@ -53,6 +64,12 @@ protected function fetchFile($filename, $cacheKey = null, $sha256 = null, $store
             return [];
         }
 
-        return parent::fetchFile($filename, $cacheKey, $sha256, $storeLastModifiedTime);
+        $data = parent::fetchFile($filename, $cacheKey, $sha256, $storeLastModifiedTime);
+
+        if (0 === strpos($filename, 'http://packagist.org/p/symfony/') || 0 === strpos($filename, 'https://packagist.org/p/symfony/')) {
```
nalekberov: Would it make more sense if we used regular expressions like `^https?:\/\/packagist\.com\/p\/symfony`?

SpartakusMd: From PHP documentation : `Do not use preg_match() if you only want to check if one string is contained in another string. Use strpos() instead as it will be faster.`

src/Cache.php
nicolas-grekas: I don't see why a (public) constant would be better honestly...

fabpot:
Thank you @nicolas-grekas.

src/ComposerRepository.php
Seldaek: It'd make more sense to either use preg_match or parse_url and check for the domain being packagist.org OR repo.packagist.org as we are likely going to switch to that domain in the future.

src/Cache.php
Pierstoval: Yep, and constants visibility is only for PHP 7.1 so it wouldn't work for Flex which is compatible with 7.0

src/ComposerRepository.php
dunglas: Actually, do we care about the domain? We should discard the legacy tags even if a proxy or a private Packagist is used.

PR #380, opened 2018-07-11
Move away from symfony.sh

fabpot:
Several companies reported that they have problems with symfony.sh (blocked by their proxies). I got one more report today, so let's move away from this specific domain and move to `https://flex.symfony.com` as the default endpoint (I didn't want to do that at first because I didn't want to expose `flex` anywhere, but I suppose that purely internal and I can live with that).

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,4 +10,4 @@ dependencies.
 
 [1]: https://symfony.com/doc/current/setup/flex.html
 [2]: https://symfony.com
-[3]: https://symfony.sh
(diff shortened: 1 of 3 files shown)


Koc:
https://symfony.sh/ now duplicates https://flex.symfony.com/ . Is it better to make redirect?

fabpot:
@Koc That's an intermediary step. Next move will be to indeed redirect traffic from the old domain to the new one.

PR #384, opened 2018-07-11
keep all custom Composer repositories

### LINKED ISSUE #382: Repository type is not registered - v1.0.81
The latest release seems to break [custom repositories](https://getcomposer.org/doc/05-repositories.md#vcs) feature in `composer.json`.

Step to reproduce:

Add this in your `composer.json`: 
```
"repositories": [
        {
            "type": "vcs",
            "url": "https://github.com/0x46616c6b/monolog"
        }
    ],
```

Then run `composer update`.

Expected result: the update run properly

Actual result: 
```
Loading composer repositories with package information                                                                                                                         Updating dependencies (including require-dev)         
Package operations: 74 installs, 0 updates, 0 removals
  - Installing symfony/flex (v1.0.81): Loading from cache
Plugin installation failed, rolling back
  - Removing symfony/flex (v1.0.81)

                                          
  [InvalidArgumentException]              
  Repository type is not registered: vcs  
```

The update works fine if I manually set `symfony/flex` to v1.0.80.

composer v1.6.5
PHP 7.2.7

xabbuh:
This fixes #382.

### DIFF
--- a/src/Flex.php
+++ b/src/Flex.php
@@ -37,7 +37,6 @@ use Composer\Plugin\PluginEvents;
 use Composer\Plugin\PluginInterface;
 use Composer\Plugin\PreFileDownloadEvent;
 use Composer\Repository\ComposerRepository as BaseComposerRepository;
-use Composer\Repository\RepositoryFactory;
 use Composer\Repository\RepositoryManager;
 use Composer\Script\Event;
 use Composer\Script\ScriptEvents;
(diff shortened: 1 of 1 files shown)


src/Flex.php
```diff
@@ -104,11 +103,13 @@ public function activate(Composer $composer, IOInterface $io)
 
         $rfs = Factory::createRemoteFilesystem($this->io, $this->config);
         $this->rfs = new ParallelDownloader($this->io, $this->config, $rfs->getOptions(), $rfs->isTlsDisabled());
+        $repositoryClassesProperty = new \ReflectionProperty(RepositoryManager::class, 'repositoryClasses');
+        $repositoryClassesProperty->setAccessible(true);
         $manager = new RepositoryManager($this->io, $this->config, $composer->getEventDispatcher(), $this->rfs);
```
xabbuh: An alternative solution would be to use `RepositoryFactory::manager()` here to create the `RepositoryManager` instance, but that would not take into account any changes that could have been made (by plugins, for example) to the original manager.

stof:
isn't this missing adding the repos ?

xabbuh:
Composer seems to do that itself by calling `createRepository()` based on the configured repository classes. At least, it did that on a local project where I have a `path` repository.

xabbuh:
But we may indeed pull the formerly set repositories with `getRepositories()` and set them as is (except for the `composer` repository). What do you think?

PR #389, opened 2018-07-18
Fixing duplicate Cache error

### LINKED ISSUE #388: Cache namespace is already in use
Since v1.0.83, a new file Cache.php was added in src/
This file cause an error in Download.php beacuse it uses  (line 14):
use Composer\Cache;
which causes an error because "Cache" already exists in this namespace (because of that new file)
Can you fix this ?

weaverryan:
Fixes #388

### DIFF
--- a/src/Downloader.php
+++ b/src/Downloader.php
@@ -11,7 +11,7 @@
 
 namespace Symfony\Flex;
 
-use Composer\Cache;
+use Composer\Cache as ComposerCache;
 use Composer\Composer;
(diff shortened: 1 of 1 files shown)


maxhelias:
I think we should instead use the `Cache` defined in Flex because it extends from `Composer\Cache`.
Just delete line 14 should be enough in my opinion

nicolas-grekas:
@maxhelias Cache in the flex namespace has a different responsibility, we should really use Composer's here.

fabpot:
Thank you @weaverryan.

PR #391, opened 2018-07-25
Avoid removing versions of unrelated packages when loading non-provider files

Seldaek:
Not all files loaded from cache are provider-style files. Private Packagist for example serves all packages in one request for efficiency and the current flex code was then happily removing any package's version below 3.4.0 when doing so.


fabpot:
Thank you @Seldaek.

PR #395, opened 2018-07-26
Handle composer special aliases "mirrors" and "nothing"

### LINKED ISSUE #394: "mirrors" is not a valid alias
When using [packagist.com](https://packagist.com) we need to launch the command `composer update mirrors` but it fail with the error `"mirrors" is not a valid alias`.

I only get this since I'm using flex.

nicolas-grekas:
Fixes #394

The recipe validation server should also have a rule to fail when an alias on "mirrors" or "nothing" is submitted.

Composer also handles "lock" as a special alias, but we already have the lock component, let's take priority.

### DIFF
--- a/src/PackageResolver.php
+++ b/src/PackageResolver.php
@@ -48,7 +48,7 @@ class PackageResolver
         // second pass to resolve package names
         $packages = [];
         foreach ($explodedArguments as $i => $argument) {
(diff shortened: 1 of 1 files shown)


stof:
@fabpot how does the backend handle the `lock` alias ?

fabpot:
What do you mean? With the new version, nobody can "register" those special aliases, but that does not block those aliases from being used by symfony/*.

nicolas-grekas:
Oh indeed, all is fine then :)

fabpot:
Thank you @nicolas-grekas.

PR #396, opened 2018-07-26
Add compatibility with updated RemoteFilesystem::getRemoteContents() signature

nicolas-grekas:
Required after https://github.com/composer/composer/pull/7495

Code is compatible with both signatures.


fabpot:
Thank you @nicolas-grekas.

PR #399, opened 2018-07-27
Enable HTTP/2

nicolas-grekas:
See https://curl.haxx.se/docs/http2.html for details.
And  https://twitter.com/seldaek/status/1022179344091893760 for motivation :)
Over unencrypted links, HTTP/2 is disabled as there is no benefit to using it then (there would be drawbacks actually: TCP congestion slows down HTTP/2 so that opening several unencrypted HTTP/1.1 connections should be better - already handled by cURL.)


src/Flex.php
```diff
@@ -659,7 +659,7 @@ private function populateRepoCacheDir()
                 continue;
             }
 
-            if (!preg_match('#^https\?://#', $repo['url'])) {
```
stof: why was it matching strings starting with `https?://` ? Wasn't the escaping of the  `?` in the regex done by mistake ?

nicolas-grekas: not a mistake, but a special case that was not handled properly, see:
https://github.com/composer/composer/blob/master/src/Composer/Repository/ComposerRepository.php#L74

glensc: permalink for above comment:
- https://github.com/composer/composer/blob/master@%7B2018-07-27T10:39:09Z%7D/src/Composer/Repository/ComposerRepository.php#L74

PR #401, opened 2018-07-27
Fix using php://memory on Windows

### LINKED ISSUE #400: getRemoteContents broke composer 1.7.0-RC
Hi,
`getRemoteContents` not return correctly `$http_response_header` when it use in composer `1.7.0-RC`.
Only when `composer update` is called.

Symfony/flex override the method `getRemoteContents` line 582 in `src/Composer/Util/RemoteFilesystem.php` call line 318 in same file and return an empty `$http_response_header`.

It breaks the rest of the code.

cf issue : [composer#7493](https://github.com/composer/composer/issues/7493)

nicolas-grekas:
Fixes #400
Fixes https://github.com/composer/composer/issues/7493
I'm not sure why (php or ext-curl bug?), but this change fixes the issue (reproduced locally).

### DIFF
--- a/src/CurlDownloader.php
+++ b/src/CurlDownloader.php
@@ -64,7 +64,7 @@ class CurlDownloader
         $params = stream_context_get_params($context);
 
         $ch = curl_init();
-        $hd = fopen('php://memory', 'w+b');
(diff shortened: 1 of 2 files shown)


fabpot:
Thank you @nicolas-grekas.

PR #403, opened 2018-07-28
Add http2 support

staabm:
adding http2 support, as packagist.org supports it now
https://twitter.com/seldaek/status/1022179344091893760

composer will use repo.packagist.org in the future for package metadata
see https://github.com/composer/composer/commit/c5fa3bdde0e042b64e32102c21edb2e12e474746


staabm:
Already supported in a different path

PR #409, opened 2018-08-13
Add SYMFONY_REQUIRE env var + disable recipes when flex is not in the root composer.json

nicolas-grekas:
The C.I. for `symfony/symfony` is starting to fail transiently because solving the dependency graph takes too much memory.
See e.g. https://ci.appveyor.com/project/fabpot/symfony/build/1.0.39377

But thanks to recent improvements in Flex, we know why and how to fix that: skipping old tags.
This PR allows using Flex while disabling the recipes subsystem and allows configuring the minimum version of components we give to the solver.

I plan to use the next version of Flex that embeds this PR (could be v1.1.0 btw) on `symfony/symfony`.

### DIFF
--- a/src/Cache.php
+++ b/src/Cache.php
@@ -12,70 +12,27 @@
 namespace Symfony\Flex;
 
 use Composer\Cache as BaseCache;
+use Composer\IO\IOInterface;
+use Composer\Semver\Constraint\Constraint;
+use Composer\Semver\VersionParser;
 
 /**
  * @author Nicolas Grekas <p@tchwork.com>
  */
 class Cache extends BaseCache
 {
-    private static $lowestTags = [
-        'symfony/symfony' => [
-            'version' => 'v3.4.0',
-            'replaces' => [
-                'symfony/asset',
-                'symfony/browser-kit',
-                'symfony/cache',
-                'symfony/config',
-                'symfony/console',
-                'symfony/css-selector',
-                'symfony/dependency-injection',
-                'symfony/debug',
-                'symfony/debug-bundle',
-                'symfony/doctrine-bridge',
-                'symfony/dom-crawler',
-                'symfony/dotenv',
-                'symfony/event-dispatcher',
-                'symfony/expression-language',
-                'symfony/filesystem',
-                'symfony/finder',
-                'symfony/form',
-                'symfony/framework-bundle',
-                'symfony/http-foundation',
-                'symfony/http-kernel',
-                'symfony/inflector',
-                'symfony/intl',
-                'symfony/ldap',
-                'symfony/lock',
-                'symfony/messenger',
-                'symfony/monolog-bridge',
-                'symfony/options-resolver',
-                'symfony/process',
-                'symfony/property-access',
-                'symfony/property-info',
-                'symfony/proxy-manager-bridge',
-                'symfony/routing',
-                'symfony/security',
-                'symfony/security-core',
-                'symfony/security-csrf',
-                'symfony/security-guard',
-                'symfony/security-http',
-                'symfony/security-bundle',
-                'symfony/serializer',
-                'symfony/stopwatch',
-                'symfony/templating',
-                'symfony/translation',
-                'symfony/twig-bridge',
-                'symfony/twig-bundle',
-                'symfony/validator',
-                'symfony/var-dumper',
-                'symfony/web-link',
-                'symfony/web-profiler-bundle',
-                'symfony/web-server-bundle',
-                'symfony/workflow',
-                'symfony/yaml',
-            ],
-        ],
-    ];
+    private $versionParser;
+    private $symfonyRequire;
+    private $symfonyContraints;
+    private $io;
+
+    public function setSymfonyRequire(string $symfonyRequire, IOInterface $io)
+    {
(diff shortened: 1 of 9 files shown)


nicolas-grekas:
For `symfony/symfony` master:

before:
`Memory usage: (peak: 1065.49MB), time: 9.91s`

 with `SYMFONY_LOWEST_VERSION=4.1`:
`Memory usage: (peak: 257.37MB), time: 2.57s`

nicolas-grekas:
PR now handles a new `SYMFONY_REQUIRE` env var that should contain a version constraint that will apply to all packages replaced by `symfony/symfony`.
This works without having to explicitly list these packages in the source, easing maintaince.
This will also allow deprecating `symfony/lts` and `symfony/forcee-lowest`, allowing to simplify the releasing process.

nicolas-grekas:
Note that as a side effect, this will allow any PHP project to benefit from flex' parallel downloader while disabling the recipe subsystem: just require it + add extra.symfony.endpoint=none in composer.json.

BoShurik:
@nicolas-grekas is it possible to set `extra.symfony.require` instead of `SYMFONY_REQUIRE` env var?

src/Cache.php
```diff
@@ -90,23 +42,24 @@ public function read($file)
...
-            foreach ($data['packages'] as $package => $versions) {
-                if ($package !== $lowestPackage && !in_array($package, $replacedPackages, true)) {
+        if (!$this->symfonyRequire || !isset($data['packages']['symfony/symfony'])) {
+            return $data;
+        }
+        $symfonyVersions = $data['packages']['symfony/symfony'];
+
+        foreach ($data['packages'] as $name => $versions) {
+            foreach ($versions as $version => $package) {
+                if ('symfony/symfony' !== $name && !isset($symfonyVersions[preg_replace('/^(\d++\.\d++)\..*/', '$1.x-dev', $version)]['replace'][$name])) {
```
maxhelias: Would not it be better to put two if and put the comparison `'symfony/symfony' === $name` in first with a `continue2;` ?

nicolas-grekas:
@BoShurik I forgot to mention it, but Yes! I even plan to use it in our skeletons to replace `symfony/lts`.

src/Cache.php
nicolas-grekas: the comparison is using `!==`, not `===`, so this wouldn't work.

maxhelias: I meant, explode the if in two parts and change the `!==` by `===`

nicolas-grekas: I'm sorry, I don't understand, please suggest the patch you have in mind.

maxhelias: Like : 
```
if ('symfony/symfony' === $name) {
   continue2;
}
if (!isset($symfonyVersions[preg_replace('/^(\d++\.\d++)\..*/', '$1.x-dev', $version)]['replace'][$name])) {
   // ....
}
```

nicolas-grekas: But this would *not* be doing the same.

maxhelias: Indeed, i just understood 😅

nicolas-grekas:
PR ready, works as expected in https://github.com/symfony/symfony/pull/28199

nicolas-grekas:
Now with a notice:
image

PR ready, it works well :)

src/Downloader.php
```diff
@@ -42,21 +42,22 @@ public function __construct(Composer $composer, IoInterface $io, ParallelDownloa
...
-        $this->endpoint = rtrim($endpoint, '/');
+
         $this->io = $io;
         $config = $composer->getConfig();
         $this->rfs = $rfs;
         $this->cache = new ComposerCache($io, $config->get('cache-repo-dir').'/'.preg_replace('{[^a-z0-9.]}i', '-', $this->endpoint));
         $this->sess = bin2hex(random_bytes(16));
-
-        if (self::$DEFAULT_ENDPOINT !== $endpoint) {
-            $this->io->writeError('<warning>Warning: Using '.$endpoint.' as the Symfony endpoint</warning>');
```
nicolas-grekas: warning move below, so that unrelated Composer commands don't display it.

nicolas-grekas:
I removed the special handling of the "none" value. Instead, flex disables recipes when it can't find itself in the "require" of the root composer.json

src/Downloader.php
```diff
@@ -129,6 +138,12 @@ public function getRecipes(array $operations): array
 
         $data = [];
         foreach ($bodies as $body) {
+            // add default values in case the response is empty
+            $body += [
+                'manifests' => [],
+                'vulnerabilities' => [],
+                'locks' => [],
+            ];
```
fabpot: I don"t think this is needed and should probably be removed.

nicolas-grekas: L135 populates $bodies with empty arrays, that's why.

nicolas-grekas: Pfff but the "if" L135 also prevents it...

fabpot: But this code works right now without adding this, right?

nicolas-grekas: Removed

fabpot:
Thank you @nicolas-grekas.

PR #411, opened 2018-08-21
Fix a typo ("contraints" -> "constraints")

pamil:


nicolas-grekas:
Good catch, thanks @pamil.

PR #412, opened 2018-08-21
Don't filter packages that have empty intersection with symfony/symfony when matching extra.symfony.require

### LINKED ISSUE #410: "extra.symfony.require" setting restrict packages outside the range of "symfony/symfony"
Behat requires `symfony/class-loader` in `~2.1||~3.0` version. This Symfony package has been dropped in `4.0` branch.

I'd like to use `extra.symfony.require` set to `4.1.*` in order to test my application in that Symfony version, but it's impossible right now:

```
Restricting packages listed in "symfony/symfony" to "4.1.*"
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - behat/behat v3.5.0 requires symfony/class-loader ~2.1||~3.0 -> satisfiable by symfony/symfony[v3.4.14].
    - behat/behat v3.5.0 requires symfony/class-loader ~2.1||~3.0 -> satisfiable by symfony/symfony[v3.4.14].
    - behat/behat v3.5.0 requires symfony/class-loader ~2.1||~3.0 -> satisfiable by symfony/symfony[v3.4.14].
    - Conclusion: remove symfony/symfony v3.4.14
    - Installation request for behat/behat (locked at v3.5.0, required as ^3.2) -> satisfiable by behat/behat[v3.5.0].
```

Fixing this bug would require checking if the package being restricted is defined in `symfony/symfony`'s replace list at given version: https://github.com/symfony/symfony/blob/0874f7098fb92d8773e252a0076f2947bff45047/composer.json#L34-L86

nicolas-grekas:
Fixes #410
ping @pamil, please confirm.

### DIFF
--- a/src/Cache.php
+++ b/src/Cache.php
@@ -26,7 +26,7 @@ class Cache extends BaseCache
     private $symfonyConstraints;
     private $io;
 
-    public function setSymfonyRequire(string $symfonyRequire, IOInterface $io)
+    public function setSymfonyRequire(string $symfonyRequire, IOInterface $io = null)
     {
         $this->versionParser = new VersionParser();
         $this->symfonyRequire = $symfonyRequire;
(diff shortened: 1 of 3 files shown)


pamil:
It solved the issue, thanks! 🎉

src/Cache.php
```diff
@@ -50,24 +50,49 @@ public function removeLegacyTags(array $data): array
...
+
+        foreach ($symfonySymfony as $version => $composerJson) {
+            if ('dev-master' === $version) {
+                $normalizedVersion = $this->versionParser->normalize($composerJson['extra']['branch-alias']['dev-master']);
+            } else {
+                $normalizedVersion = $composerJson['version_normalized'];
+            }
+
+            if ($this->symfonyConstraints->matches(new Constraint('==', $normalizedVersion))) {
+                $symfonyPackages += $composerJson['replace'];
```
stof: should use `?? []` to ensure we don't trigger a notice if the json does not contain this key (which is optional)

nicolas-grekas: `symfony/symfony` always has this "replace" key (same for the other ones that are unchecked btw.) I'd prefer not crippling the code for these checks...

fabpot:
Thank you @nicolas-grekas.

teohhanhui:
Wondering if a new version could be tagged? This bug sounds like it'll be a blocker for many Symfony 4 projects.

Issue #414, opened 2018-08-23
Symfony recipes are disabled when flex set as dev dependency


jlekowski:
Since `symfony/flex 1.1` (https://github.com/symfony/flex/pull/409) when I run `composer install` I get `Symfony recipes are disabled: "symfony/flex" is not in the "require" section of the root composer.json`.
I do want to both have recipes, and `symfony/flex` in `require-dev` (as decided here https://github.com/symfony/skeleton/issues/24).

So I would like to run e.g. `composer req weblink` (on `dev` env only of course), but it's not working. It would work if `symfony/flex` was `prod` dependency (in `require` section).

xabbuh:
The move of Flex to `require-dev` in the skeleton was reverted in #26.

jlekowski:
Thanks @xabbuh, I will move flex dependecy into `require` then.

garak:
I'm not sure to understand the reason for moving back flex to `require` section, nor the reason to disabling flex when in `require-dev`

nicolas-grekas:
The reason is that flex does important work on composer install also, related to executing scripts. That's why it must be in "require", so that those scripts run also when you deploy your prod. Then, disabling flex in require-dev makes it more evident that it should be moved to require.

garak:
Right, but what if I don't need to run auto-scripts? That's a practical case, I'm running other deploy commands and I can perfectly deploy a Symfony application without flex. But I don't want to give up the opportunity to run flex during development, so why you want to force me to use flex in production?

nicolas-grekas:
I'm just giving you the rationale so you can challenge it.

garak:
And I absolutely appreciate that. But I'm wondering if there can be room for an improvement, giving the option to developers to allow flex in `require-dev` instead of forcing them (of course, defaulting to `require`, as explained). If so, I can open an issue.

nicolas-grekas:
See #415

jlekowski:
Feels like I opened Pandora's box :)

I agree with @garak regarding this - keeping flex in `require-dev` would be nice option to have.

Thanks @nicolas-grekas for taking your time on the issue and this PR https://github.com/symfony/flex/pull/415

kosztadani:
This change still breaks our workflow.

[...]

PR #415, opened 2018-08-24
Reenable flex when it is found in require-dev

nicolas-grekas:
As discussed in #414


fabpot:
Thank you @nicolas-grekas.

PR #423, opened 2018-09-21
Change the EnvConfigurator to only handle the .env file

dunglas:
Companion PR for symfony/recipes#466 and symfony/symfony#28533.

As explained in https://github.com/symfony/recipes/pull/466#discussion_r219453664, it will require to tag a v2 to prevent problems with previous installation.

In a next PR, we may want to allow recipes to add values in specific environment env files (such as `.env.test`. It will require to tweak a bit the Flex server. I'm not sure that it's really needed.

### DIFF
--- a/src/Configurator/EnvConfigurator.php
+++ b/src/Configurator/EnvConfigurator.php
@@ -22,20 +22,18 @@ class EnvConfigurator extends AbstractConfigurator
     {
         $this->write('Added environment variable defaults');
 
-        $this->configureEnvDist($recipe, $vars);
-        $this->configurePhpUnit($recipe, $vars);
+        $this->configureEnv($recipe, $vars);
     }
 
     public function unconfigure(Recipe $recipe, $vars)
     {
-        $this->unconfigureEnvFiles($recipe, $vars);
-        $this->unconfigurePhpUnit($recipe, $vars);
+        $this->unconfigureEnvFiles($recipe);
     }
 
-    private function configureEnvDist(Recipe $recipe, $vars)
+    private function configureEnv(Recipe $recipe, $vars)
     {
-        $distenv = getcwd().'/.env.dist';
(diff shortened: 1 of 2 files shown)


maxhelias:
In this case, would not it be more prudent to merge it on a branch `dev` to prepare a new version ?
Because for this PR, this implies a change of Git workflow for the devs because the advantage of the `.env.dist` file is that it is versioned.


Or I suggest defining a system like symfony/symfony#28533 to edit all environments and have a recipe structure like this for add values in specific environment :
```
"env" : {
   "dev": {
      // ...
   },
   "test": {
     // ...
   }
}
```
We would just have to make conditions on the keys in the env configurator without creating a break change.

WDYT?

dunglas:
> Because for this PR, this implies a change of Git workflow for the devs because the advantage of the .env.dist file is that it is versioned.

Now it's the `.env` file that is versioned. The developper can override it by using a `.env.local` file that isn't versioned. See https://github.com/symfony/recipes/pull/466/files#diff-48a706859e5c9d4e903b5065baa57726R21

> Or I suggest defining a system like symfony/symfony#28533 to edit all environments and have a recipe structure like this for add values in specific environment :

This PR already requires symfony/symfony#28533 to be merged! The config you propose would be fine, but in another PR (probably after 4.2). I think (as it requires to modify the Flex server, on the hand the current PR works with the current server).

> without creating a break change.

Actually, we *need* the breaking because of this: https://github.com/symfony/recipes/pull/466#discussion_r219450121

nicolas-grekas:
I'm not sure we need this PR anymore.
Instead, I'd suggest doing just one thing: when a `.env.test` file is found, skip updating phpunit.xml files.
And keep all the other behaviors identical.

dunglas:
Closing in favor of #434

PR #434, opened 2018-11-09
Skip updating phpunit.xml files when .env.test exists

nicolas-grekas:
This is to support `.env` < `.env.local` < `.env.test` < `.env.test.local` dotenv files *withouth affecting existing workflows* base on `.env(.dist)`, `phpunit.xml(.dist)`.

Leveraged by https://github.com/symfony/recipes/pull/481


nicolas-grekas:
Replaces https://github.com/symfony/flex/pull/423

nicolas-grekas:
Just tested with all recent changes all together: it's deceptively simple. It feels like "it just works" :)

Issue #433, opened 2018-11-08
composer require keep unbounded constraint


henri9813:
When i run 
`composer require symfony/console`

My composer.json got a new line which correspond to this package, but the version is unbounded, 
```
 "symfony/console": "*",
```
How can i get fixed version ? ( example: ^4.1 etc.. ) 

Regards

Pierstoval:
You do not need it when using Flex, because the version constraint for Symfony packages is in the `extra.symfony.require` composer option.

Soullivaneuh:
@Pierstoval We especially don't want this option, that's why we removed it.

But composer still require an unbound constraint.

nicolas-grekas:
Hello, wrong repo: constraints are set on this one: https://github.com/symfony/skeleton
But the constraints *are* bounded, thanks to the extra.symfony.require line in
https://github.com/symfony/skeleton/blob/master/composer.json#L59

Soullivaneuh:
Hello @nicolas-grekas,

I'm not sure it's related to the skeleton. Constraint are indeed bounded thanks to the `extra.symfony.require` option.

The issue we have is we removed this option and `composer require symfony/package` still add un unbound constraint instead of searching for the last stable one.

As the option resolving is related to flex, isn't that the good place for this issue?

stof:
Well, if you remove the Flex-specific constraint, you are left with normal composer constraints.

What Flex does is filter out the available versions based on `extra.symfony.require` when reading Packagist metadata. It does not do special things in the solver.
Btw, this Flex feature does not require using a `*` constraint for your requirement (you only need to make sure that your `extra.symfony.require` constraint gives you some versions matching your requirement)

nicolas-grekas:
> composer require symfony/console
> My composer.json got a new line which correspond to this package, but the version is unbounded,
> "symfony/console": "*",

OK, this is Flex doing it, reopening :)

nicolas-grekas:
fixed in #435

Issue #421, opened 2018-09-13
extra.symfony.require breaks composer.json validity


fbourigault:
Since the introduction of the `extra.symfony.require` setting, Symfony dependencies constraints are considered unbound by composer. So running `composer validate --strict` output some warnings and return an non-0 exit code which, in my case, breaks CI composer validation.

To reproduce the issue, just create an new Symfony flex using `composer create-project symfony/skeleton flex-issue`, then fill the `name` and `properties` fields and run `composer validate --strict`

```
./composer.json is valid, but with a few warnings
See https://getcomposer.org/doc/04-schema.md for details on the schema
require.symfony/console : unbound version constraints (*) should be avoided
require.symfony/framework-bundle : unbound version constraints (*) should be avoided
require.symfony/yaml : unbound version constraints (*) should be avoided
```

nicolas-grekas:
We could use `4.*` in the skeletons maybe. But from a maintainer perspective, this will create merge conflicts when merging 3.4 into 4.* (because we'll use `3.*` there of course).

nicolas-grekas:
fixed in #435

PR #435, opened 2018-11-09
Bound constraints by copying extra.symfony.require

nicolas-grekas:
Fixes https://github.com/symfony/flex/issues/433
Fixes https://github.com/symfony/flex/issues/421

When a new project is created, this makes flex replace `"*"` constraints by the one specified in `extra.symfony.require`.
The same logic also applies to `composer require symfony/*`

### DIFF
--- a/src/Flex.php
+++ b/src/Flex.php
@@ -238,8 +238,23 @@ class Flex implements PluginInterface, EventSubscriberInterface
     {
         $json = new JsonFile(Factory::getComposerFile());
         $manipulator = new JsonManipulator(file_get_contents($json->getPath()));
+
         // new projects are most of the time proprietary
         $manipulator->addMainKey('license', 'proprietary');
+
+        // replace unbounded contraints for symfony/* packages by extra.symfony.require
+        $config = json_decode($contents);
+        if ($symfonyVersion = $config['extra']['symfony']['require'] ?? null) {
+            $versions = $this->downloader->get('/versions.json');
+            foreach (['require', 'require-dev'] as $type) {
+                foreach ($config[$type] ?? [] as $package => $version) {
+                    if ('*' === $version && isset($versions['splits'][$package])) {
+                        $manipulator->addLink($type, $package, $symfonyVersion);
+                    }
+                }
+            }
+        }
+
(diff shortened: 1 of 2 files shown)


teohhanhui:
Magic! 🔮

javiereguiluz:
@teohhanhui I understand we don't always have enough time to write detailed comments ... but please don't add comments like *"Magic!"* because don't help much to the discussion.

For example, you think this is magic and therefore bad. Could you please describe some scenarios where this feature would backfire us? For example:

* If I do ........ and then ......., this feature will be confusing because the result will be ....... instead of ........
* This feature will change .......... when doing ......... and that will be confusing to users
* etc.

Thanks!

teohhanhui:
@javiereguiluz Sorry, it was just a fun comment. I didn't mean for it to be negative. :smile:

teohhanhui:
But, yes, I do have a concern. This might affect other projects that use Symfony Flex and actually intend to use unbounded constraints. (*Why?*)

nicolas-grekas:
In case one is wondering, this applies only to packages in symfony/symfony

ByScripts:
Just to be sure, will this not cause any problems with `symfony/<some>-pack`? Since the version of the packs are 1.* at the moment.

nicolas-grekas:
> cause any problems with symfony/<some>-pack

I need to check that, you might be right!

nicolas-grekas:
Fixed: the replacement applies now only to packages listed at https://flex.symfony.com/versions.json

fabpot:
Thank you @nicolas-grekas.

src/Flex.php
```diff
@@ -238,8 +238,23 @@ public function configureProject(Event $event)
     {
         $json = new JsonFile(Factory::getComposerFile());
         $manipulator = new JsonManipulator(file_get_contents($json->getPath()));
+
         // new projects are most of the time proprietary
         $manipulator->addMainKey('license', 'proprietary');
+
+        // replace unbounded contraints for symfony/* packages by extra.symfony.require
+        $config = json_decode($contents);
```
fabpot: `$contents` does not exist here.

Issue #436, opened 2018-11-09
composer outdated -D don't work


tarlepp:
Now that Symfony version is set on `extra.symfony.require` part it seems like if I have eg `4.1.6` there `composer outdated -D` command won't notice any updates about those packages.

tarlepp:
Also I did some testing with this, and noticed another (or basically the same) bug.

1) run command `composer create-project symfony/skeleton foobar`
2) that installs Symfony with [this composer.json](https://gist.github.com/tarlepp/bd688b465ea25bb5abf0aee62375db6b#file-composer_start-json) and [this composer.lock](https://gist.github.com/tarlepp/bd688b465ea25bb5abf0aee62375db6b#file-composer_start-lock)
3) then I change `composer.json` file to contain specified Symfony version like [this](https://gist.github.com/tarlepp/bd688b465ea25bb5abf0aee62375db6b#file-composer_after-json)
4) and after that change I run `composer update` command and I'm seeing this:
```
wunder@code2018-VirtualBox:~/projects/flex-bug$ composer update
Loading composer repositories with package information
Updating dependencies (including require-dev)
Restricting packages listed in "symfony/symfony" to "4.1.6"
Nothing to install or update
Writing lock file
Generating autoload files
Executing script cache:clear [OK]
Executing script assets:install public [OK]
```
5) What I expected is that actually version `4.1.6` is used on _every_ package, but my `composer.lock` file haven't really changed those versions at all, [see here](https://gist.github.com/tarlepp/bd688b465ea25bb5abf0aee62375db6b#file-composer_after-lock) there is only two differences between those lock files:
![image](https://user-images.githubusercontent.com/595561/48287318-2b147b00-e471-11e8-94a3-f7c910e629cb.png)

tarlepp:
Oh, and composer version which I used:
```
wunder@code2018-VirtualBox:~/projects/flex-bug$ composer --version
Composer version 1.7.3 2018-11-01 10:05:06
```

also tried with `1.6.5` (which seems to be the latest rollback) and the end results was the same with both of these.

nicolas-grekas:
For the second comment: Composer never downgrades so if you didn't change the constraint itself in require(-dev) then Composer won't downgrade from 4.1.7 to 4.1.6.

For the outdated command, we should bypass the restriction and basically ignore extra.symfony.require

nicolas-grekas:
Fixed in https://github.com/symfony/flex/pull/437

[...]

PR #437, opened 2018-11-10
Ignore "extra.symfony.require" when running the "outdated" command

nicolas-grekas:
Fixes https://github.com/symfony/flex/issues/436


fabpot:
Thank you @nicolas-grekas.

PR #438, opened 2018-11-14
Fix undefined index in package resolver

juchi:
When trying to install a symfony/* package without the `extra.symfony.require` composer option we run into an undefined index error.
Seems to be caused by db60ae8

### DIFF
--- a/src/PackageResolver.php
+++ b/src/PackageResolver.php
@@ -98,7 +98,7 @@ class PackageResolver
             try {
                 $config = @json_decode(file_get_contents(Factory::getComposerFile()), true);
             } finally {
-                if (!$isRequire || !($config['extra']['symfony']['require'] || isset($config['require']['symfony/framework-bundle']))) {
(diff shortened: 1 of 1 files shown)


src/PackageResolver.php
```diff
@@ -98,7 +98,7 @@ private function parseVersion(string $package, string $version, bool $isRequire)
             try {
                 $config = @json_decode(file_get_contents(Factory::getComposerFile()), true);
             } finally {
-                if (!$isRequire || !($config['extra']['symfony']['require'] || isset($config['require']['symfony/framework-bundle']))) {
+                if (!$isRequire || !(isset($config['extra']['symfony']['require']) || isset($config['require']['symfony/framework-bundle']))) {
```
fabpot: That does not mean the same thing. Here, is it is defined, we still need to check if it's true or false. Can you update the patch?

fabpot:
Thank you @juchi.

PR #439, opened 2018-11-20
Implement "sync-recipes --force" option

### LINKED ISSUE #179: [Proposal] Create a command to reset a recipe
### Current situation

As we already know it today, there is no upgrade point for Symfony applications.

With Standard Edition, we need to copy/paste code from the SE into our app when we want to "follow the SE upgrade path".

With Flex, it's gonna be the same in the case of the skeleton `composer.json` file, but worst part is for the recipes.

### Upgrade example

Let's imagine a case that can easily occur in real life:

I'm using SF3.3 today with the good ol' Standard Edition. I'm upgrading to Flex in November when it releases. Fine, a few days of work to adapt it, it's a big app, but it seems ok, and I can deploy on prod.

A few days/weeks/months/years/whatever later, I need to upgrade my app to Symfony 4.

We all know the huge amount of changes in Symfony best practices since Symfony 3.0, maybe my app was using old practices that are highly deprecated, and in order to avoid having to spend hundreds of hours debugging my code etc., I'd like at least to upgrade my packages without them to throw deprecations in their own code because I'm using old config.

And I don't want to run `composer update` and make it fail everytime because a package is badly configured.

So I might need the "new" configuration for _every_ package that is covered by a recipe.

### Proposal

My idea is to create a command in the Flex plugin to do something like this:

```bash
$ composer flex:recipe:reset [--diff] [--overwrite] [packages...]
```

This command would download the full recipe of the package's currently installed version, and may, depending on the best / easily doable solutions:

* Maybe uninstall & reinstall recipe files. The advantage is that it would avoid having files that are no longer used by the new version of the recipe.
* Just reinstall the recipe without removing old files, but instead, overwrite all files of the recipe that have to be installed.
* Maybe compare files, it seems like a very hard challenge, but actually, `apt` is doing so when package config files are updated in your `/etc` folder (let's say, `php.ini` 😄)

Any of these options can be harmless for the end-user when using versionning, because any `git diff` in CLI allow seeing the changes, any visual Git diff application (like the PHPStorm's bundled one) can help debugging the upgrade changes and adapt them to our needs.
And as said before, this diff can also be handled directly in the command line, even though it's not the best option to me, but it has the advantage of being used programmatically by any CI or even by symfony.sh's bot for testing purposes.

Maybe other ideas about this might come, but with this first shot, what do you think?

Are the use cases relevant enough?

Pierstoval:
Closes #179

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,7 +15,8 @@
     },
     "require-dev": {
         "composer/composer": "^1.0.2",
-        "symfony/phpunit-bridge": "^3.4.19|^4.1.8"
+        "symfony/phpunit-bridge": "^3.4.19|^4.1.8",
+        "symfony/process": "^2.7|^3.0|^4.0"
     },
     "autoload": {
         "psr-4": {
--- a/src/Configurator/CopyFromPackageConfigurator.php
+++ b/src/Configurator/CopyFromPackageConfigurator.php
@@ -19,11 +19,11 @@ use Symfony\Flex\Recipe;
  */
 class CopyFromPackageConfigurator extends AbstractConfigurator
 {
-    public function configure(Recipe $recipe, $config)
+    public function configure(Recipe $recipe, $config, array $options = [])
     {
         $this->write('Setting configuration and copying files');
         $packageDir = $this->composer->getInstallationManager()->getInstallPath($recipe->getPackage());
-        $this->copyFiles($config, $packageDir, getcwd());
+        $this->copyFiles($config, $packageDir, getcwd(), $options['force'] ?? false);
     }
 
     public function unconfigure(Recipe $recipe, $config)
@@ -33,12 +33,12 @@ class CopyFromPackageConfigurator extends AbstractConfigurator
         $this->removeFiles($config, $packageDir, getcwd());
     }
 
-    private function copyFiles(array $manifest, string $from, string $to)
+    private function copyFiles(array $manifest, string $from, string $to, bool $overwrite = false)
     {
         foreach ($manifest as $source => $target) {
             $target = $this->options->expandTargetDir($target);
             if ('/' === substr($source, -1)) {
-                $this->copyDir($this->path->concatenate([$from, $source]), $this->path->concatenate([$to, $target]));
+                $this->copyDir($this->path->concatenate([$from, $source]), $this->path->concatenate([$to, $target]), $overwrite);
             } else {
                 $targetPath = $this->path->concatenate([$to, $target]);
                 if (!is_dir(\dirname($targetPath))) {
@@ -46,9 +46,7 @@ class CopyFromPackageConfigurator extends AbstractConfigurator
                     $this->write(sprintf('Created <fg=green>"%s"</>', $this->path->relativize(\dirname($targetPath))));
                 }
 
-                if (!file_exists($targetPath)) {
-                    $this->copyFile($this->path->concatenate([$from, $source]), $targetPath);
-                }
+                $this->copyFile($this->path->concatenate([$from, $source]), $targetPath, $overwrite);
             }
         }
     }
@@ -69,7 +67,7 @@ class CopyFromPackageConfigurator extends AbstractConfigurator
         }
     }
 
-    private function copyDir(string $source, string $target)
+    private function copyDir(string $source, string $target, bool $overwrite)
     {
         if (!is_dir($target)) {
             mkdir($target, 0777, true);
@@ -84,14 +82,14 @@ class CopyFromPackageConfigurator extends AbstractConfigurator
                     $this->write(sprintf('Created <fg=green>"%s"</>', $this->path->relativize($targetPath)));
                 }
             } elseif (!file_exists($targetPath)) {
-                $this->copyFile($item, $targetPath);
+                $this->copyFile($item, $targetPath, $overwrite);
             }
         }
     }
 
-    public function copyFile(string $source, string $target)
+    public function copyFile(string $source, string $target, bool $overwrite = false)
     {
-        if (file_exists($target)) {
+        if (!$this->options->shouldWriteFile($target, $overwrite)) {
             return;
         }
 
--- a/src/Configurator/CopyFromRecipeConfigurator.php
+++ b/src/Configurator/CopyFromRecipeConfigurator.php
@@ -18,10 +18,10 @@ use Symfony\Flex\Recipe;
  */
 class CopyFromRecipeConfigurator extends AbstractConfigurator
 {
-    public function configure(Recipe $recipe, $config)
+    public function configure(Recipe $recipe, $config, array $options = [])
     {
         $this->write('Setting configuration and copying files');
-        $this->copyFiles($config, $recipe->getFiles(), getcwd());
+        $this->copyFiles($config, $recipe->getFiles(), getcwd(), $options['force'] ?? false);
     }
 
     public function unconfigure(Recipe $recipe, $config)
@@ -30,31 +30,31 @@ class CopyFromRecipeConfigurator extends AbstractConfigurator
         $this->removeFiles($config, $recipe->getFiles(), getcwd());
     }
 
-    private function copyFiles(array $manifest, array $files, string $to)
+    private function copyFiles(array $manifest, array $files, string $to, bool $overwrite = false)
     {
         foreach ($manifest as $source => $target) {
             $target = $this->options->expandTargetDir($target);
             if ('/' === substr($source, -1)) {
-                $this->copyDir($source, $this->path->concatenate([$to, $target]), $files);
+                $this->copyDir($source, $this->path->concatenate([$to, $target]), $files, $overwrite);
             } else {
-                $this->copyFile($this->path->concatenate([$to, $target]), $files[$source]['contents'], $files[$source]['executable']);
+                $this->copyFile($this->path->concatenate([$to, $target]), $files[$source]['contents'], $files[$source]['executable'], $overwrite);
             }
         }
     }
 
-    private function copyDir(string $source, string $target, array $files)
+    private function copyDir(string $source, string $target, array $files, bool $overwrite = false)
     {
         foreach ($files as $file => $data) {
             if (0 === strpos($file, $source)) {
                 $file = $this->path->concatenate([$target, substr($file, \strlen($source))]);
-                $this->copyFile($file, $data['contents'], $data['executable']);
+                $this->copyFile($file, $data['contents'], $data['executable'], $overwrite);
             }
         }
     }
 
-    private function copyFile(string $to, string $contents, bool $executable)
+    private function copyFile(string $to, string $contents, bool $executable, bool $overwrite = false)
     {
-        if (file_exists($to)) {
+        if (!$this->options->shouldWriteFile($to, $overwrite)) {
             return;
         }
 
--- a/src/Configurator/EnvConfigurator.php
+++ b/src/Configurator/EnvConfigurator.php
@@ -18,7 +18,7 @@ use Symfony\Flex\Recipe;
  */
 class EnvConfigurator extends AbstractConfigurator
 {
-    public function configure(Recipe $recipe, $vars)
+    public function configure(Recipe $recipe, $vars, array $options = [])
     {
         $this->write('Added environment variable defaults');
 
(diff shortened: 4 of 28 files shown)


src/Configurator/CopyFromRecipeConfigurator.php
```diff
@@ -30,31 +30,31 @@ public function unconfigure(Recipe $recipe, $config)
...
+                $this->copyFile($file, $data['contents'], $data['executable'], $overwrite);
             }
         }
     }
 
-    private function copyFile(string $to, string $contents, bool $executable)
+    private function copyFile(string $to, string $contents, bool $executable, bool $overwrite = false)
     {
-        if (file_exists($to)) {
+        if (false === $overwrite && file_exists($to)) {
```
nicolas-grekas: if (!$overwrite && file_exists($to)) {

src/Configurator/EnvConfigurator.php
```diff
@@ -18,7 +18,7 @@
  */
 class EnvConfigurator extends AbstractConfigurator
 {
-    public function configure(Recipe $recipe, $vars)
+    public function configure(Recipe $recipe, $vars, array $contextParameters = [])
```
nicolas-grekas: maybe just "$options" or "$parameters" (should match getParameters/getOptions on UpdateEvent)

src/Configurator/CopyFromPackageConfigurator.php
```diff
@@ -33,7 +33,7 @@ public function unconfigure(Recipe $recipe, $config)
         $this->removeFiles($config, $packageDir, getcwd());
     }
 
-    private function copyFiles(array $manifest, string $from, string $to)
+    private function copyFiles(array $manifest, string $from, string $to, array $contextParameters = [])
```
maxhelias: You do not use override here as for CopyFromRecipeConfigurator.php ?

src/Configurator/EnvConfigurator.php
Pierstoval: Depending on the configurators, the previous argument can be called `$config` or `$parameters` , so there might be potential name conflicts :confused:

Pierstoval:
> Either we use a single `--force` parameter that does both (ignore symfony.lock an overwrite existing files)
> Or we create two separate arguments if worth it.

What about `[-f] --force` and `[-l] --ignore-lock`?

src/Configurator/CopyFromPackageConfigurator.php
Pierstoval: I should, yes :smile:

src/Event/UpdateEvent.php
```diff
@@ -0,0 +1,33 @@
...
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Flex\Event;
+
+use Composer\Script\Event;
+
+class UpdateEvent extends Event
+{
+    private static $defaultParameters = [
```
yceruto: I'd suggest setting the default value directly in the property bellow and getting rid of this, later, in constructor:
```php
$this->parameters = $parameters + $this->parameters;
```

Pierstoval: The array has been removed in favor of a boolean flag :)

src/Configurator/ModifyFileTrait.php
```diff
@@ -0,0 +1,42 @@
...
+            return true;
+        }
+
+        if (!$overwrite) {
+            return false;
+        }
+
+        return
+            ($filePath = $this->getFilePathIfModified($file))
+            && !$this->io->askConfirmation("File $filePath exists and has modifications, overwrite?")
```
nicolas-grekas: `sprintf('File "%s" has uncommitted changes, overwrite?', $filePath)`
+ confirmation should default to `N`

src/Configurator/ModifyFileTrait.php
```diff
@@ -0,0 +1,42 @@
...
+        ;
+    }
+
+    private function getFilePathIfModified(string $file): string
+    {
+        exec('git status --short --ignored -- '.escapeshellarg($file).' 2>&1', $output, $status);
+
+        if (0 !== $status) {
+            return basename($file);
```
nicolas-grekas: basename should be removed

src/Configurator/ModifyFileTrait.php
```diff
@@ -0,0 +1,42 @@
...
+        }
+
+        return substr($output[0], 0, 3);
```
nicolas-grekas: `substr($output[0], 3);`

tests/FlexTest.php
```diff
@@ -119,6 +119,9 @@ public function testPostInstall()
         $lock = $this->getMockBuilder(Lock::class)->disableOriginalConstructor()->getMock();
         $lock->expects($this->any())->method('has')->will($this->returnValue(false));
 
+        $updateEvent = $this->getMockBuilder(UpdateEvent::class)->disableOriginalConstructor()->getMock();
+        $updateEvent->expects($this->once())->method('getParameters')->willReturn([]);
```
nicolas-grekas: method doesn't exist anymore

tests/FlexTest.php
```diff
@@ -119,6 +119,9 @@ public function testPostInstall()
         $lock = $this->getMockBuilder(Lock::class)->disableOriginalConstructor()->getMock();
         $lock->expects($this->any())->method('has')->will($this->returnValue(false));
 
+        $updateEvent = $this->getMockBuilder(UpdateEvent::class)->disableOriginalConstructor()->getMock();
```
nicolas-grekas: can't we get rid of the mock?

Pierstoval:
Everything should be fixed now 👍

Pierstoval:
@nicolas-grekas Tests are fixed, but I have an issue with Windows, I'm sure it's mostly because of the tests and not Flex itself, but it seems that `git` behaves differently on Linux and Windows in certain cases. This might introduce edge cases for Windows users... but I'm not even sure that this would happen.

For example, here's a difference in terms of exit code: 

On windows:
```
> php -r "exec('git branch', $o, $c); var_dump($o, $c);"
fatal: not a git repository (or any of the parent directories): .git
Command line code:1:
array(0) {
}
Command line code:1:
int(0)
```

On linux:

```
$ php -r "exec('git branch', \$o, \$c); var_dump(\$o, \$c);"
fatal: Not a git repository (or any of the parent directories): .git
array(0) {
}
int(128)
```

In the tests, this is caused by the fact that the test files are located in `/tmp`. I'm thinking about creating a `build/` directory covered by `.gitignore` to put the files there (as if they're ignored we can expect the configurator to ask the user whether to overwrite the file or not), but not sure.

Should I change the tests so they use files directly in the project so it's covered by Git? Or is there anything to do that would make the tests compatible with both Linux and Windows? Can we put the Flex repo on AppVeyor? 😄

src/Flex.php
```diff
@@ -299,7 +300,7 @@ public function install(Event $event)
         $this->update($event);
     }
 
-    public function update(Event $event, $operations = [])
+    public function update(UpdateEvent $event, $operations = [])
```
nicolas-grekas: This should be reverted: Composer itself calls update() when triggering events and doesn't know about UpdateEvent.

src/Flex.php
```diff
@@ -387,7 +388,9 @@ function ($value) {
             switch ($recipe->getJob()) {
                 case 'install':
                     $this->io->writeError(sprintf('  - Configuring %s', $this->formatOrigin($recipe->getOrigin())));
-                    $this->configurator->install($recipe);
+                    $this->configurator->install($recipe, [
+                        'force' => $event->force(),
```
nicolas-grekas: `'force' => $event instanceof UpdateEvent && $event->force(),`

src/Configurator/ModifyFileTrait.php
```diff
@@ -0,0 +1,70 @@
...
+        if (!file_exists($file)) {
+            return true;
+        }
+
+        if (!$overwrite) {
+            return false;
+        }
+
+        if (!$this->gitExists()) {
+            return (bool) $this->io->askConfirmation(\sprintf('Cannot determine the state of the "%s" file, overwrite anyway? [y/N] ', $file), false);
```
nicolas-grekas: `\` should be removed

src/Configurator/ModifyFileTrait.php
```diff
@@ -0,0 +1,70 @@
...
+        if (0 === $status) {
+            $basename = basename($file);
+
+            $filePath = isset($output[0]) && $output[0]
+                ? (false !== strpos($output[0], $basename, -\strlen($basename)) ? substr($output[0], 3) : $file)
+                : ''
+            ;
+        }
+
+        if (\file_exists($filePath ?: $file)) {
```
nicolas-grekas: `\` should be removed

src/Configurator/ModifyFileTrait.php
```diff
@@ -0,0 +1,70 @@
...
+            return (bool) $this->io->askConfirmation(\sprintf('File "%s" has uncommitted changes, overwrite? [y/N] ', $filePath ?: $file), false);
```
nicolas-grekas: `\` should be removed (there may be more places, please check)

nicolas-grekas:
Except for some minor changes, I played with the feature and this looks great to me!

The experience could be improved, eg:
- files that are in the stash could be overridden without confirmation
- there could be more options than just `y/N` when asking confirmation, eg show the diff, or edit the patch before applying, or approve all or just quit or `?` to display help (very much like what `git add -p` does).

BUT this would be for future PRs.

:+1:

Pierstoval:
I fixed the reviews! For the next features, I might create other PRs targeted on this one so we could work on this in parallel without the need for tons of new releases, if you like :slightly_smiling_face:

src/Configurator/ModifyFileTrait.php
```diff
@@ -0,0 +1,70 @@
...
+        if (0 === $status) {
+            $basename = basename($file);
+
+            $filePath = isset($output[0]) && $output[0]
+                ? (false !== strpos($output[0], $basename, -\strlen($basename)) ? substr($output[0], 3) : $file)
+                : ''
+            ;
+        }
+
+        if (file_exists($filePath ?: $file)) {
```
nicolas-grekas: file_exists($file) would be enough, isn't it? if yes, let's kiss

tests/FlexTest.php
```diff
@@ -119,6 +119,8 @@ public function testPostInstall()
         $lock = $this->getMockBuilder(Lock::class)->disableOriginalConstructor()->getMock();
         $lock->expects($this->any())->method('has')->will($this->returnValue(false));
 
+        $updateEvent = new UpdateEvent(false);
```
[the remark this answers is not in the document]
Pierstoval: You mean I should change this for `new Event()`?

src/Configurator/ModifyFileTrait.php
Pierstoval: `$filePath` can be null if the status code is different than 0, in which case we simply fallback to _"If exists, confirm removal"_ instead of _"If has modifications, confirm removal"_.

I could change to `if ((0 !== $status && file_exists($file)) || file_exists($filePath)) ` though, this should produce the same, but I'm not sure this simplifies things 😕

tests/FlexTest.php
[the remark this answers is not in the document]
nicolas-grekas: or the change just be reverted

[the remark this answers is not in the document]
Pierstoval: Ok let's do this 👍

nicolas-grekas:
2nd commit fixes https://github.com/symfony/symfony/issues/29510
(fabbot failure is a false-positive)

composer.json
```diff
@@ -15,7 +15,8 @@
     },
     "require-dev": {
         "composer/composer": "^1.0.2",
-        "symfony/phpunit-bridge": "^3.2.8"
+        "symfony/phpunit-bridge": "^3.4.19",
+        "symfony/process": "^2.8"
```
nicolas-grekas: that's what composer forces already (because it's PHP 5.3 compatible)

fabpot:
Needs a rebase now that the commands has been renamed.

nicolas-grekas:
Thank you @Pierstoval.

PR #441, opened 2018-12-06
Do not configure project if Flex is not explicitly required

### LINKED ISSUE #440: create-project command modifies the composer.json file when installed as global dependency
When Flex is installed as a global dependency (to take advantage of its parallel downloader), `composer create-project` modifies the `composer.json` file.

For example, running the following:

```
composer create-project ezsystems/ezplatform
```

will:

1) remove author names from `authors` section, making the `composer.json` file invalid
2) remove `name` and `description` from the file
3) Change license to `proprietary`

Flex should not modify the `composer.json` file if it is not specified as a hard dependency in the project itself.

emodric:
Fixes #440

### DIFF
--- a/src/Flex.php
+++ b/src/Flex.php
@@ -239,6 +239,12 @@ class Flex implements PluginInterface, EventSubscriberInterface
 
     public function configureProject(Event $event)
     {
+        if (null === $this->downloader->getEndpoint()) {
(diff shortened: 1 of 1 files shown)


fabpot:
Thank you @emodric.

PR #442, opened 2018-12-10
Rename fix-recipes to sync-recipes

nicolas-grekas:
Would make it a bit more obvious what the command is for IMHO.
Related to https://github.com/symfony/flex/pull/439


fabpot:
Thank you @nicolas-grekas.

PR #443, opened 2018-12-10
Copy extra.symfony.require when unpacking

nicolas-grekas:
eg when unpacking the test pack:

```diff
--- a/composer.json
+++ b/composer.json
@@ -5,11 +5,14 @@
         "php": "^7.1.3",
         "ext-ctype": "*",
         "ext-iconv": "*",
+        "symfony/browser-kit": "4.2.*",
         "symfony/console": "4.2.*",
+        "symfony/css-selector": "4.2.*",
         "symfony/dotenv": "4.2.*",
         "symfony/flex": "^1.1",
         "symfony/framework-bundle": "4.2.*",
-        "symfony/test-pack": "^1.0",
+        "symfony/panther": "*",
+        "symfony/phpunit-bridge": "4.2.*",
         "symfony/yaml": "4.2.*"
     },
```

### DIFF
--- a/src/Unpacker.php
+++ b/src/Unpacker.php
@@ -21,10 +21,12 @@ use Symfony\Flex\Unpack\Result;
 class Unpacker
 {
     private $composer;
+    private $resolver;
 
-    public function __construct(Composer $composer)
+    public function __construct(Composer $composer, PackageResolver $resolver)
     {
         $this->composer = $composer;
+        $this->resolver = $resolver;
     }
 
     public function unpack(Operation $op): Result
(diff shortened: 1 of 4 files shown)


src/Unpacker.php
```diff
@@ -53,7 +55,10 @@ public function unpack(Operation $op): Result
                     continue;
                 }
 
-                if (!$manipulator->addLink($package['dev'] ? 'require-dev' : 'require', $link->getTarget(), $link->getPrettyConstraint(), $op->shouldSort())) {
+                $constraint = $link->getPrettyConstraint();
+                $constraint = substr($this->resolver->parseVersion($link->getTarget(), $constraint, !$package['dev']), 1) ?: $constraint;
```
fabpot: Are we sure that removing the first char is safe in 100% of the cases?

nicolas-grekas: Yep: https://3v4l.org/29ujj

PR #444, opened 2018-12-10
Avoid remove author name

HugoSantiagoBecerraAdan:
Fixes part of #440

### DIFF
--- a/src/Flex.php
+++ b/src/Flex.php
@@ -268,7 +268,7 @@ class Flex implements PluginInterface, EventSubscriberInterface
 
         // 'name' and 'description' are only required for public packages
         // don't use $manipulator->removeProperty() for BC with Composer 1.0
-        $contents = preg_replace('{^\s*+"(?:name|description)":.*,$\n}m', '', $manipulator->getContents());
+        $contents = preg_replace(['{^\s*+"name":.*,$\n}m', '{^\s*+"description":.*,$\n}m'], '', $manipulator->getContents(), 1);
         file_put_contents($json->getPath(), $contents);
 
         $this->updateComposerLock();


nicolas-grekas:
Did you check #441? Can you explain what this changes? I don't get it by reading the code.

HugoSantiagoBecerraAdan:
Search for "name" and "description" just once each.

It does not continue searching for more matches of the same word if it is already found, to be sure only the root name is removed and keep the name on the author section.

Already check #441 and I'm trying to solve comment https://github.com/symfony/flex/issues/440#issuecomment-443721144

hectorprats:
I think this is still not a complete solution. You should not delete name / description in any case that is not a skeleton. But if you are going to do something like that, better to do:
```
$manipulator->removeMainKey('name');
$manipulator->removeMainKey('description');
$manipulator->getContents();
```

instead of doing a regexp

HugoSantiagoBecerraAdan:
@hectorprats It's not a complete solution but a partial solution just for https://github.com/symfony/flex/issues/440#issuecomment-443721144

In [`src/Flex.php` line 264](https://github.com/symfony/flex/blob/master/src/Flex.php#L264) there's a comment that says:
`// don't use $manipulator->removeProperty() for BC with Composer 1.0`

I guess also applys to `removeMainKey`.

hectorprats:
Ok, @HugoSantiagoBecerraAdan . But your solution does not solve the problem if there is no 'name' key and yes a subkey inside authors.

If I do a create-project to make a git clone & composer install, it can not erase things from the composer.json

In a skeleton it makes sense. But never in other cases.

stof:
A solution could be to use the manipulator if possible, and use the regex only as a fallback for older composer versions.

nicolas-grekas:
There is no need to look for a solution when there is no problem ;)
Let's keep the code simple.

nicolas-grekas:
Thank you @HugoSantiagoBecerraAdan.

hectorprats:
@nicolas-grekas It's one thing to keep the code simple, and another to do ugly things.

The system of using a regexp to cancel 2 keys of a json is very very ugly. And in the end composer is full of ugly things. For example (although this is another topic), there are dozens of if where you check 2 terms and first check the slow and then the fast.

Or like the other day. Everywhere in the code the labels are closed with </ info>, and if I make a PR to homogenize and remove the </>, someone prefers to leave the 2 forms.

Simple code? Use the methods created for it, and if they are not compatible with something, make that method compatible.

nicolas-grekas:
@hectorprats thanks for the opinion. Looking forward to your PR since you seem to care.

PR #446, opened 2018-12-11
Use ProcessUtils::escapeArgument() instead of escapeshellarg()

nicolas-grekas:


composer.json
```diff
@@ -15,7 +15,8 @@
     },
     "require-dev": {
         "composer/composer": "^1.0.2",
-        "symfony/phpunit-bridge": "^3.4.19|^4.1.8"
+        "symfony/phpunit-bridge": "^3.4.19|^4.1.8",
+        "symfony/process": "^2.8"
```
stof: shouldn't it be a non-dev requirement ? Otherwise, the method might be undefined at runtime.

nicolas-grekas: It cannot be a non-dev requirement because that would prevent using flex with symfony >=3 apps.
But flex always runs in a composer context where process 2.8 is always available.

stof: but that would break in case Composer updates that (composer itself is compatible with different versions of `symfony/process` in case you don't force it to install deps for PHP 5.3)

nicolas-grekas: The composer.json of Composer has config.platform.php set to 5.3.9, so that only process 2.8 is ever installed. Its "require" is misleading.

stof: the composer.lock and the platform config only impact the root project (so the composer phar). I know of other projects bringing in composer as a dependency. And then, newer versions can be used.

stof: btw, composer also has its own version of `escapeArgument` precisely because it could not stop using it easily, while they wanted to add support for newer Symfony versions

nicolas-grekas: I missed that one, fixed in https://github.com/symfony/flex/pull/447

Issue #243, opened 2017-12-13
Introduce config option for root directory


simshaun:
This is a continuation of #236 -- there are still many instances of `getcwd()`.

- https://github.com/symfony/flex/blob/master/src/Flex.php
  (copies `.env.dist to .env`)

- https://github.com/symfony/flex/blob/master/src/Configurator/CopyFromPackageConfigurator.php
- https://github.com/symfony/flex/blob/master/src/Configurator/CopyFromRecipeConfigurator.php
- https://github.com/symfony/flex/blob/master/src/Configurator/EnvConfigurator.php
- https://github.com/symfony/flex/blob/master/src/Configurator/GitignoreConfigurator.php
- https://github.com/symfony/flex/blob/master/src/Configurator/MakefileConfigurator.php

They all assume that the file/dir paths they need are relative to the dir composer is executed from. For my personal needs that's fine, but I'm wondering if some people could need that to be configurable.

For that, I think a new config option (e.g. something like `root-dir`) would need to be introduced.

Thoughts? I will open a PR if this is something others are interested in.

fabschurt:
@simshaun I so do agree with this feature request! The new SF 4 project structure is great, but I’m currently a bit annoyed by the fact that Flex can’t support another «Symfony root». I generally design my projects with the help of the Hexagonal/Clean architecture, so everything Symfony-related is generally stored into `src/Symfony/` and not directly into `src/`, so every time some file/directory is created via a Flex recipe, I have to manually move it to the right destination.

As a side note, I have the same problem with the `make:controller` console command from the *MakerBundle* for example: it will create the new controller in `src/Controller/` with an `App\Controller` namespace, whereas I would like it to be in `src/Symfony/Controller/` with the `App\Symfony\Controller` namespace. I’m not sure how much this is related to our main issue here, maybe not that much.

nicolas-grekas:
Implemented in https://github.com/symfony/flex/pull/448

PR #448, opened 2018-12-11
 Allow configuring the "extra.symfony.root-dir" of the Symfony app 

nicolas-grekas:
Fixes https://github.com/symfony/flex/issues/243
and makes path management a bit more flexible.

### DIFF
--- a/src/Configurator/CopyFromPackageConfigurator.php
+++ b/src/Configurator/CopyFromPackageConfigurator.php
@@ -23,14 +23,14 @@ class CopyFromPackageConfigurator extends AbstractConfigurator
     {
         $this->write('Setting configuration and copying files');
         $packageDir = $this->composer->getInstallationManager()->getInstallPath($recipe->getPackage());
-        $this->copyFiles($config, $packageDir, getcwd(), $options['force'] ?? false);
+        $this->copyFiles($config, $packageDir, $this->options->get('root-dir'), $options['force'] ?? false);
     }
 
     public function unconfigure(Recipe $recipe, $config)
     {
         $this->write('Removing configuration and files');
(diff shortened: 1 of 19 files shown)


src/Configurator/CopyFromPackageConfigurator.php
```diff
@@ -97,7 +97,7 @@ public function copyFile(string $source, string $target, bool $overwrite = false
             throw new LogicException(sprintf('File "%s" does not exist!', $source));
         }
 
-        copy($source, $target);
+        file_put_contents($target, $this->options->expandTargetDir(file_get_contents($source)));
```
nicolas-grekas: This replaces eg `%CONFIG_DIR%` in files copied from packages (same for recipes below).
Will allow replacing `config/bootstrap.php` by `%CONFIG_DIR%/bootstrap.php` in a few months when everyone will have moved to a more recent flex version.

src/Flex.php
```diff
@@ -629,14 +629,17 @@ private function fetchRecipes(): array
...
+
         $options = array_merge([
             'bin-dir' => 'bin',
             'conf-dir' => 'conf',
             'config-dir' => 'config',
             'src-dir' => 'src',
             'var-dir' => 'var',
             'public-dir' => 'public',
-        ], $this->composer->getPackage()->getExtra());
+            'root-dir' => $extra['symfony']['root-dir'] ?? '.',
```
Pierstoval: How is `.` resolved when using it? Is it related to the `composer.json` file by composer or something?

nicolas-grekas: It's the same as using getenv()

Pierstoval: You mean `getcwd()` I guess 😄

PR #449, opened 2018-12-12
Add "dump-env" command to compile .env files to .env.local.php

nicolas-grekas:
This adds a new command that compiles .env files to .env.local.php:
```
$ composer dump-env prod
```

The generated file is used in https://github.com/symfony/recipes/pull/501


src/Command/DumpEnvCommand.php
```diff
@@ -0,0 +1,90 @@
...
+        parent::__construct();
+    }
+
+    protected function configure()
+    {
+        $this->setName('symfony:dump-env')
+            ->setAliases(['dump-env'])
+            ->setDescription('Compiles .env files to .env.local.php.')
+            ->setDefinition([
+                new InputArgument('env', InputArgument::REQUIRED, 'The application environment to dump .env files for.'),
```
weaverryan: > The application environment to dump the .env files for - e.g. "prod"

nicolas-grekas: updated thanks.

lyrixx:
Why is this a composer command? IMHO it should be a Symfony command (`bin/console env:dump prod`)

nicolas-grekas:
At least because we want to ship this for all Symfony versions without waiting for 4.3.

lyrixx:
This seems a hack to me. If we go in that sens, we will move many code here.
I understand the reason, but I don't like the implementation :/ 
What about a `symfony/backport` ?

Of course I'm not strongly against this hack, but if we could find a better alternative, this will be much better

nicolas-grekas:
Please, not another package to maintain. If you want, you can provide a command to 4.3 and then we could deprecate this one. Personally, I don't think it's worth the effort. `dump:autoload` is a Composer command, so would be `dump:env`. Both are run at the same stage, looks consistent to me.

nicolas-grekas:
Thank you @renanbr.

Bilge:
I cannot understand the point of this at all. All it does is substitute the `APP_ENV` var for whatever name you specify (e.g. `prod`) and pretends that all other dev settings will somehow be suitable for that environment. That is almost never going to be the case, particularly if you have separate dev and production databases, which is fairly common. How are you expected to merge the rest of the production settings into this config, because this command doesn't appear to support that?

PR #451, opened 2018-12-13
Lock copied files to prevent removing files needed by other recipes

### LINKED ISSUE #428: Handling of already existing files when requiring/removing a pack
Hi there,

I've seen multiple issues opened on the Encore repository that were caused by an already existing `package.json` file when requiring the related pack (the most recent one being https://github.com/symfony/webpack-encore/issues/411).

Since the recipe [adds this file](https://github.com/symfony/recipes/blob/0516ca401683a82fb37b27296fd4f2803b257d54/symfony/webpack-encore-pack/1.0/manifest.json#L5) using the `copy-from-recipe` configurator it fails silently when the user already has one (which may not be an uncommon situation and means that Encore won't be installed). 

Also annoying: if the user decides to remove the pack it will also remove the `package.json` file (which has probably been modified by then) without any warning.

The first issue could probably be handled by adding a message if `copy-from-recipe` wasn't able to copy a file for this reason, but:

* I'm not sure that it will be enough for the users to know what they have to do (even though we could add something in the doc about it I'm afraid it won't be *that* obvious)
* There may be some situations where that message would probably not be needed and have the opposite effect (people being worried about a warning message that they could safely ignore)

For the second problem, `package.json` could probably be ignored in the `removeFiles` method [the same way it's currently done for the ".git" repository](https://github.com/symfony/flex/blob/c218d231f8dba262e6b5c99a5e720861864ece12/src/Configurator/CopyFromRecipeConfigurator.php#L78), but I'm not sure that adding something that specific to a recipe into Flex's code would be a good idea...

dbrumann:
This PR modifies the `CopyFromRecipeConfigurator` so that it will store a list of copied files in symfony.lock and when a recipe is about to be unconfigured, it will check the list of locked files to prevent removal. This should fix #428 

## How to reproduce the issue

```
composer create-project symfony/skeleton example
cd example
composer remove console
```

This will remove both files installed through the recipe `bin/console` and `config/bootstrap.php`. As `symfony/framework-bundle` relies on the bootstrap-file as well, when trying to access the website, you will get an error due to the missing file.

## How to verify the fix

Create a new project based on `symfony/skeleton` and replace flex inside the example project with a local copy of this branch:

```
composer create-project symfony/skeleton example
cd example
rm -rf vendor/symfony/flex
ln -s path/to/local/flex vendor/symfony/flex
```

Since the `symfony.lock` was created during setup of the project it will not yet keep track of the copied files and needs to be updated first. An easy way to do this, is to remove the current file and run `composer fix-recipes` to create a new one:

```
rm symfony.lock
composer fix-recipes
```

Alternatively removing a recipe and then requiring it again will update the lock as well, but it needs to be done for each recipe, as otherwise the list of locked files would be incomplete.

Now when we remove the console again, opening the page in the browser should still work and the file `config/bootstrap.php` should still exist as it is locked by the framework-bundle.

## BC Breaks

None, as far as I can tell. For newly required recipes the files will be tracked. The new key for files in `symfony.lock` was not previously used and should not interfere with existing behavior.

## Known limitations

Unfortunately if any of the previously installed recipes has conflicting files, this will not be recognized unless the recipe is removed and then installed again.

### DIFF
--- a/src/Configurator/CopyFromPackageConfigurator.php
+++ b/src/Configurator/CopyFromPackageConfigurator.php
@@ -12,6 +12,7 @@
 namespace Symfony\Flex\Configurator;
 
 use LogicException;
+use Symfony\Flex\Lock;
 use Symfony\Flex\Recipe;
 
 /**
@@ -19,26 +20,29 @@ use Symfony\Flex\Recipe;
  */
 class CopyFromPackageConfigurator extends AbstractConfigurator
 {
-    public function configure(Recipe $recipe, $config, array $options = [])
+    public function configure(Recipe $recipe, $config, Lock $lock, array $options = [])
     {
         $this->write('Setting configuration and copying files');
         $packageDir = $this->composer->getInstallationManager()->getInstallPath($recipe->getPackage());
-        $this->copyFiles($config, $packageDir, $this->options->get('root-dir'), $options['force'] ?? false);
+        $options = array_merge($this->options->toArray(), $options);
+
+        $this->copyFiles($config, $packageDir, $options);
     }
 
-    public function unconfigure(Recipe $recipe, $config)
+    public function unconfigure(Recipe $recipe, $config, Lock $lock)
     {
         $this->write('Removing configuration and files');
         $packageDir = $this->composer->getInstallationManager()->getInstallPath($recipe->getPackage());
         $this->removeFiles($config, $packageDir, $this->options->get('root-dir'));
     }
 
-    private function copyFiles(array $manifest, string $from, string $to, bool $overwrite = false)
+    private function copyFiles(array $manifest, string $from, array $options)
     {
+        $to = $options['root-dir'] ?? '.';
         foreach ($manifest as $source => $target) {
             $target = $this->options->expandTargetDir($target);
             if ('/' === substr($source, -1)) {
-                $this->copyDir($this->path->concatenate([$from, $source]), $this->path->concatenate([$to, $target]), $overwrite);
+                $this->copyDir($this->path->concatenate([$from, $source]), $this->path->concatenate([$to, $target]), $options);
             } else {
                 $targetPath = $this->path->concatenate([$to, $target]);
                 if (!is_dir(\dirname($targetPath))) {
@@ -46,7 +50,7 @@ class CopyFromPackageConfigurator extends AbstractConfigurator
                     $this->write(sprintf('Created <fg=green>"%s"</>', $this->path->relativize(\dirname($targetPath))));
                 }
 
-                $this->copyFile($this->path->concatenate([$from, $source]), $targetPath, $overwrite);
+                $this->copyFile($this->path->concatenate([$from, $source]), $targetPath, $options);
             }
         }
     }
@@ -67,7 +71,7 @@ class CopyFromPackageConfigurator extends AbstractConfigurator
         }
     }
 
-    private function copyDir(string $source, string $target, bool $overwrite)
+    private function copyDir(string $source, string $target, array $options)
     {
         if (!is_dir($target)) {
             mkdir($target, 0777, true);
@@ -82,13 +86,14 @@ class CopyFromPackageConfigurator extends AbstractConfigurator
                     $this->write(sprintf('Created <fg=green>"%s"</>', $this->path->relativize($targetPath)));
                 }
             } elseif (!file_exists($targetPath)) {
-                $this->copyFile($item, $targetPath, $overwrite);
+                $this->copyFile($item, $targetPath, $options);
             }
         }
     }
 
-    public function copyFile(string $source, string $target, bool $overwrite = false)
+    public function copyFile(string $source, string $target, array $options)
     {
+        $overwrite = $options['force'] ?? false;
         if (!$this->options->shouldWriteFile($target, $overwrite)) {
             return;
         }
--- a/src/Configurator/CopyFromRecipeConfigurator.php
+++ b/src/Configurator/CopyFromRecipeConfigurator.php
@@ -11,6 +11,7 @@
 
 namespace Symfony\Flex\Configurator;
 
+use Symfony\Flex\Lock;
 use Symfony\Flex\Recipe;
 
 /**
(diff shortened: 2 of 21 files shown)


nicolas-grekas:
(rebase needed)

nicolas-grekas:
Hi @dbrumann, friendly ping to know the status here :)

dbrumann:
Hi @nicolas-grekas. Sorry for not updating. I'm currently busy settling into a new client project.

Writing tests for this was a bit tricky due to the Lock-class requiring a file to write to. I think the rebase and the flexible root-dir made this a bit easier, but I had no time to confirm this. I don't feel comfortable mocking the calls to Lock as I want to ensure the file list is handled properly.

I won't have time until Wednesday next week though. I set the end of the first week of February as my personal deadline for this.

dbrumann:
Hey @nicolas-grekas I think this is ready. Let me know if something is missing for you.

src/Configurator/CopyFromPackageConfigurator.php
```diff
@@ -12,41 +12,45 @@
...
     {
         $this->write('Removing configuration and files');
         $packageDir = $this->composer->getInstallationManager()->getInstallPath($recipe->getPackage());
         $this->removeFiles($config, $packageDir, $this->options->get('root-dir'));
     }
 
-    private function copyFiles(array $manifest, string $from, string $to, bool $overwrite = false)
+    private function copyFiles(array $manifest, string $from, array $options)
     {
+        $to = $options['root-dir'] ?? getcwd();
```
nicolas-grekas: not sure about this: getcwd introduces external state - can't we remove it from the PR altogether?

nicolas-grekas: Hum, we use `'root-dir' => $extra['symfony']['root-dir'] ?? '.'` elsewhere so my argument is void :)
Shouldn't we use the same?

src/Configurator/CopyFromRecipeConfigurator.php
```diff
@@ -11,51 +11,105 @@
...
     }
 
-    private function copyFiles(array $manifest, array $files, string $to, bool $overwrite = false)
+    private function getRemovableFilesFromRecipeAndLock(Recipe $recipe, Lock $lock): array
     {
+        $lockedFiles = array_unique(
+            array_reduce(
+                array_map(
+                    function (array $package) {
+                        return $package['files'];
```
nicolas-grekas: looks a lot like array_column ;)

src/Configurator/CopyFromRecipeConfigurator.php
```diff
@@ -11,51 +11,105 @@
...
+                ),
+                function (array $carry, array $package) {
+                    return array_merge($carry, $package);
+                },
+                []
+            )
+        );
+
+        $removableFiles = $recipe->getFiles();
+        if (empty($lockedFiles)) {
```
nicolas-grekas: we prefer not using empty, here `if (!$lockedFiles) {` would do the job

src/Configurator/CopyFromRecipeConfigurator.php
```diff
@@ -11,51 +11,105 @@
...
+            return $removableFiles;
+        }
+
+        foreach ($removableFiles as $file => $data) {
+            if (\in_array($file, $lockedFiles)) {
```
nicolas-grekas: `, true` for strictness by default

tests/Configurator/EnvConfiguratorTest.php
```diff
@@ -41,18 +43,22 @@ public function testConfigure()
...
-            'APP_PARAGRAPH' => "foo\n\"bar\"\\t",
-            'DATABASE_URL' => 'mysql://root@127.0.0.1:3306/symfony?charset=utf8mb4&serverVersion=5.7',
-            'MAILER_URL' => 'null://localhost',
-            'MAILER_USER' => 'fabien',
-            '#1' => 'Comment 1',
-            '#2' => 'Comment 3',
-            '#TRUSTED_SECRET' => 's3cretf0rt3st"<>',
-            'APP_SECRET' => 's3cretf0rt3st"<>',
-        ]);
+        $configurator->configure(
```
nicolas-grekas: let's keep the CS when it's unrelated

tests/Configurator/EnvConfiguratorTest.php
```diff
@@ -111,27 +117,35 @@ public function testConfigure()
...
 
-        $configurator->configure($recipe, [
-            'APP_ENV' => 'test',
-            'APP_DEBUG' => '0',
-            '#1' => 'Comment 1',
-            '#2' => 'Comment 3',
-            '#TRUSTED_SECRET' => 's3cretf0rt3st',
-            'APP_SECRET' => 's3cretf0rt3st',
-        ]);
+        $configurator->configure(
```
nicolas-grekas: unneeded CS change

tests/Configurator/MakefileConfiguratorTest.php
```diff
@@ -112,13 +114,20 @@ public function testConfigureForce()
...
+        $lock = $this->getMockBuilder(Lock::class)->disableOriginalConstructor()->getMock();
+
+        $configurator->configure($recipe, $bundleLinesConfigure, $lock);
         file_put_contents($makefile, "\n# new content", \FILE_APPEND);
         $this->assertStringEqualsFile($makefile, $contentConfigure);
 
-        $configurator->configure($recipe, $bundleLinesForce, [
-            'force' => true,
-        ]);
+        $configurator->configure(
```
nicolas-grekas: please revert all unneeded CS changes

src/Configurator/CopyFromRecipeConfigurator.php
```diff
@@ -11,51 +11,96 @@
...
+            )
+        );
+
+        $removableFiles = $recipe->getFiles();
+
+        if (!$lockedFiles) {
+            return $removableFiles;
+        }
+
+        foreach ($removableFiles as $file => $data) {
```
xabbuh: What about iterating `$lockedFiles` instead? Since the filename already is the key `$removableFiles` that would just be a constant-time lookup.

src/Configurator/CopyFromRecipeConfigurator.php
```diff
@@ -11,51 +11,95 @@
...
+                array_column($lock->all(), 'files'),
+                function (array $carry, array $package) {
+                    return array_merge($carry, $package);
+                },
+                []
+            )
+        );
+
+        $removableFiles = $recipe->getFiles();
+        if (!$lockedFiles) {
```
xabbuh: not really needed anymore now

nicolas-grekas:
Thank you @dbrumann.

Issue #431, opened 2018-11-07
Flex breaks composer scripts


alexander-schranz:
When you have flex installed it breaks [composer scripts](https://getcomposer.org/doc/articles/scripts.md#executing-php-scripts):

Install [flex](https://github.com/symfony/symfony/blob/8d277ce3e594c196594170e4fbbc37be98ba4f73/.travis.yml#L210):

```bash
composer global require --no-progress --no-scripts --no-plugins symfony/flex dev-master
```

run `composer install` with following composer.json

```json
{
    "scripts": {
        "auto-scripts": [
            "@php -r 'echo 1;'"
        ],
        "post-install-cmd": [
            "@auto-scripts"
        ],
        "post-update-cmd": [
            "@auto-scripts"
        ]
    }
}
```

you get:

```bash
  [InvalidArgumentException]
  Command type "@php -r 'echo test;'" is not valid.
```

without flex this works as expected.

nicolas-grekas:
Would you be able to send a fix?

nicolas-grekas:
Couldn't you just use another name than "auto-scripts" in your not-flex project?

alexander-schranz:
@nicolas-grekas we did currently. But think ignoring autoscripts by the flex executor if they  are not in the symfony format will avoid other developers investigate time finding the error. Think we are not the only developer which tries to structure the composer.json in all projects the same way.

fabpot:
Ignoring auto-scripts not in the right format would be bad for DX as no errors would be raised.

alexander-schranz:
@fabpot thats not correct as the default composer script executor is called which will raise an error when it's not compatible with the composer script execution.

nicolas-grekas:
See my proposal in #459. It should help understand the issue earlier, which is the most important thing here to me.

PR #453, opened 2018-12-16
Fix using of @php scripts in autoscripts

alexander-schranz:
This will **not** fix other @ commands like reference calls:

```json
    "scripts": {
        "auto-scripts": [
            "@test-scripts"
        ],
        "test-scripts": [
        ],
     }
```

and also **not** fixing running `@composer` commands:

```json
    "scripts": {
        "auto-scripts": [
            "@composer something"
        ]
     }
```

fixes parts of #431

### DIFF
--- a/src/ScriptExecutor.php
+++ b/src/ScriptExecutor.php
@@ -43,6 +43,11 @@ class ScriptExecutor
      */
     public function execute(string $type, string $cmd)
     {
+        if (0 === strpos($type,'@php ')) {
+            $cmd = explode(' ', $type, 2)[1];
+            $type = 'php-script';
+        }
+
         $parsedCmd = $this->options->expandTargetDir($cmd);
         if (null === $expandedCmd = $this->expandCmd($type, $parsedCmd)) {
             return;
--- a/src/ScriptExecutor.php
+++ b/src/ScriptExecutor.php
@@ -43,7 +43,7 @@ class ScriptExecutor
      */
     public function execute(string $type, string $cmd)
     {
-        if (0 === strpos($type,'@php ')) {
+        if (0 === strpos($type, '@php ')) {
             $cmd = explode(' ', $type, 2)[1];
             $type = 'php-script';
         }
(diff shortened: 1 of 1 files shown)


src/ScriptExecutor.php
```diff
@@ -43,12 +43,17 @@ public function __construct(Composer $composer, IOInterface $io, Options $option
...
+            $type = 'php-script';
+        }
+
         $parsedCmd = $this->options->expandTargetDir($cmd);
         if (null === $expandedCmd = $this->expandCmd($type, $parsedCmd)) {
             return;
         }
 
-        $cmdOutput = new StreamOutput(fopen('php://temp', 'rw'), OutputInterface::VERBOSITY_VERBOSE, $this->io->isDecorated());
+        $cmdOutput = new StreamOutput(fopen('php://temp', 'rwb'), OutputInterface::VERBOSITY_VERBOSE, $this->io->isDecorated());
```
alexander-schranz: this was changed by the code style fixer

alexander-schranz:
@nicolas-grekas not sure how to fix this issue as there are other cases which currnetly don't work as `@php`. The could be `@composer` or `@customscript` but don't know how to call composer directly with it.

alexander-schranz:
@fabpot I don't think so when a look at the [composer documentation](https://getcomposer.org/doc/articles/scripts.md#command-events)

fabpot:
ok, so I don't understand what we are trying to fix here.

alexander-schranz:
@fabpot we did run into this problems on a project which doesn't use flex but a developer had flex installed globally and got this [strange error message](https://github.com/symfony/flex/blob/ef8b60eed1d51aae05feb79bb60edc52f55c8142/src/ScriptExecutor.php#L89).

At least we should change the error message that `flex` does not support running this type of scripts. In our errors case supporting for `@php` would be enough but just wanted to mention that there are also other cases which are documented in the composer docs which don't work when used inside auto-scripts in symfony flex documentation.

fabpot:
`@auto-scripts` has a special syntax that you should use. Don't use it with the "standard" syntax from Composer (for which you can use another alias).

PR #458, opened 2018-12-19
Unconfigure .gitignore before reconfiguring it with "sync-recipes --force"

### LINKED ISSUE #457: sync:recipes --force does not update .gitignore entries
When testing the `sync:recipes --force`, my `.gitignore` file WAS updated to add *new* packages that previously did not have any `.gitignore` rules and now have some.

But, for FrameworkBundle (for example), I DID already have some entries in `.gitignore` from its recipe. However, those entries were *not* updated.

To replicate

```
git clone git@github.com:symfony/symfony-demo.git
cd symfony-demo
git checkout v1.2.7
composer install
rm -rf vendor/symfony/flex
git clone git@github.com:symfony/flex.git vendor/symfony/flex
composer sync-recipes --force
```

Pierstoval:
Fixes #457 

This is the most straightforward way to force sync for gitignore: remove all rules & re-add them.

### DIFF
--- a/src/Configurator/GitignoreConfigurator.php
+++ b/src/Configurator/GitignoreConfigurator.php
@@ -24,6 +24,10 @@ class GitignoreConfigurator extends AbstractConfigurator
 
         $gitignore = $this->options->get('root-dir').'/.gitignore';
         if ($this->isFileMarked($recipe, $gitignore)) {
+            if ($options['force'] ?? false) {
+                $this->doUnconfigure($recipe, true);
+            }
+
             return;
         }
 
@@ -37,6 +41,11 @@ class GitignoreConfigurator extends AbstractConfigurator
 
     public function unconfigure(Recipe $recipe, $vars)
     {
+        $this->doUnconfigure($recipe);
+    }
+
+    private function doUnconfigure(Recipe $recipe, bool $silent = false)
+    {
         $file = $this->options->get('root-dir').'/.gitignore';
         if (!file_exists($file)) {
             return;
@@ -47,7 +56,9 @@ class GitignoreConfigurator extends AbstractConfigurator
             return;
         }
 
-        $this->write('Removed entries in .gitignore');
+        if (!$silent) {
+            $this->write('Removed entries in .gitignore');
+        }
         file_put_contents($file, ltrim($contents, "\r\n"));
     }
 }


src/Configurator/GitignoreConfigurator.php
```diff
@@ -36,6 +40,11 @@ public function configure(Recipe $recipe, $vars, array $options = [])
     }
 
     public function unconfigure(Recipe $recipe, $vars)
+    {
+        $this->doUnconfigure($recipe);
+    }
+
+    private function doUnconfigure(Recipe $recipe, bool $silent = false)
```
nicolas-grekas: not sure we need another method, let's add `bool $silent = false` to unconfigure.

Pierstoval: I'll answer you the same thing as when we discussed about the addition of `array $options = []` to all the `configure()` methods: this one is an edge case, but if configurators need more metadata one day, it's better to choose between adding `array $options = []` in **all** `unconfigure()` methods, or just keep it as-is for consistency and use an internal flag.

For the initial feature we added the options because it was used in more than one configurator.
Here, the "silent" is only one edge case (and not used in other configurators, at least not yet), so IMO it's better to keep it as-is than update all `unconfigure()` methods.

WDYT?

nicolas-grekas:
We should do the same for configurator that handled markers.

That means:
- GitignoreConfigurator
- MakefileConfigurator
- EnvConfigurator

Ideally, we would preserve the order of sections to reduce the diff.

Pierstoval:
> Maybe we could make these configurators have a `reconfigure` method, which would solve the order preserving, as we would not remove the markers entirely before calling the configure method again.

I totally agree with this idea.

A default `reconfigure` method could be implemented, like a generic "replace all files", but some may ask for something else (like merging things, or executing other things).

WDYT?

nicolas-grekas:
friendly ping @Pierstoval

Pierstoval:
Thanks for the ping, I wish I could continue to work on this but I'm not sure about where to go.

Should I add the `reconfigure` method in the configurators? 😄

OskarStark:
> Should I add the reconfigure method in the configurators? 😄

I really like the idea 👍

nicolas-grekas:
Continued in  #469, thanks @Pierstoval

PR #459, opened 2018-12-19
Improve error when an invalid auto-script is found

### LINKED ISSUE #431: Flex breaks composer scripts

nicolas-grekas:
Fixes #431. Replaces #455 and #453.

### DIFF
--- a/src/ScriptExecutor.php
+++ b/src/ScriptExecutor.php
@@ -86,7 +86,7 @@ class ScriptExecutor
             case 'script':
                 return $cmd;
             default:
-                throw new \InvalidArgumentException(sprintf('Command type "%s" is not valid.', $type));
+                throw new \InvalidArgumentException(sprintf('Invalid symfony/flex auto-script in composer.json: "%s" is not a valid type of command.', $type));
         }
     }
 


src/ScriptExecutor.php
```diff
@@ -86,7 +86,7 @@ private function expandCmd(string $type, string $cmd)
             case 'script':
                 return $cmd;
             default:
-                throw new \InvalidArgumentException(sprintf('Command type "%s" is not valid.', $type));
+                throw new \InvalidArgumentException(sprintf('Invalid symfony/flex auto-script in composer.json: "%s" is not a valid type of command.', $type));
```
alexander-schranz: would suggest the following:

`type for command "%s", $type, $cmd`

as the type in my case is 0, 1, 2, 3

nicolas-grekas: Are you sure? The type is the right side, so it's the string here to me. I also ran out he code to reproduce your situation and the error was fine.

alexander-schranz: @nicolas-grekas sorry you are correct!

PR #462, opened 2018-12-19
Do nothing when uninstalling unconfigured packages

HeahDude:
When uninstalling a package with a found recipe, Flex unconfigures it even if it was not configured before.


nicolas-grekas:
Thank you @HeahDude.

PR #464, opened 2018-12-20
Added support for recipes conflict

HeahDude:
Allows to ignore incompatible recipes, helpful to avoid breaking apps using FrameworkBundle LTS versions.

flex_conflict

Unlocks https://github.com/symfony/recipes/pull/506.

We need to update the Symfony bot to allow the `conflict` key in the `manifest.json`, ping @fabpot :).

### DIFF
--- a/src/Flex.php
+++ b/src/Flex.php
@@ -587,6 +587,18 @@ class Flex implements PluginInterface, EventSubscriberInterface
             $name = $package->getNames()[0];
             $job = $operation->getJobType();
 
+            if (!empty($manifests[$name]['manifest']['conflict']) && !$operation instanceof UninstallOperation) {
+                $lockedRepository = $this->composer->getLocker()->getLockedRepository();
+
+                foreach ($manifests[$name]['manifest']['conflict'] as $conflictingPackage => $constraint) {
+                    if ($lockedRepository->findPackage($conflictingPackage, $constraint)) {
+                        $this->io->writeError(sprintf('  - Skipping recipe for %s: it conflicts with %s %s.', $name, $conflictingPackage, $constraint), true, IOInterface::VERBOSE);
+
+                        continue 2;
+                    }
+                }
+            }
+
             if ($operation instanceof InstallOperation && isset($locks[$name])) {
                 $ref = $this->lock->get($name)['recipe']['ref'] ?? null;
                 if ($ref && ($locks[$name]['recipe']['ref'] ?? null) === $ref) {


src/Flex.php
```diff
@@ -298,7 +298,17 @@ public function update(Event $event, $operations = [])
             copy($rootDir.'/.env.dist', $rootDir.'/.env');
         }
 
-        $recipes = $this->fetchRecipes();
+        list($recipes, $conflicts) = $this->fetchRecipes();
+        if ($conflicts) {
+            $this->io->writeError(sprintf('<info>%d Conflict%s:</>', \count($conflicts), \count($conflicts) > 1 ? 's' : ''));
```
nicolas-grekas: I don't think any of the log messages are needed in this PR.
Which means all the logic can be moved inside fetchRecipes.
If we really want some log messages, it could be inside that method, in verbose mode only.

src/Flex.php
```diff
@@ -594,6 +605,17 @@ private function fetchRecipes(): array
...
 
             if (isset($manifests[$name])) {
+                if (isset($manifests[$name]['manifest']['conflict'])) {
+                    foreach ($manifests[$name]['manifest']['conflict'] as $conflictingPackageName => $constraint) {
+                        if ($conflictingPackage = $this->composer->getLocker()->getLockedRepository()->findPackage($conflictingPackageName, $constraint)) {
+                            $conflicts[$name][$conflictingPackageName] = [$conflictingPackage->getPrettyVersion(), $constraint];
+                        }
+                    }
+                    unset($manifests[$name]);
+                    $this->lock->remove($name);
```
nicolas-grekas: really? feels unrelated to me.

src/Flex.php
```diff
@@ -594,6 +605,17 @@ private function fetchRecipes(): array
             }
 
             if (isset($manifests[$name])) {
+                if (isset($manifests[$name]['manifest']['conflict'])) {
+                    foreach ($manifests[$name]['manifest']['conflict'] as $conflictingPackageName => $constraint) {
+                        if ($conflictingPackage = $this->composer->getLocker()->getLockedRepository()->findPackage($conflictingPackageName, $constraint)) {
+                            $conflicts[$name][$conflictingPackageName] = [$conflictingPackage->getPrettyVersion(), $constraint];
+                        }
+                    }
+                    unset($manifests[$name]);
```
nicolas-grekas: not needed.

src/Flex.php
```diff
@@ -594,6 +605,17 @@ private function fetchRecipes(): array
...
+                    $this->lock->remove($name);
+
+                    continue;
```
nicolas-grekas: why is it unconditional? shouldn't it be only when a conflict has been found?

src/Flex.php
HeahDude: If there is a conflict we need to revert that call, https://github.com/symfony/flex/pull/464/files#diff-46b78198ae7ea525f04268205dd782c3R602.

src/Flex.php
HeahDude: >shouldn't it be only when a conflict has been found?

It is already https://github.com/symfony/flex/pull/464/files#diff-46b78198ae7ea525f04268205dd782c3R608, no?

src/Flex.php
```diff
@@ -594,6 +594,18 @@ private function fetchRecipes(): array
             }
 
             if (isset($manifests[$name])) {
+                if (!empty($manifests[$name]['manifest']['conflict'])) {
+                    $lockedRepository = $this->composer->getLocker()->getLockedRepository();
+
+                    foreach ($manifests[$name]['manifest']['conflict'] as $conflictingPackageName => $constraint) {
+                        if ($conflictingPackage = $lockedRepository->findPackage($conflictingPackageName, $constraint)) {
+                            $this->io->writeError(sprintf('    <warning>Recipe for %s conflicts with %s:%s which should not be %s</>', $name, $conflictingPackageName, $conflictingPackage->getPrettyVersion(), $constraint));
+                            $this->lock->remove($name);
```
nicolas-grekas: A conflict should not change the state of the lock file.

src/Flex.php
```diff
@@ -594,6 +594,18 @@ private function fetchRecipes(): array
             }
 
             if (isset($manifests[$name])) {
+                if (!empty($manifests[$name]['manifest']['conflict'])) {
+                    $lockedRepository = $this->composer->getLocker()->getLockedRepository();
+
+                    foreach ($manifests[$name]['manifest']['conflict'] as $conflictingPackageName => $constraint) {
+                        if ($conflictingPackage = $lockedRepository->findPackage($conflictingPackageName, $constraint)) {
+                            $this->io->writeError(sprintf('    <warning>Recipe for %s conflicts with %s:%s which should not be %s</>', $name, $conflictingPackageName, $conflictingPackage->getPrettyVersion(), $constraint));
```
nicolas-grekas: That's now a warning - all is fine, keep cool :)
I'd just remove the call personally. Nobody expects a recipe to come. Things work reverse: you install and you see what *might* come, without knowing beforehand.

src/Flex.php
HeahDude: Got ya, fixed, thx

src/Flex.php
HeahDude: Got ya, fixed, thx

src/Flex.php
HeahDude: Done, without warning, on verbose

src/Flex.php
HeahDude: If verbose, without warning, ok for you?

src/Flex.php
HeahDude: I guess that a package should not be added to the lock if it's not configured because of a conflict

src/Flex.php
```diff
@@ -587,6 +587,20 @@ private function fetchRecipes(): array
             $name = $package->getNames()[0];
             $job = $operation->getJobType();
 
+            if (!empty($manifests[$name]['manifest']['conflict'])) {
+                $lockedRepository = $this->composer->getLocker()->getLockedRepository();
+
+                foreach ($manifests[$name]['manifest']['conflict'] as $conflictingPackageName => $constraint) {
+                    if ($conflictingPackage = $lockedRepository->findPackage($conflictingPackageName, $constraint)) {
+                        if ($this->io->isVerbose()) {
+                            $this->io->writeError(sprintf('    Recipe for %s conflicts with %s:%s which should not be %s', $name, $conflictingPackageName, $conflictingPackage->getPrettyVersion(), $constraint));
```
nicolas-grekas: Skipping recipe for %s: it conflicts with %s %s.

HeahDude: Done.

src/Flex.php
```diff
@@ -587,6 +587,20 @@ private function fetchRecipes(): array
             $name = $package->getNames()[0];
             $job = $operation->getJobType();
 
+            if (!empty($manifests[$name]['manifest']['conflict'])) {
+                $lockedRepository = $this->composer->getLocker()->getLockedRepository();
+
+                foreach ($manifests[$name]['manifest']['conflict'] as $conflictingPackageName => $constraint) {
+                    if ($conflictingPackage = $lockedRepository->findPackage($conflictingPackageName, $constraint)) {
+                        if ($this->io->isVerbose()) {
+                            $this->io->writeError(sprintf('  - Skipping recipe for %s: it conflicts with %s %s.', $name, $conflictingPackageName, $constraint));
```
weaverryan: What does the latest version of the message look like? I'm not sure exactly what each variable holds.

Also, does this only show in verbose mode? Does that mean it will not sure unless the user adds `-v`?

weaverryan: Ha! Ok, I feel the opposite :p. Well, sort of. Even if you assume that you don't know anything about recipes, a user expects that when they install a package, it will work just like if they follow the docs. In the case of Messenger, it means that the latest Messenger is being installed (so, ideally the user is following the latest messenger docs), but they do NO get the recipe that the docs expect. I don't see a lot of harm in showing it always - it's not a message that they will see over and over again.

I'd also like to add:
> If you want to install the recipe, fix the conflicting package(s) and run composer sync-recipes

nicolas-grekas: One wouldn't expect that in the use case that triggered the PR: if you install messenger with framework bundle, you *do not* expect it to be configured (because that's a feature of fwb >=4.1, which you decide not to use.)

Fixing a dep to get a recipe is really the reverse way of thinking: you first require deps, then you configure them, not the opposite.

weaverryan: Ok, good for me. I forgot that, for normal users, it's not possible to get a "mix" of symfony/* versions anyways, due to the locking on a specific minor thanks to Flex.

HeahDude:
Last comments adresses, thanks for the review guys.

@weaverryan I've updated the screenshot in the description.

fabpot:
Thank you @HeahDude.

PR #466, opened 2018-12-21
Add support for multiple bundles in auto-generated recipes

### LINKED ISSUE #427: Add support for multiple bundles in auto-configured packages
Currently, if a package has a `symfony-bundle` type and does not have a recipe, it is automatically added to `bundles.php`, extracted from PSR4 autoload config. However, this only works for the first bundle listed in PSR-4 autoload.

It would be great if Flex would automatically activate all bundles it finds in autoload config, not just the first one.

~Additionally, if some of them are listed in `autoload-dev` section, it would be great if those were automatically activated for `dev` environment too.~ Disregard this, this is not always wanted and most probably cannot be automated.

emodric:
Fixes #427

### DIFF
--- a/src/SymfonyBundle.php
+++ b/src/SymfonyBundle.php
@@ -32,7 +32,7 @@ class SymfonyBundle
 
     public function getClassNames(): array
     {
-        $all = 'uninstall' === $this->operation;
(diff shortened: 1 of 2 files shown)


fabpot:
Thank you @emodric.

PR #472, opened 2019-02-04
Fix file_get_content exception when installing symfony 4.3-dev

jlslew:
For more information, see issue [#523](https://github.com/symfony/recipes/issues/523) from the symfony/recipes repo.

### DIFF
--- a/src/Configurator/AbstractConfigurator.php
+++ b/src/Configurator/AbstractConfigurator.php
@@ -75,6 +75,10 @@ abstract class AbstractConfigurator
      */
     protected function updateData(string $file, string $data): bool
     {
+        if (!file_exists($file)) {
+            return false;
+        }
+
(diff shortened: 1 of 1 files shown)


src/Configurator/AbstractConfigurator.php
```diff
@@ -78,7 +78,7 @@ protected function updateData(string $file, string $data): bool
         $pieces = explode("\n", trim($data));
         $startMark = trim(reset($pieces));
         $endMark = trim(end($pieces));
-        $contents = file_get_contents($file);
+        $contents = @file_get_contents($file);
```
xabbuh: What about adding the following before instead?

```php
if (!file_exists($file)) {
    return false;
}
```

nicolas-grekas:
Thank you @jlslew.

PR #478, opened 2019-03-04
Ignore existing env vars when running dump-env

### LINKED ISSUE #475: Flex 1.2 / dump-env command and existing environment variables
Hi,

```
$ cat .env|wc -l
101

$ composer dump-env dev && cat .env.local.php 
<?php

// This file was generated by running "composer dump-env dev"

return array (
  'APP_ENV' => 'dev',
);
```

In the execution context of the `composer dump-env dev` command, all my env variables in my `.env` file are in fact already sourced (already present as env variables).

I would expect the `dump-env` command to not be impacted by what variables are already set in the env  and just dump whatever is in `.env*` to `.env.local.php`. Otherwise it leads to unexpected results (empty or partial env cache file generated).

Is that my mistake to have such env variables already sourced or is it an unexpected behavior of `dump-env`, or both ?

Thanks!

### LINKED ISSUE #476: Loading values from `.env.local.php` is not equivalent to `DotEnv::loadEnv`
Hi,

It seems that `composer dump-env` is not equivalent to `DotEnv::loadEnv` loading yaml files.

Please see this repository for a reproducer and the README for a description of the problem.
https://github.com/bendavies/symfony-flex-env-issue

Thanks

nicolas-grekas:
Fix #475
Fix #476
Fix https://github.com/symfony/symfony/pull/30418

### DIFF
--- a/src/Command/DumpEnvCommand.php
+++ b/src/Command/DumpEnvCommand.php
@@ -17,7 +17,7 @@ use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
-use Symfony\Component\Dotenv\Dotenv;
+use Symfony\Component\Process\PhpProcess;
 use Symfony\Flex\Options;
 
 class DumpEnvCommand extends BaseCommand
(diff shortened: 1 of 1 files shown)


bendavies:
didn't get a notification for this on my linked issue. thanks - i'll take a look.

DavG:
This introduces bug #500 

```
foreach ($_SERVER as $k => $v) {
    if (\is_string($v) && false !== getenv($k)) {
        unset($_SERVER[$k]);
        putenv($k);
    }
}
```

bendavies:
@nicolas-grekas I thought I checked but it doesn't not seem that this fixed the issue described at https://github.com/symfony/flex/issues/476. https://github.com/bendavies/symfony-flex-env-issue is still an issue.

bendavies:
looking at the fix in this PR, I cant see the issue was resolved actually. the problem was not with `dump-env`, but with the loading of var in `bootstrap.php`

PR #482, opened 2019-04-02
Fix bundles.php generation when false value is used

fferriere:
Fix when we set an environment to `false` in bundle.php it set at `true` on generation.


nicolas-grekas:
Thank you @fferriere.

PR #486, opened 2019-04-11
Removed CURLOPT_SSL_CIPHER_LIST which causes curl unknown cipher errors

phoenix1337:


nicolas-grekas:
Thank you @phoenix1337.

PR #489, opened 2019-04-15
Fix proxy handling

### LINKED ISSUE #488: "https://repo.packagist.org/packages.json" does not contain valid JSON
Similar issue as #484 when trying to `composer update`.
As the original issue is closed, I create this one.

It happens on Windows with symfony/flex v1.2.2, working behind a company HTTP Proxy.

```
λ composer show | grep "symfony/flex"
symfony/flex                        v1.2.2   Composer plugin for Symfony
```
Output of composer diagnose:
```
λ composer diagnose
Checking composer.json: OK
Checking platform settings: OK
Checking git settings: OK
Checking http connectivity to packagist: OK
Checking https connectivity to packagist: OK
Checking HTTP proxy: OK
Checking HTTP proxy support for request_fulluri: OK
Checking HTTPS proxy support for request_fulluri: OK
Checking github.com rate limit: OK
Checking disk free space: OK
Checking pubkeys:
Tags Public Key Fingerprint: 57815BA2 7E54DC31 7ECC7CC5 573090D0  87719BA6 8F3BB723 4E5D42D0 84A14642
Dev Public Key Fingerprint: 4AC45767 E5EC2265 2F0C1167 CBBB8A2B  0C708369 153E328C AD90147D AFE50952
OK
Checking composer version: OK
Composer version: 1.8.5
PHP version: 7.3.4
PHP binary path: C:\php\php.exe
```

When composer is run with --no-plugins option, it works:
```
λ composer update --no-plugins
Nothing to install or update
Generating autoload files
> symfony-cmd
'symfony-cmd' is not recognized as an internal or external command,
operable program or batch file.
Script symfony-cmd handling the auto-scripts event returned with error code 1
Script @auto-scripts was called via post-update-cmd
```

When composer is run without --no-plugins option, it doesn't:
```
λ composer update
"https://repo.packagist.org/packages.json" does not contain valid JSON
Parse error on line 1:

^
Expected one of: 'STRING', 'NUMBER', 'NULL', 'TRUE', 'FALSE', '{', '['
https://repo.packagist.org could not be fully loaded, package information was loaded from the local cache and may be out of date
"https://repo.packagist.org/packages.json" does not contain valid JSON
Parse error on line 1:

^
Expected one of: 'STRING', 'NUMBER', 'NULL', 'TRUE', 'FALSE', '{', '['
https://repo.packagist.org could not be fully loaded, package information was loaded from the local cache and may be out of date


  [Composer\Repository\RepositorySecurityException]
  The contents of https://repo.packagist.org/p/phpunit/phpunit%2409ce48554409f24a1144e25d466169aaf1eab60b5ca98df75998dd2cf14db9d4.json do not match its signature. This could indicate a man-in-the-
  middle attack or e.g. antivirus software corrupting files. Try running composer again and report this if you think it is a mistake.


update [--prefer-source] [--prefer-dist] [--dry-run] [--dev] [--no-dev] [--lock] [--no-custom-installers] [--no-autoloader] [--no-scripts] [--no-progress] [--no-suggest] [--with-dependencies] [--with-all-dependencies] [-v|vv|vvv|--verbose] [-o|--optimize-autoloader] [-a|--classmap-authoritative] [--apcu-autoloader] [--ignore-platform-reqs] [--prefer-stable] [--prefer-lowest] [-i|--interactive] [--root-reqs] [--] [<packages>]...
```

We work behind a company HTTP Proxy, which is already set on env variables:
```
https_proxy=http://x.x.x.x:3128
http_proxy=http://x.x.x.x:3128
```

Full verbose of Composer update command:
```
λ composer -vvv update
Reading ./composer.json
Loading config file C:/Users/[my_username]/AppData/Roaming/Composer/config.json
Loading config file C:/Users/[my_username]/AppData/Roaming/Composer/auth.json
Loading config file ./composer.json
Checked CA file C:\Users\[my_username]\AppData\Local\Temp\opeB424.tmp: valid
Executing command (C:\Users\[my_username]\Documents\projects\[project_name]): git branch --no-color --no-abbrev -v
Reading C:/Users/[my_username]/AppData/Roaming/Composer/composer.json
Loading config file C:/Users/[my_username]/AppData/Roaming/Composer/config.json
Loading config file C:/Users/[my_username]/AppData/Roaming/Composer/auth.json
Loading config file C:/Users/[my_username]/AppData/Roaming/Composer/composer.json
Loading config file C:\Users\[my_username]\AppData\Roaming\Composer/auth.json
Reading C:\Users\[my_username]\AppData\Roaming\Composer/auth.json
Reading C:\Users\[my_username]\Documents\projects\[project_name]/vendor/composer/installed.json
Loading plugin PackageVersions\Installer
Loading plugin Symfony\Flex\Flex
Downloading https://repo.packagist.org/packages.json using proxy tcp://x.x.x.x:3128
Downloading https://repo.packagist.org/packages.json using proxy tcp://x.x.x.x:3128
Downloading https://repo.packagist.org/packages.json using proxy tcp://x.x.x.x:3128
Reading C:/Users/[my_username]/AppData/Local/Composer/repo/https---repo.packagist.org/packages.json from cache
"https://repo.packagist.org/packages.json" does not contain valid JSON
Parse error on line 1:

^
Expected one of: 'STRING', 'NUMBER', 'NULL', 'TRUE', 'FALSE', '{', '['
https://repo.packagist.org could not be fully loaded, package information was loaded from the local cache and may be out of date
```

Lctrs:
Fix #488.

I'm working behind a corporate proxy and since few days, I get errors like in the screenshot below whenever I use flex.

issue

Further debugging revealed that [composer replace the http scheme by tcp in the proxy url](https://github.com/composer/composer/blob/1.8.5/src/Composer/Util/StreamContextFactory.php#L78), which curl doesn't support.

As curl reads proxy conf from env by default, let's just not override it.

### DIFF
--- a/src/CurlDownloader.php
+++ b/src/CurlDownloader.php
@@ -27,7 +27,6 @@ class CurlDownloader
         'http' => [
             'method' => CURLOPT_CUSTOMREQUEST,
             'content' => CURLOPT_POSTFIELDS,
-            'proxy' => CURLOPT_PROXY,
         ],
         'ssl' => [
             'cafile' => CURLOPT_CAINFO,


stof:
What looks weird is that Composer 2.x (still in dev) is still doing the replacement even when using Curl. So either that change is wrong, or Composer 2.x will also suffer from it.

src/CurlDownloader.php
```diff
@@ -27,7 +27,6 @@ class CurlDownloader
         'http' => [
             'method' => CURLOPT_CUSTOMREQUEST,
             'content' => CURLOPT_POSTFIELDS,
-            'proxy' => CURLOPT_PROXY,
```
nicolas-grekas: does it work if we only remove this line and don't add the other ones?
AFAIK, curl already reads from the env by default so there might be no need to do it in PHP.

fherbin: > does it work if we only remove this line and don't add the other ones?
> AFAIK, curl already reads from the env by default so there might be no need to do it in PHP.

yes that works too for me

nicolas-grekas:
@stof Composer 2 borrowed code from flex, so it might benefit from the exposure we have here and thus should be patched accordingly (#486 also)

src/CurlDownloader.php
nicolas-grekas: great, let's remove all added lines then :)

Lctrs: Just tested. It works. Commit is coming (very soon).

Lctrs: @nicolas-grekas Done :)

fabpot:
Thank you @Lctrs.

PR #493, opened 2019-05-01
Improve regex for container configurator

maxhelias:
Escapes the regex of the container configurator because for the case of the variables "env()" this one are broken.
Example with the mercure or doctrine/mongodb-odm-bundle recipe if you remove the dependence.


maxhelias:
@nicolas-grekas It's done 😃

fabpot:
Thank you @maxhelias.

PR #494, opened 2019-05-07
Display stdout too when dump-env fails

nicolas-grekas:
Sometimes the useful info is on stdout.

### DIFF
--- a/src/Command/DumpEnvCommand.php
+++ b/src/Command/DumpEnvCommand.php
@@ -127,7 +127,7 @@ EOPHP;
         $process->run();
 
         if (!$process->isSuccessful()) {
-            throw new \RuntimeException($process->getErrorOutput());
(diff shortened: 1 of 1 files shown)


fabpot:
Thank you @nicolas-grekas.

src/Command/DumpEnvCommand.php
```diff
@@ -127,7 +127,7 @@ private function loadEnv(string $path, string $env): array
         $process->run();
 
         if (!$process->isSuccessful()) {
-            throw new \RuntimeException($process->getErrorOutput());
+            throw new \RuntimeException($process->getErrorOutput().$process->getOutput());
```
lyrixx: 1. there is a ProcessException that does that, but in a better way
2. but there is the `mustRun()` method that is even simpler

https://github.com/symfony/flex/pull/496

PR #499, opened 2019-05-13
Stop relying on symfony/symfony replace section

fabpot:


src/Flex.php
```diff
@@ -108,17 +108,20 @@ public function activate(Composer $composer, IOInterface $io)
...
+        $setRepositories = \Closure::bind(function (RepositoryManager $manager) use (&$symfonyRequire, $downloader) {
             $manager->repositoryClasses = $this->repositoryClasses;
             $manager->setRepositoryClass('composer', TruncatedComposerRepository::class);
             $manager->repositories = $this->repositories;
             $i = 0;
             foreach (RepositoryFactory::defaultRepos(null, $this->config, $manager) as $repo) {
                 $manager->repositories[$i++] = $repo;
                 if ($repo instanceof TruncatedComposerRepository && $symfonyRequire) {
-                    $repo->setSymfonyRequire($symfonyRequire, $this->io);
+                    $versions = $downloader->get('/versions.json')->getBody();
```
nicolas-grekas: we could put this outside of the loop

src/Cache.php
```diff
@@ -51,19 +53,16 @@ public function removeLegacyTags(array $data): array
             return $data;
         }
 
-        $symfonyPackages = [];
+        $masterVersion = $this->versions['master'];
         $symfonySymfony = $data['packages']['symfony/symfony'];
-
```
nicolas-grekas: could be kept :)

src/Flex.php
fabpot: Done

src/Cache.php
fabpot: Done

nicolas-grekas:
Looks like we missed something, this break the Symfony CI :)
https://travis-ci.org/symfony/symfony/builds/540305191

nicolas-grekas:
Maybe because 5.0 is not listed here?
https://flex.symfony.com/versions.json

fabpot:
I've just updated `versions.json`.

fabpot:
@nicolas-grekas I'm going to have a look at this issue this morning.

PR #505, opened 2019-06-03
Further decoupling from symfony/symfony replace section

nicolas-grekas:
Continues #499

### DIFF
--- a/src/Cache.php
+++ b/src/Cache.php
@@ -25,15 +25,32 @@ class Cache extends BaseCache
     private $versionParser;
     private $symfonyRequire;
     private $symfonyConstraints;
-    private $io;
 
     public function setSymfonyRequire(string $symfonyRequire, array $versions, IOInterface $io = null)
     {
         $this->versionParser = new VersionParser();
         $this->symfonyRequire = $symfonyRequire;
         $this->symfonyConstraints = $this->versionParser->parseConstraints($symfonyRequire);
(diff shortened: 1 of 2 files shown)


src/Cache.php
```diff
@@ -25,15 +25,32 @@ class Cache extends BaseCache
...
+                    if (null !== $io) {
+                        $io->writeError(sprintf('<info>Restricting packages listed in "symfony/symfony" to "%s"</info>', $this->symfonyRequire));
+                        $io = null;
+                    }
+                    unset($vers[$i]);
+                }
+            }
+
+            if (!$vers || $vers === $versions['splits'][$name]) {
+                unset($versions['splits'][$name]);
```
nicolas-grekas: we keep only the packages that are affected by the version requirements and have a non-empty intersection with existing versions of it

tests/CacheTest.php
```diff
@@ -29,7 +29,7 @@ public function testRemoveLegacyTags(array $expected, array $packages, string $s
 
     public function provideRemoveLegacyTags()
     {
-        yield 'no-symfony/symfony' => [[123], [123], '~1', ['master' => '3.5', 'splits' => []]];
```
nicolas-grekas: we don't read the "master" entry anymore

fabpot:
Thank you @nicolas-grekas.

PR #507, opened 2019-06-05
Fix referencing real $ENV vars when running dump-env

### LINKED ISSUE #500: Referencing "real" env var with dump-env
Hello,

Suppose I have in my etc/environment DEFAULT_APP_CDN=https://cdn.somewhere.com

In Flex 1.2.0, I can write in my .env
`APP_CDN=${DEFAULT_APP_CDN}`
Then with 
`composer symfony:dump-env whatever`
I got in .env.local.php
`'APP_CDN' => 'https://cdn.somewhere.com'`
And that's pretty cool.

But starting with Flex 1.2.1, I got 
`'APP_CDN' => ''`

Was it not supposed to work from the start ? 
Is there another solution to do this ?

Thx

nicolas-grekas:
Fixes #500

### DIFF
--- a/src/Command/DumpEnvCommand.php
+++ b/src/Command/DumpEnvCommand.php
@@ -17,7 +17,7 @@ use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
-use Symfony\Component\Process\PhpProcess;
+use Symfony\Component\Dotenv\Dotenv;
 use Symfony\Flex\Options;
 
 class DumpEnvCommand extends BaseCommand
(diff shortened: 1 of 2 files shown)


stof:
shouldn't a reference to another env variable be resolved at runtime rather than at dumping time, so that you don't need to re-dump when changing the other env variable ?
This improved dump command is still not equivalent to running Dotenv (changes to the `.env` file of course require re-dumping, but I'm talking about changes to real env variables).

Btw, what happens if we *add* a real env variable, for which the `.env` file contains a default value ? Would re-dumping be necessary ?

nicolas-grekas:
Yes: dump-env is about freezing the default, nothing more. Then real env var can take over at runtime.

DavG:
Sounds perfect to me @nicolas-grekas , thanks for reconsidering the issue. 
I'm gonna do some manual tests too.

fabpot:
Thank you @nicolas-grekas.

DavG:
It works perfectly. Thanks again !

adioe3:
I just ran into this as well. Given this `.env` file:

```
FOO=beans
BAR="cool $FOO"
```

and then in my env exporting another value for `FOO`:

```
export FOO=cats
```

dumping will use the previous value:

```
$ composer dump-env -- prod

$ cat .env.local.php
<?php

// This file was generated by running "composer dump-env prod"

return array (
  'FOO' => 'beans',
  'BAR' => 'cool beans',
);
```

this is obviously not what we'd expect.

Our original use case was splitting a DSN connection into strings so we have

```
TRANSPORT_EVENT_DSN=amqp://${AMQP_USER:-api}:${AMQP_PASS:-api}@${AMQP_HOST:-rabbitmq}:${AMQP_PORT:-5672}/${AMQP_VHOST:-%2f}
```

after dumping this now has some default value and we can't use it for configuration anymore...

sherifabdlnaby:
I think this should be re-opened. `dump-env` doesn't do correct Env Variable Expansions and it ignores Real Env Vars during expansions as shown [here](https://github.com/symfony/flex/pull/507#issuecomment-1021122359)

PR #509, opened 2019-06-05
Fix dealing with locally patched packages

### LINKED ISSUE #508: Undefined index: version_normalized
Error: 
 * AppVeyor: https://ci.appveyor.com/project/fabpot/symfony/builds/25060723#L73
 * Travis: https://travis-ci.org/symfony/symfony/jobs/541735605#L2605

https://github.com/symfony/flex/blob/bc7390dcfa6d50e431e4d15b647544e59a6e3146/src/Cache.php#L82

or

https://github.com/symfony/flex/blob/bc7390dcfa6d50e431e4d15b647544e59a6e3146/src/Cache.php#L101

nicolas-grekas:
Fixes #508
Happens on the CI when local packages are built for testing PRs.

### DIFF
--- a/src/Cache.php
+++ b/src/Cache.php
@@ -66,7 +66,7 @@ class Cache extends BaseCache
 
     public function removeLegacyTags(array $data): array
     {
-        if (!$this->symfonyConstraints) {
+        if (!$this->symfonyConstraints || !isset($data['packages'])) {
             return $data;
         }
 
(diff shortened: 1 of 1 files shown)


fabpot:
Thank you @nicolas-grekas.

src/Cache.php
```diff
@@ -78,6 +78,8 @@ public function removeLegacyTags(array $data): array
             foreach ($versions as $version => $composerJson) {
                 if ('dev-master' === $version) {
                     $normalizedVersion = $this->versionParser->normalize($devMasterAlias);
+                } elseif (!isset($composerJson['version_normalized'])) {
```
stof: why not attempting to read `$composerJson['version']`  and normalizing it (or just `$version`) ? We have access to the VersionParser to normalize here.

nicolas-grekas: because we don't really care: when this happens, we always want to keep the "version" anyway

PR #514, opened 2019-06-12
Add ext-openssl to the requirements

nicolas-grekas:
Spotted today during the SFLive Warszawa workshop.


nicolas-grekas:
Mmm actually curl can be enough as it can embed its own SSL routines...

PR #519, opened 2019-06-20
Don't preload packages listed in "conflict" nor "replace"

nicolas-grekas:
This is not needed anymore since https://github.com/composer/composer/pull/7454 and Composer 1.7.3 and nobody told us? /cc @stof :)


stof:
I forgot that Flex was preloading these :smile:

PR #533, opened 2019-07-08
Fix docker configurator with many mark in volumes section

maxhelias:
When you delete a recipe from the volumes section when there are only recipes, the "volumes" keywords are deleted.

cc @dunglas

### DIFF
--- a/tests/Configurator/DockerComposeConfiguratorTest.php
+++ b/tests/Configurator/DockerComposeConfiguratorTest.php
@@ -234,8 +234,6 @@ YAML
       - db-data:/var/lib/postgresql/data:rw
       # You may use a bind-mounted host directory instead, so that it is harder to accidentally remove the volume and lose all your data!
       # - ./docker/db/data:/var/lib/postgresql/data:rw
-    ports:
-      - "5432:5432"
 ###< doctrine/doctrine-bundle ###
 
 volumes:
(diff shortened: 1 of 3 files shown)


maxhelias:
friendly ping @nicolas-grekas 😃

tests/Configurator/DockerComposeConfiguratorTest.php
```diff
@@ -234,8 +234,6 @@ public function testConfigureFileWithExistingMarks()
       - db-data:/var/lib/postgresql/data:rw
       # You may use a bind-mounted host directory instead, so that it is harder to accidentally remove the volume and lose all your data!
       # - ./docker/db/data:/var/lib/postgresql/data:rw
-    ports:
-      - "5432:5432"
```
fabpot: Why have you removed the ports in the tests?

maxhelias: We deleted them from the recipes but i can revert that

fabpot: Removing them from the recipes is a mistake IMHO. What we should do is only expose the port but not attach it to a specific port locally. So, I would use:

```
ports:
    - 5432
```

That allows the port to be exposed and a random port to be used locally (which can be picked up by the Symfony CLI Web Server -- https://symfony.com/doc/current/setup/symfony_server.html#docker-integration). That way, you can also have several projects using the same deps and running independently at the same time.

dunglas: Anyway we don’t have to sync these test files with the recipes.

fabpot: Sure thing, just wanted to reiterate my point of view on ports and Docker :)

fabpot:
Thank you @maxhelias.

tests/Configurator/DockerComposeConfiguratorTest.php
maxhelias: You're doing well because I misunderstood initially. I'm going to edit the PR symfony/recipes#413

PR #535, opened 2019-07-10
Fix var_export on ContainerConfigurator

maxhelias:
Fix var_export on ContainerConfigurator

### DIFF
--- a/src/Configurator/ContainerConfigurator.php
+++ b/src/Configurator/ContainerConfigurator.php
@@ -31,10 +31,8 @@ class ContainerConfigurator extends AbstractConfigurator
         $target = $this->options->get('root-dir').'/'.$this->options->expandTargetDir('%CONFIG_DIR%/services.yaml');
         $lines = [];
         foreach (file($target) as $line) {
-            foreach (array_keys($parameters) as $key) {
-                if (preg_match(sprintf('/^\s+%s\:/', preg_quote($key, '/')), $line)) {
-                    continue 2;
-                }
+            if ($this->removeParameters(1, $parameters, $line)) {
+                continue;
             }
             $lines[] = $line;
         }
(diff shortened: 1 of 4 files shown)


src/Configurator/ContainerConfigurator.php
```diff
@@ -76,13 +76,31 @@ private function addParameters(array $parameters)
...
+            $parametersLines[] = sprintf("    %s: '%s'%s", $key, $value, "\n");
         }
         if (!$endAt) {
             $parametersLines[] = "\n";
         }
         array_splice($lines, $endAt, 0, $parametersLines);
         file_put_contents($target, implode('', $lines));
     }
+
+    private function parse($level, $array): string
```
nicolas-grekas: `dumpYaml()`? could deserve some test cases?

src/Downloader.php
```diff
@@ -110,8 +110,7 @@ public function getRecipes(array $operations): array
                 }
             }
 
-            // FIXME: getNames() can return n names
-            $name = str_replace('/', ',', $package->getNames()[0]);
+            $name = str_replace('/', ',', $package->getName());
```
nicolas-grekas: This look suspicious to me: there must be a reason why we didn't write this code originally.
What should we do with the extra names? When does it happen in the first place?
This should be investigated to me.

maxhelias: I agree, but I try to find out why and I still do not have the answer. `getNames()` always return `getName()` first :
https://github.com/composer/composer/blob/master/src/Composer/Package/BasePackage.php#L95

src/Configurator/ContainerConfigurator.php
```diff
@@ -76,13 +76,31 @@ private function addParameters(array $parameters)
             $parametersLines[] = "parameters:\n";
         }
         foreach ($parameters as $key => $value) {
-            // FIXME: var_export() only works for basics types, but we don't have access to the Symfony YAML component here
-            $parametersLines[] = sprintf('    %s: %s%s', $key, var_export($value, true), "\n");
+            if (\is_array($value)) {
+                $parametersLines[] = $this->parse(1, $value);
+                continue;
+            }
+            $parametersLines[] = sprintf("    %s: '%s'%s", $key, $value, "\n");
```
fabpot: What if `$value` contains `'`?

src/Downloader.php
fabpot: Works for me. The FIXME was more about what to do when we have more names. So, I would keep the FIXME and change the currently implementation as you did with `getName()`.

src/Configurator/ContainerConfigurator.php
maxhelias: I add `addslashes` for escape `$value`

src/Configurator/ContainerConfigurator.php
maxhelias: There was still some bug that are now corrected and covered by tests 😃

src/Configurator/ContainerConfigurator.php
```diff
@@ -76,13 +74,45 @@ private function addParameters(array $parameters)
             $parametersLines[] = "parameters:\n";
         }
         foreach ($parameters as $key => $value) {
-            // FIXME: var_export() only works for basics types, but we don't have access to the Symfony YAML component here
-            $parametersLines[] = sprintf('    %s: %s%s', $key, var_export($value, true), "\n");
+            if (\is_array($value)) {
+                $parametersLines[] = sprintf("    %s:\n%s", $key, $this->dumpYaml(2, $value));
+                continue;
+            }
+            $parametersLines[] = sprintf("    %s: '%s'%s", $key, addslashes($value), "\n");
```
fabpot: What about using str_replace instead as we only need to escape quotes?

maxhelias: Yes, I think it will be better because i juste thinking it's not a slash for escape in yaml

fabpot:
Thank you @maxhelias.

Issue #527, opened 2019-06-29
Reading flex.symfony.com on composer install when composer.lock and symfony.lock files are present


bbaga:
We are facing a flex related problem, it looks like flex is trying to fetch the `versions.json` during `composer install`, even though the lock files are present. Our workflow and setup is nothing special, composer packages mirrored with artifactory, composer.lock committed to the repo and we run `composer install` before creating the deploy artifacts.

We had no problem with this workflow until `1.2.6` was released, then our build pipelines started to fail, because `flex.symfony.com` is unreachable from our build machines (they can access internal resources only).

I was wondering, what is the purpose of that http call. My understanding is that, at that point all necessary package information should be available from the lock files.

`symonfy/flex`: `>1.2.5`

bbaga:
Additionally, we tried to run `composer install --no-plugins`, but then our `post-install-cmd` scripts are failing, supposedly due to composer's scripts plugin getting disabled as well.

This issue forced us to lock flex's version on `1.2.5` and to start investigating mirroring solutions.

bitgandtter:
we are facing the same issue

bbaga:
@bitgandtter this is the only thing which looked like a reasonable project is this: https://github.com/moay/server-for-symfony-flex

We haven't tried it yet, still hoping this is going to be resolved within `flex`.
At the moment we are locked on `1.2.5`.

bitgandtter:
wow a week now without response, I'm not able to use the latest version of flex because of this issue

nicolas-grekas:
@bitgandtter would you mind having a look how to prevent doing this call? I agree it should be preventable and prevented.

bitgandtter:
hello again, so weird, after I give it a try today and after an also weird error message 

`"https://flex.symfony.com/versions.json" does not contain valid JSON
Parse error on line 1:

^
Expected one of: 'STRING', 'NUMBER', 'NULL', 'TRUE', 'FALSE', '{', '['
https://flex.symfony.com/versions.json could not be fully loaded, package information was loaded from the local cache and may be out of date
`

it works again with the latest flex version. I also review the latest commits and I don't see any related to it.

o-alquimista:
> hello again, so weird, after I give it a try today and after an also weird error message
> 
> `"https://flex.symfony.com/versions.json" does not contain valid JSON
> Parse error on line 1:
> 
> ^
> Expected one of: 'STRING', 'NUMBER', 'NULL', 'TRUE', 'FALSE', '{', '['
[...]

PR #537, opened 2019-07-11
Fix undefined index 'splits'

yceruto:
See https://travis-ci.org/symfony/symfony/jobs/557427277#L2601

### DIFF
--- a/src/Cache.php
+++ b/src/Cache.php
@@ -118,6 +118,10 @@ class Cache extends BaseCache
         $versions = $this->downloader->getVersions();
         $this->downloader = null;
 
+        if (!isset($versions['splits'])) {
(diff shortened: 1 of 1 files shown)


nicolas-grekas:
When does this happen?

yceruto:
Maybe because https://github.com/symfony/flex/issues/527#issuecomment-510220086 ?

yceruto:
Although it shouldn't happen, it's always present at https://flex.symfony.com/versions.json, isn't it?

Maybe a download error from Travis because AppVeyor don't have this issue.

nicolas-grekas:
Fixed in https://github.com/symfony/flex/pull/538

PR #542, opened 2019-07-15
Fix running other plugins on create-project

### LINKED ISSUE #529: Execution of "post-create-project-cmd" is prevented by symfony/flex
If you create a project with Symfony Flex and you have other Composer Plugins in your composer.json or globally installed which are listening on "post-create-project-cmd", those events do not get executed anymore when creating a project.

This is only happening if composer has randomly installed the Flex Plugin before the other plugins as the other plugins are then installed by Flex itself.

This causes those Plugins to register there Event Listeners to the new Event Dispacher created for the composer update command executed by Flex. But with this Event Dispatcher an "post-create-project-cmd" is never emitted. This is only emitted on the "root" composer Event Dispatcher. So i guess the bug is that Flex should also emit the "post-create-project-cmd" after it emits the "post-root-package-install" manually.

nicolas-grekas:
Instead of creating a separate set of "composer" and "installer" objects, this PR recycles them so that plugins run in the same scope.

Fix #529

### DIFF
--- a/src/Flex.php
+++ b/src/Flex.php
@@ -361,44 +361,17 @@ class Flex implements PluginInterface, EventSubscriberInterface
 
         file_put_contents($jsonPath, $manipulator->getContents());
 
-        $composer = Factory::create($this->io);
-        $composer->getDownloadManager()->setOutputProgress($this->progress);
-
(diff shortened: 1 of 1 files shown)


fabpot:
Thank you @nicolas-grekas.

PR #543, opened 2019-07-15
Fix unpacking when private packagist is used

### LINKED ISSUE #502: Unable to unpack packages
Using composer 1.8.5 and PHP 7.3 (we use private packagist)

If I install symfony like

~` composer create-project symfony/skeleton my-project`~
`composer create-project symfony/website-skeleton my-project`

And then I run `composer unpack orm`, it just says "Nothing to unpack".
Same for other packs.

Unpack doesn't work either for other projects. Thank you.

Am I missing something?

nicolas-grekas:
Fix #502


fabpot:
Thank you @nicolas-grekas.

PR #547, opened 2019-07-19
Fix "Cannot declare class because the name is already in use" errors

nicolas-grekas:
`require_once` doesn't work when flex is loaded from two separate directories (global + local)


fabpot:
Thank you @nicolas-grekas.

PR #558, opened 2019-09-13
Handle package versions with slashes

nenad:
This PR fixes a flaw in the parsing of packages that require a branch alias version which contains a slash, by treating each argument in the resolver as package with version. This matches the documentation from [`compose require [packages]`](https://github.com/composer/composer/blob/master/src/Composer/Command/RequireCommand.php/#L48)

This means that if you require packages split by a space through the CLI, you'll need to quote them. Instead of `composer require guzzlehttp/guzzle ^6.0`, you'll need to use `composer require "guzzlehttp/guzzle ^6.0"` so that it's treated as one argument, and matches the composer documentation.

Solves #559

### DIFF
--- a/src/PackageResolver.php
+++ b/src/PackageResolver.php
@@ -37,41 +37,47 @@ class PackageResolver
         }
 
         // first pass split on ':', '=' or ' ' to separate package names and versions
-        $explodedArguments = [];
+        $searchInfo = [];
         foreach ($arguments as $argument) {
             if (
                 (false !== $pos = strpos($argument, ':')) ||
                 (false !== $pos = strpos($argument, '=')) ||
                 (false !== $pos = strpos($argument, ' '))
             ) {
-                $explodedArguments[] = substr($argument, 0, $pos);
-                $explodedArguments[] = substr($argument, $pos + 1);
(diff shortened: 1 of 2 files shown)


fabpot:
Can. you try to fix it?

nenad:
@fabpot Sorry for the force-push noise. Just wrapped up the PR, please take a look when you get the chance.

src/PackageResolver.php
```diff
@@ -32,42 +32,52 @@ public function __construct(Downloader $downloader)
...
-                    } catch (\UnexpectedValueException $e) {
-                        // is it a special Symfony version?
-                        if (!\in_array($argument, self::$SYMFONY_VERSIONS, true)) {
-                            $this->throwAlternatives($argument, $i);
-                        }
+        $i = 0;
+        foreach ($searchInfo as $name => $version) {
+            $version = self::$aliases[$version] ?? $version;
+            $name = self::$aliases[$name] ?? $name;
+            $name .= ($version ? ":$version" : '');
```
nicolas-grekas: extra brackets should be removed

nicolas-grekas:
Closing due to lack of activity.

Issue #548, opened 2019-08-08
Package is not found when trying to install locale zip package


gmponos:
According to https://github.com/composer/composer/issues/8265

My `composer.json`:

```json
    "repositories": [
        ...
        {
            "type": "package",
            "package": {
                "name": "private-wordpress-plugin/sitepress-multilingual-cms",
                "version": "4.2.1",
                "dist": {
                    "type": "zip",
                    "url": "../www/sitepress-multilingual-cms.4.2.1.zip"
                },
                "type": "wordpress-plugin",
                "require": {
                    "composer/installers": "^1.6.0"
                }
            }
        }
```

Output of `composer diagnose`:

```
Checking composer.json: OK
Checking platform settings: OK
Checking git settings: OK
Checking http connectivity to packagist: OK
Checking https connectivity to packagist: OK
Checking github.com rate limit: OK
Checking disk free space: OK
Checking pubkeys: 
Tags Public Key Fingerprint: 57815BA2 7E54DC31 7ECC7CC5 573090D0  87719BA6 8F3BB723 4E5D42D0 84A14642
Dev Public Key Fingerprint: 4AC45767 E5EC2265 2F0C1167 CBBB8A2B  0C708369 153E328C AD90147D AFE50952
OK
Checking composer version: You are not running the latest stable version, run `composer self-update` to update (1.8.4 => 1.9.0)
Composer version: 1.8.4
PHP version: 7.2.19
PHP binary path: /usr/bin/php
```

**Note** I checked also with latest version.

When I run this command:

```
composer install
```

I get the following output:

```
private-wordpress-plugin/sitepress-multilingual-cms (4.2.1): Reading ../www/sitepress-multilingual-cms.4.2.1.zip
Downloading (100%)         
    Download failed, retrying...
Reading ../www/sitepress-multilingual-cms.4.2.1.zip
Downloading (100%)         
    Download failed, retrying...
Reading ../www/sitepress-multilingual-cms.4.2.1.zip
Downloading (100%)         
Failed: [Composer\Downloader\TransportException] 0: "../www/sitepress-multilingual-cms.4.2.1.zip" appears broken, and returned an empty 200 response

                                                                                                    
  [Composer\Downloader\TransportException]                                                          
  "../www/sitepress-multilingual-cms.4.2.1.zip" appears broken, and returned an empty 200 response  
                                                                                                    

Exception trace:
[...]

PR #561, opened 2019-09-15
Skip parallel downloader for local files

### LINKED ISSUE #560: Flex breaks composers artifact repository
Hi!

When using Flex I cannot install packages via composers built-in [artifact repository](https://getcomposer.org/doc/05-repositories.md#artifact).

**Expected behavior:**
Flex should not break composer built-in functionality.

**Actual behavior:**
My composer.json includes
```
    "repositories": [
        {
            "type": "artifact",
            "url": "var/installer/"
        }
    ]
```
and when having `symfony/flex:1.4.5` installed I run into this problem:
```
$ composer require foo/my-package 1.0
Restricting packages listed in "symfony/symfony" to "4.3.*"
Package operations: 1 install, 0 updates, 0 removal
  - Installing foo/my-package (1.0): Downloading (100%)
Downloading (100%)
Downloading (100%)

  [Composer\Downloader\TransportException]
  "var/installer/my-package-1.0.zip" appears broken, and returned an empty 200 response
```
When removing Symfony Flex the installation from my local zip works as expected.

**What I tried:**

Seeing `Downloading (100%)` 3 times lead me to `Composer\Downloader\FileDownloader` and its retries loop. So I debugged further (with and without Flex) and finally found the main difference being 
- `Composer\Util\RemoteFilesystem::getRemoteContents()` using `file_get_contents()` 
- `Symfony\Flex\ParallelDownloader::getRemoteContents()` using the `Symfony\Flex\CurlDownloader`

The latter fails when trying to use curl to fetch the zip from my local filesystem. 

@nicolas-grekas any idea how to work around that problem without removing flex? Is Flex supposed to work in such a setup?

I tried [replacing this line](https://github.com/symfony/flex/blob/master/src/ParallelDownloader.php#L247) with `if (stripos($originUrl, '://')  === false || !$this->downloader) {` and everything worked. But not knowing enough about composer internals I am not sure if this is the right place to start working on a bugfix.

kevinpapst:
Fixes #560 

Currently Flex breaks installations of local packages via the composer built-in repository types:
- artifact repository (see #560)
- package repository (see #548)

This PR leaves the download to the original composer downloader, in case the URL of the package does not start with `http(s):`.

I found a very similar code piece in https://github.com/symfony/flex/blob/master/src/Flex.php#L605 (just for reference for maintainer).

### DIFF
--- a/src/ParallelDownloader.php
+++ b/src/ParallelDownloader.php
@@ -244,7 +244,7 @@ class ParallelDownloader extends RemoteFilesystem
             return $result;
         }
 
-        if (!$this->downloader) {
+        if (!$this->downloader || !preg_match('/^https?:/', $originUrl)) {
             return parent::getRemoteContents($originUrl, $fileUrl, $context, $responseHeaders);
         }
 


src/ParallelDownloader.php
```diff
@@ -244,7 +244,7 @@ protected function getRemoteContents($originUrl, $fileUrl, $context, array &$res
             return $result;
         }
 
-        if (!$this->downloader) {
+        if (!preg_match('/^https?:/', $originUrl) || !$this->downloader) {
```
stof: I would swap the conditions. Checking for a `null` downloader is cheaper than running the regex.

kevinpapst: True, changed.

kevinpapst:
ping @nicolas-grekas @stof - anything else that I can/should do to progress?

fabpot:
Thank you @kevinpapst.

Issue #518, opened 2019-06-19
[RFC][DX] Set of commands & changes to add transparency to Flex


weaverryan:
Hi friends!

I ❤️the recipe system. The idea of having packages auto-configure and scaffold themselves is the *best*. However, after a few recent conversations, I think Flex & the recipe system is a bit "murky" to people:

* they're not sure where the recipes are coming from or how to find them
* they don't know if/when they should update recipes or even how they would find which file came from which recipe to know how to find that file
* related to above, they don't know about the `symfony:sync-recipes` command. Rather, the installation of recipes feels like a "magic" thing that happens once, and can never be repeated afterwards.

This is a big RFC to make numerous, minor changes to help make this experience more transparent and user-friendly:

**1) ~symfony:recipes~** DONE! ✅ 

This command would simply list all of the installed recipes, probably only listing those that actually *had* a recipe (not a full list of what's in symfony.lock). Ideally, it would also show you if updates are available:

```
Installed recipes:

 * symfony/framework-bundle 4.2
 * symfony/console 3.4 (update available)
 * ...

Run:
 * composer symfony:recipe:show vendor/package to see details about a recipe.
 * composer symfony:recipe:update vendor/package to update that recipe.
 * composer symfony:recipe:blame to see a tree of all of the installed/updated files
```

**2) ~symfony:recipes:install~** DONE! ✅ 

This would simply be a rename of `symfony:sync-recipes`. It's "installing" recipes, so i think it makes more sense.

**3) symfony:recipe:update**

This would effectively be an alias for the current `symfony:sync-recipes --force`. The name "update" is more revealing to what it does. However, because it's not a real "merge/update", we'll of course still have output that describes what happened and how to proceed, etc.

Additionally, it will now accept the package name(s) that should be updated:

```
composer symfony:recipe:update symfony/framework-bundle
```

When using this option, I think we should also print a list of the PR's that updated the recipes, for each updated library:

```
Recipe updates for symfony/framework-bundle:
* ensure $_ENV is contains real env vars from $_SERVER , or defaults from .env.local.php
    https://github.com/symfony/recipes/pull/611
* Don't use putenv by default
    https://github.com/symfony/recipes/pull/572
```

**4) ~symfony:recipes <package/name>~** DONE! ✅ 
Usage: `composer symfony:recipes symfony/framework-bundle`

[...]

PR #562, opened 2019-09-18
[DX] Command to list all available recipes

maxhelias:
First step of #518, the command `symfony:recipes:show` and alias `symfony:recipes`

Result :
Capture d’écran 2019-09-18 à 15 46 09

I need feedback on the disposition and maybe i can add the version of the recipe installed like
 - doctrine/annontations 1.0

Tell me what you think :) 

Note: the last section is currently commented out, and will be added later

### DIFF
--- a/src/Command/RecipesCommand.php
+++ b/src/Command/RecipesCommand.php
@@ -1,0 +1,255 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Flex\Command;
+
+use Composer\Command\BaseCommand;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Flex\InformationOperation;
+use Symfony\Flex\Lock;
+use Symfony\Flex\Recipe;
+
+/**
+ * @author Maxime Hélias <maximehelias16@gmail.com>
+ */
+class RecipesCommand extends BaseCommand
+{
+    /** @var \Symfony\Flex\Flex */
+    private $flex;
+
+    /** @var Lock */
+    private $symfonyLock;
+
+    public function __construct(/* cannot be type-hinted */ $flex, Lock $symfonyLock)
+    {
+        $this->flex = $flex;
+        $this->symfonyLock = $symfonyLock;
+
+        parent::__construct();
+    }
+
+    protected function configure()
+    {
+        $this->setName('symfony:recipes')
+            ->setAliases(['recipes'])
+            ->setDescription('Shows information about all available recipes.')
+            ->setDefinition([
+                new InputArgument('package', InputArgument::OPTIONAL, 'Package to inspect, if not provided all packages are.'),
+            ])
+        ;
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        $installedRepo = $this->getComposer()->getRepositoryManager()->getLocalRepository();
+
+        // Inspect one or all packages
+        $package = $input->getArgument('package');
+        if (null !== $package) {
+            $packages = [0 => ['name' => strtolower($package)]];
+        } else {
+            $locker = $this->getComposer()->getLocker();
+            $lockData = $locker->getLockData();
+
+            // Merge all packages installed
+            $packages = array_merge($lockData['packages'], $lockData['packages-dev']);
+        }
+
+        $operations = [];
+        foreach ($packages as $value) {
+            if (null === $pkg = $installedRepo->findPackage($value['name'], '*')) {
+                $this->getIO()->writeError(sprintf('<error>Package %s is not installed</error>', $value['name']));
+
+                continue;
+            }
+
+            $operations[] = new InformationOperation($pkg);
+        }
+
+        $recipes = $this->flex->fetchRecipes($operations);
+        ksort($recipes);
+
+        $nbRecipe = \count($recipes);
+        if ($nbRecipe <= 0) {
+            $this->getIO()->writeError('<error>No recipe found</error>');
+
+            return 1;
+        }
+
+        // Display the information about a specific recipe
+        if (1 === $nbRecipe) {
+            $this->displayPackageInformation(current($recipes));
+
+            return 0;
+        }
+
+        // display a resume of all packages
+        $write = [
+            '',
+            '<bg=blue;fg=white>                      </>',
+            '<bg=blue;fg=white> Available recipes.   </>',
+            '<bg=blue;fg=white>                      </>',
+            '',
+        ];
+
+        /** @var Recipe $recipe */
+        foreach ($recipes as $name => $recipe) {
+            $lockRef = $this->symfonyLock->get($name)['recipe']['ref'] ?? null;
+
+            $additional = '';
+            if (null === $lockRef && null !== $recipe->getRef()) {
+                $additional = '<comment>(recipe not installed)</comment>';
+            } elseif ($recipe->getRef() !== $lockRef) {
+                $additional = '<comment>(update available)</comment>';
+            }
+            $write[] = sprintf(' * %s %s', $name, $additional);
+        }
+
+        $write[] = '';
+        $write[] = 'Run:';
+        $write[] = ' * <info>composer recipes vendor/package</info> to see details about a recipe.';
+        $write[] = ' * <info>composer recipes:install vendor/package --force -v</info> to update that recipe.';
+        $write[] = '';
+
+        $this->getIO()->write($write);
+
+        return 0;
+    }
+
+    private function displayPackageInformation(Recipe $recipe)
+    {
+        $recipeLock = $this->symfonyLock->get($recipe->getName());
+
+        $lockRef = $recipeLock['recipe']['ref'] ?? null;
+        $lockFiles = $recipeLock['files'] ?? null;
+
+        $status = '<comment>up to date</comment>';
+        if ($recipe->isAuto()) {
+            $status = '<comment>auto-generated recipe</comment>';
+        } elseif (null === $lockRef && null !== $recipe->getRef()) {
+            $status = '<comment>recipe not installed</comment>';
+        } elseif ($recipe->getRef() !== $lockRef) {
+            $status = '<comment>update available</comment>';
+        }
+
+        $io = $this->getIO();
+        $io->write('<info>name</info>     : '.$recipe->getName());
+        $io->write('<info>version</info>  : '.$recipeLock['version']);
+        if (!$recipe->isAuto()) {
+            $io->write('<info>repo</info>     : '.sprintf('https://%s/tree/master/%s/%s', $recipeLock['recipe']['repo'], $recipe->getName(), $recipeLock['version']));
+        }
+        $io->write('<info>status</info>   : '.$status);
+
+        if (null !== $lockFiles) {
+            $io->write('<info>files</info>    : ');
+            $io->write('');
+
+            $tree = $this->generateFilesTree($lockFiles);
+
+            $this->displayFilesTree($tree);
+        }
+
+        $io->write([
+            '',
+            'Update this recipe by running:',
+            sprintf('<info>composer recipes:install %s --force -v</info>', $recipe->getName()),
+        ]);
+    }
+
+    private function generateFilesTree(array $files): array
+    {
+        $tree = [];
+        foreach ($files as $file) {
+            $path = explode('/', $file);
+
+            $tree = array_merge_recursive($tree, $this->addNode($path));
+        }
+
+        return $tree;
+    }
+
+    private function addNode(array $node): array
+    {
+        $current = array_shift($node);
+
+        $subTree = [];
+        if (null !== $current) {
+            $subTree[$current] = $this->addNode($node);
+        }
+
+        return $subTree;
+    }
+
+    /**
+     * Note : We do not display file modification information with Configurator like ComposerScripts, Container, DockerComposer, Dockerfile, Env, Gitignore and Makefile.
+     */
+    private function displayFilesTree(array $tree)
+    {
+        $endKey = array_key_last($tree);
+        foreach ($tree as $dir => $files) {
+            $treeBar = '├';
+            $total = \count($files);
+            if (0 === $total || $endKey === $dir) {
+                $treeBar = '└';
+            }
+
+            $info = sprintf(
+                '%s──%s',
+                $treeBar,
+                $dir
+            );
+            $this->writeTreeLine($info);
+
+            $treeBar = str_replace('└', ' ', $treeBar);
+
+            $this->displayTree($files, $treeBar);
+        }
+    }
+
+    private function displayTree(array $tree, $previousTreeBar = '├', $level = 1)
+    {
+        $previousTreeBar = str_replace('├', '│', $previousTreeBar);
+        $treeBar = $previousTreeBar.'  ├';
+
+        $i = 0;
+        $total = \count($tree);
+
+        foreach ($tree as $dir => $files) {
+            ++$i;
+            if ($i === $total) {
+                $treeBar = $previousTreeBar.'  └';
+            }
+
+            $info = sprintf(
+                '%s──%s',
+                $treeBar,
+                $dir
+            );
+            $this->writeTreeLine($info);
+
+            $treeBar = str_replace('└', ' ', $treeBar);
+
+            $this->displayTree($files, $treeBar, $level + 1);
+        }
+    }
+
+    private function writeTreeLine($line)
+    {
+        $io = $this->getIO();
+        if (!$io->isDecorated()) {
+            $line = str_replace(['└', '├', '──', '│'], ['`-', '|-', '-', '|'], $line);
+        }
+
+        $io->write($line);
+    }
+}
--- a/src/Flex.php
+++ b/src/Flex.php
@@ -243,6 +243,7 @@ class Flex implements PluginInterface, EventSubscriberInterface
             $app->add(new Command\UpdateCommand($resolver));
             $app->add(new Command\RemoveCommand($resolver));
             $app->add(new Command\UnpackCommand($resolver));
+            $app->add(new Command\RecipesCommand($this, $this->lock));
             $app->add(new Command\InstallRecipesCommand($this, $this->options->get('root-dir')));
             $app->add(new Command\GenerateIdCommand($this));
             $app->add(new Command\DumpEnvCommand($this->config, $this->options));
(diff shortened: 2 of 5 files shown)


src/Command/RecipesCommand.php
```diff
@@ -0,0 +1,102 @@
...
+    {
+        $this->setName('symfony:recipes:show')
+            ->setAliases(['symfony:recipes'])
+            ->setDescription('Shows information about all available recipes.')
+        ;
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output)
+    {
+        $symfonyLock = new Lock(getenv('SYMFONY_LOCKFILE') ?: str_replace('composer.json', 'symfony.lock', Factory::getComposerFile()));
```
weaverryan: could move this down, closer to where it's needed.

src/Command/RecipesCommand.php
```diff
@@ -0,0 +1,102 @@
...
+        $packages = array_merge($lockData['packages'], $lockData['packages-dev']);
+
+        $installedRepo = $this->getComposer()->getRepositoryManager()->getLocalRepository();
+
+        $operations = [];
+        foreach ($packages as $key => $value) {
+            if (null === $pkg = $installedRepo->findPackage($value['name'], '*')) {
+                $this->getIO()->writeError(sprintf('<error>Package %s is not installed</error>', $value['name']));
+
+                return 1;
```
weaverryan: could we continue here?

src/Command/RecipesCommand.php
```diff
@@ -0,0 +1,102 @@
...
+            $operations[] = new InformationOperation($pkg);
+        }
+
+        $recipes = $this->flex->fetchRecipes($operations);
+        ksort($recipes);
+
+        $write = [
+            '',
+            '<bg=blue;fg=white>                      </>',
+            '<bg=blue;fg=white> Recipes available.   </>',
```
weaverryan: `Installed recipes:`?

src/Flex.php
```diff
@@ -653,23 +654,23 @@ public function generateFlexId()
         $this->updateComposerLock();
     }
 
-    private function fetchRecipes(): array
+    public function fetchRecipes(array &$operations): array
```
weaverryan: Is the by-reference necessary? It looks like it's just to clear `$this->operations` - could we just do that above after calling this method?

src/InformationOperation.php
```diff
@@ -0,0 +1,52 @@
...
+class InformationOperation extends SolverOperation
+{
+    protected $package;
+
+    /**
+     * Initializes operation.
+     *
+     * @param PackageInterface $package package instance
+     * @param string           $reason  operation reason
+     */
```
weaverryan: phpdoc is probably unnecessary

src/InformationOperation.php
```diff
@@ -0,0 +1,52 @@
+<?php
+
+namespace Symfony\Flex;
+
+use Composer\DependencyResolver\Operation\SolverOperation;
+use Composer\Package\PackageInterface;
+
+class InformationOperation extends SolverOperation
+{
+    protected $package;
```
weaverryan: Does it need to be protected?

src/Command/RecipesCommand.php
maxhelias: It does not only list the installed recipes but all that available for the project

src/Flex.php
maxhelias: Yes we can but i need to pass the operations in the new command and the reference improve the memory. Or, I can create a new method to define the operations ?

src/InformationOperation.php
maxhelias: What do you mean ? "private" would be better ?

src/Flex.php
stof: >  the reference improve the memory.

Are you sure it does ? Do you have numbers proving it ? With PHP's copy-on-write behavior, such assumption about references is often wrong.

src/InformationOperation.php
stof: yes. The rule in Symfony is that we use `private` visibility, unless we have a use case where we need `protected`. This is one of the basis of our BC policy.

src/Flex.php
maxhelias: I read this in the blackfire's book and i could be wrong about the use case. Can you explain my mistake ?

maxhelias:
friendly ping @weaverryan 😄

src/Flex.php
weaverryan: I'd prefer code clarity over performance first. And then profile things and see where bottlenecks are. So, I would not pass as a reference, and then, after calling `$this->fetchRecipes($this->operations);` above, do `$this->operations = []`.

src/Command/RecipesCommand.php
```diff
@@ -0,0 +1,109 @@
...
+                $additional = '<comment>(update available)</comment>';
+            }
+            $write[] = sprintf(' * %s %s', $name, $additional);
+        }
+
+        // TODO : Uncomment the lines following the implemented features
+        //$write[] = '';
+        //$write[] = '<fg=blue>Run</>:';
+        //$write[] = ' * composer symfony:recipes vendor/package to see details about a recipe.';
+        //$write[] = ' * composer symfony:recipes:update vendor/package to update that recipe.';
```
weaverryan: I think to make this new command "complete", we would need to at least be able to show 2 commands down here:

A) A command to "update" a single recipe - so `symfony:recipe:update` - item (3) from #518 
B) A command to "install" missing recipes - so `symfony:recipe:install` - item (2) from #518 

The `symfony:recipes vendor/package` or whatever we call it - item (4) from #518 would be super nice to have, as would `symfony:recipes:blame` - but those could come later.

maxhelias: - I started the `symfony:recipes vendor/package` in this command. I will update this PR when it's ready
- The `symfony:recipes:install` is ready here : #565 
- For the `symfony:recipes:update`, it would be better to have some feedback on the #565 before I start it
- And for the `symfony:recipes:blame`, it will be a lot more complicated and I agree to do it later

maxhelias:
I updated the PR, there is now the possibility to inspect only one recipe 
Here is an overview, make suggestions for the final rendering 😉 

Capture d’écran 2019-11-13 à 15 30 37

src/Command/RecipesCommand.php
```diff
@@ -0,0 +1,256 @@
...
+            $status = '<comment>new recipe available</comment>';
+        } elseif ($recipe->getRef() !== $lockRef) {
+            $status = '<comment>out of date</comment>';
+        }
+
+        $io = $this->getIO();
+        $io->write('<info>name</info>     : '.$recipe->getName());
+        $io->write('<info>version</info>  : '.$recipeLock['version']);
+        $io->write('<info>repo</info>     : '.$recipeLock['recipe']['repo']);
+        $io->write('<info>sha</info>      : '.$lockRef);
```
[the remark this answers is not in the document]
fabpot: This perfect world did exist. Heck, it was part of the very first release of the server, where you were able to have an id for your project. This kind of information was stored on the server exactly to be able further down the line to implement upgrades. But "people" strongly disagreed back then. They did not want any form of identification. So, the Flex server was downgraded to a dumb server.

src/Command/RecipesCommand.php
```diff
@@ -0,0 +1,256 @@
...
+
+        return 0;
+    }
+
+    private function displayPackageInformation(Recipe $recipe)
+    {
+        $recipeLock = $this->symfonyLock->get($recipe->getName());
+
+        $lockRef = $recipeLock['recipe']['ref'] ?? null;
+        $lockFiles = $recipeLock['files'] ?? null;
```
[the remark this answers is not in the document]
maxhelias: Indeed, and I have no idea for the moment how to retrieve the information

src/Command/RecipesCommand.php
```diff
@@ -0,0 +1,256 @@
...
+        } elseif ($recipe->getRef() !== $lockRef) {
+            $status = '<comment>out of date</comment>';
+        }
+
+        $io = $this->getIO();
+        $io->write('<info>name</info>     : '.$recipe->getName());
+        $io->write('<info>version</info>  : '.$recipeLock['version']);
+        $io->write('<info>repo</info>     : '.$recipeLock['recipe']['repo']);
+        $io->write('<info>sha</info>      : '.$lockRef);
+        $io->write('<info>status</info>   : '.$status);
```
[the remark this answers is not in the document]
maxhelias: This is already the case, if we get no recipes from Flex, an error message appears on line 82

src/Command/RecipesCommand.php
```diff
@@ -0,0 +1,253 @@
...
+    /** @var \Symfony\Flex\Flex */
+    private $flex;
+
+    /** @var Lock */
+    private $symfonyLock;
+
+    public function __construct(/* cannot be type-hinted */ $flex)
+    {
+        $this->flex = $flex;
+        $this->symfonyLock = new Lock(getenv('SYMFONY_LOCKFILE') ?: str_replace('composer.json', 'symfony.lock', Factory::getComposerFile()));
```
weaverryan: I think this can be passed in via the constructor. It's created inside `Flex::activate()` before that method initializes the commands. I'm not sure why it wasn't also passed directly to the `SyncRecipesCommand` (which also instantiates the `new Lock()` like this.

src/Command/RecipesCommand.php
```diff
@@ -0,0 +1,253 @@
...
+        } else {
+            $locker = $this->getComposer()->getLocker();
+            $lockData = $locker->getLockData();
+
+            // Merge all packages installed
+            $packages = array_merge($lockData['packages'], $lockData['packages-dev']);
+        }
+
+        $operations = [];
+        foreach ($packages as $key => $value) {
```
weaverryan: It looks like `$key` isn't needed here.

src/Command/RecipesCommand.php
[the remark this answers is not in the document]
weaverryan: So I think we should remove this line entirely for now.

src/Command/RecipesCommand.php
```diff
@@ -0,0 +1,253 @@
...
+        } elseif (null === $lockRef && null !== $recipe->getRef()) {
+            $status = '<comment>recipe not installed</comment>';
+        } elseif ($recipe->getRef() !== $lockRef) {
+            $status = '<comment>upate available</comment>';
+        }
+
+        $io = $this->getIO();
+        $io->write('<info>name</info>     : '.$recipe->getName());
+        $io->write('<info>version</info>  : '.$recipeLock['version']);
+        $io->write('<info>repo</info>     : '.$recipeLock['recipe']['repo']);
```
weaverryan: For an auto-generated recipe, there is no `recipe` key - so we get an unknown array key here.

src/Command/RecipesCommand.php
```diff
@@ -0,0 +1,253 @@
...
+        $lockRef = $recipeLock['recipe']['ref'] ?? null;
+        $lockFiles = $recipeLock['files'] ?? null;
+
+        $status = '<comment>up to date</comment>';
+        if ($recipe->isAuto()) {
+            $status = '<comment>auto-generated recipe</comment>';
+        } elseif (null === $lockRef && null !== $recipe->getRef()) {
+            $status = '<comment>recipe not installed</comment>';
+        } elseif ($recipe->getRef() !== $lockRef) {
+            $status = '<comment>upate available</comment>';
```
weaverryan:
```suggestion
            $status = '<comment>update available</comment>';
```

src/Command/RecipesCommand.php
```diff
@@ -0,0 +1,253 @@
...
+
+            $tree = $this->generateFilesTree($lockFiles);
+
+            $this->displayFilesTree($tree);
+        }
+
+        // TODO : Uncomment the line when the feature will be implemented
+        //$io->write([
+        //    '',
+        //    'Update this recipe by running composer symfony:recipe:update symfony/framework-bundle',
```
weaverryan:
```suggestion
        //    'Update this recipe by running composer:',
        //    '<info>symfony:recipe:update symfony/framework-bundle -v</info>'
```

That's some reformatting, but I'd also like to add the `-v` to it - that will give a bit more information about what the recipe update is doing (what files it's creating, for example).

src/Command/RecipesCommand.php
```diff
@@ -0,0 +1,253 @@
...
+    {
+        $this->flex = $flex;
+        $this->symfonyLock = new Lock(getenv('SYMFONY_LOCKFILE') ?: str_replace('composer.json', 'symfony.lock', Factory::getComposerFile()));
+
+        parent::__construct();
+    }
+
+    protected function configure()
+    {
+        $this->setName('symfony:recipes')
```
weaverryan: Almost all the other `symfony:` commands also have an alias to a shorter version - i.e. `recipes`. Personally, I think that would be nice.

src/Command/RecipesCommand.php
weaverryan: Also, how about making this `url` instead. And instead of `github.com/symfony/recipes`, we would print (for example): `https://github.com/symfony/recipes/twig/extensions/1.0` - I think it would be more useful.

src/Command/RecipesCommand.php
```diff
@@ -0,0 +1,253 @@
...
+        if (null !== $lockFiles) {
+            $io->write('<info>files</info>    : ');
+            $io->write('');
+
+            $tree = $this->generateFilesTree($lockFiles);
+
+            $this->displayFilesTree($tree);
+        }
+
+        // TODO : Uncomment the line when the feature will be implemented
```
weaverryan: Remove this line

src/Command/RecipesCommand.php
```diff
@@ -0,0 +1,253 @@
...
+        $subTree = [];
+        if (null !== $current) {
+            $subTree[$current] = $this->addNode($node);
+        }
+
+        return $subTree;
+    }
+
+    /**
+     * TODO : We do not display file modification information with Configurator like ComposerScripts, Container, DockerComposer, Dockerfile, Env, Gitignore and Makefile.
```
weaverryan:
```suggestion
     * Note : We do not display file modification information with Configurator like ComposerScripts, Container, DockerComposer, Dockerfile, Env, Gitignore and Makefile.
```

src/Command/RecipesCommand.php
```diff
@@ -0,0 +1,253 @@
...
+            $additional = '';
+            if (null === $lockRef && null !== $recipe->getRef()) {
+                $additional = '<comment>(recipe not installed)</comment>';
+            } elseif ($recipe->getRef() !== $lockRef) {
+                $additional = '<comment>(update available)</comment>';
+            }
+            $write[] = sprintf(' * %s %s', $name, $additional);
+        }
+
+        // TODO : Uncomment the lines following the implemented features
```
weaverryan: I would remove the TODO so we can merge - and we will add the lines below when we create those commands

src/Command/RecipesCommand.php
```diff
@@ -0,0 +1,253 @@
...
+
+            $tree = $this->generateFilesTree($lockFiles);
+
+            $this->displayFilesTree($tree);
+        }
+
+        $io->write([
+            '',
+            'Update this recipe by running composer:',
+            '<info>symfony:recipes:install symfony/framework-bundle --force -v</info>',
```
weaverryan: Make this dynamic - I totally missed that :p

src/Command/RecipesCommand.php
```diff
@@ -0,0 +1,253 @@
...
+            $status = '<comment>recipe not installed</comment>';
+        } elseif ($recipe->getRef() !== $lockRef) {
+            $status = '<comment>update available</comment>';
+        }
+
+        $io = $this->getIO();
+        $io->write('<info>name</info>     : '.$recipe->getName());
+        $io->write('<info>version</info>  : '.$recipeLock['version']);
+        if (!$recipe->isAuto()) {
+            $io->write('<info>repo</info>     : '.sprintf('https://%s/tree/master/%s/%s', $recipeLock['recipe']['repo'], $recipe->getName(), $recipeLock['version']));
```
weaverryan: It would be awesome to make this clickable: https://symfony.com/blog/new-in-symfony-4-3-console-hyperlinks

We might need to determine somehow if this feature is available / if `symfony/console` 4.3 is available.

maxhelias: I agree but I feel like it doesn't look like it :/
And we are dependent on composer on this side

weaverryan: indeed - even though composer allows newer versions, it looks like they still package an old version https://github.com/composer/composer/blob/1.9.1/composer.lock

src/Command/RecipesCommand.php
```diff
@@ -0,0 +1,256 @@
...
+            if (null === $lockRef && null !== $recipe->getRef()) {
+                $additional = '<comment>(recipe not installed)</comment>';
+            } elseif ($recipe->getRef() !== $lockRef) {
+                $additional = '<comment>(update available)</comment>';
+            }
+            $write[] = sprintf(' * %s %s', $name, $additional);
+        }
+
+        $write[] = '';
+        $write[] = '<fg=blue>Run</>:';
```
weaverryan:
```suggestion
        $write[] = 'Run:';
```

The blue color doesn't really add much here - and it's such a short word (and now I'm highlighting the below in green so less color here looks better)

src/Flex.php
```diff
@@ -402,7 +403,8 @@ public function install(Event $event = null)
             copy($rootDir.'/.env.dist', $rootDir.'/.env');
         }
 
-        $recipes = $this->fetchRecipes();
+        $recipes = $this->fetchRecipes($this->operations);
+        $this->operations = [];     // Reset the operation after getting recipes
```
weaverryan: Review note: This was done so that we could re-use `$this->fetchRecipes()` elsewhere - but keep "clearing" `$this->operations`, which the existing code does currently.

src/Command/RecipesCommand.php
```diff
@@ -0,0 +1,256 @@
...
+
+            $tree = $this->generateFilesTree($lockFiles);
+
+            $this->displayFilesTree($tree);
+        }
+
+        $io->write([
+            '',
+            'Update this recipe by running composer:',
+            sprintf('<info>symfony:recipes:install %s --force -v</info>', $recipe->getName()),
```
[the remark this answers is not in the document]
maxhelias: So we need to add the alias on #565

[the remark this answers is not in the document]
maxhelias: BTW, as we remove the prefix `symfony', do we keep the support as a name and define an alias? Or we remove directly on the name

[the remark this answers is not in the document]
weaverryan: Yes we should add that alias! I added a comment about it, but I think it got hidden. Should be better now: https://github.com/symfony/flex/pull/565/files#r356801818

> BTW, as we remove the prefix `symfony', do we keep the support as a name and define an alias? Or we remove directly on the name

We should keep the `symfony:` prefix (so this would still be called `symfony:recipes` and the other `symfony:recipes:install`) but both should have aliases (`recipes` and `recipes:install`)

If I'm not answering the correct question, ping me on Slack ;)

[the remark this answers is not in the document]
maxhelias: Yes, I did miss it.

nicolas-grekas:
Can you please rebase the PR?
Maybe also update the screenshots to help next reviewers?

maxhelias:
Here is the result : 

Capture d’écran 2019-12-13 à 17 08 36

Capture d’écran 2019-12-13 à 17 09 25

nicolas-grekas:
Thank you @maxhelias.

nicolas-grekas:
(dunno what happened to the attached patch but the commit is in:
https://github.com/symfony/flex/commit/77a0cc9d3b4688e404b26d6aa2dce205dd3a41bf)

nicolas-grekas:
(dunno what happened to the attached patch but the commit is in:
https://github.com/symfony/flex/commit/77a0cc9d3b4688e404b26d6aa2dce205dd3a41bf)

PR #564, opened 2019-10-01
Uniformize the text tags

maxhelias:


fabpot:
I'd rather change ending tags that are not `</>`.

maxhelias:
I'm going to change that, but what's the benefit of having only `</>` ? 

In fact, I chose this way to correct also PhpStorm's errors
Capture d’écran 2019-10-01 à 09 37 50

fabpot:
It's shorter and you cannot have a closing tag that does not match the opening one.

PR #565, opened 2019-10-02
Rename SyncRecipes on InstallRecipes

maxhelias:
Step 2 of #518, `symfony:recipes:install` command

### DIFF
--- a/src/Command/InstallRecipesCommand.php
+++ b/src/Command/InstallRecipesCommand.php
@@ -22,7 +22,7 @@ use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Flex\Event\UpdateEvent;
 use Symfony\Flex\Lock;
 
-class SyncRecipesCommand extends BaseCommand
+class InstallRecipesCommand extends BaseCommand
 {
     private $flex;
     private $rootDir;
@@ -37,8 +37,8 @@ class SyncRecipesCommand extends BaseCommand
 
     protected function configure()
     {
-        $this->setName('symfony:sync-recipes')
-            ->setAliases(['sync-recipes', 'fix-recipes'])
+        $this->setName('symfony:recipes:install')
+            ->setAliases(['recipes:install', 'symfony:sync-recipes', 'sync-recipes', 'fix-recipes'])
             ->setDescription('Installs or reinstalls recipes for already installed packages.')
             ->addArgument('packages', InputArgument::IS_ARRAY | InputArgument::OPTIONAL, 'Recipes that should be installed.')
             ->addOption('force', null, InputOption::VALUE_NONE, 'Ignore the "symfony.lock" file and overwrite existing files')
--- a/src/Flex.php
+++ b/src/Flex.php
@@ -243,7 +243,7 @@ class Flex implements PluginInterface, EventSubscriberInterface
             $app->add(new Command\UpdateCommand($resolver));
             $app->add(new Command\RemoveCommand($resolver));
             $app->add(new Command\UnpackCommand($resolver));
-            $app->add(new Command\SyncRecipesCommand($this, $this->options->get('root-dir')));
+            $app->add(new Command\InstallRecipesCommand($this, $this->options->get('root-dir')));
             $app->add(new Command\GenerateIdCommand($this));
             $app->add(new Command\DumpEnvCommand($this->config, $this->options));
 


src/Command/InstallRecipesCommand.php
```diff
@@ -36,9 +36,10 @@ public function __construct(/* cannot be type-hinted */ $flex, string $rootDir)
 
     protected function configure()
     {
-        $this->setName('symfony:sync-recipes')
-            ->setAliases(['sync-recipes', 'fix-recipes'])
+        $this->setName('symfony:recipes:install')
+            ->setAliases(['symfony:sync-recipes', 'sync-recipes', 'fix-recipes'])
```
fabpot: Can we also print a deprecation message when one uses an old alias?

maxhelias: Yes, I just added it 
I'm not sure about the message, can you tell me if it's good for you ?

src/Command/InstallRecipesCommand.php
```diff
@@ -36,15 +36,20 @@ public function __construct(/* cannot be type-hinted */ $flex, string $rootDir)
...
             ->setDescription('Installs or reinstalls recipes for already installed packages.')
+            // TODO : Separate the force option in an update command
             ->addOption('force', null, InputOption::VALUE_NONE, 'Ignore the "symfony.lock" file and overwrite existing files')
         ;
     }
 
     protected function execute(InputInterface $input, OutputInterface $output): int
     {
+        if (in_array($input->getFirstArgument(), $this->getAliases(), true)) {
+            trigger_error(sprintf('The "%s" alias is deprecated. Please use `symfony:recipes:install`.', $input->getFirstArgument()), E_USER_DEPRECATED);
```
fabpot: ... is deprecated; please use ... instead.

src/Command/InstallRecipesCommand.php
```diff
@@ -36,15 +36,20 @@ public function __construct(/* cannot be type-hinted */ $flex, string $rootDir)
 
     protected function configure()
     {
-        $this->setName('symfony:sync-recipes')
-            ->setAliases(['sync-recipes', 'fix-recipes'])
+        $this->setName('symfony:recipes:install')
+            ->setAliases(['symfony:sync-recipes', 'sync-recipes', 'fix-recipes'])
             ->setDescription('Installs or reinstalls recipes for already installed packages.')
+            // TODO : Separate the force option in an update command
             ->addOption('force', null, InputOption::VALUE_NONE, 'Ignore the "symfony.lock" file and overwrite existing files')
```
weaverryan: Remove the TODO so this can be merged

nicolas-grekas: ping @maxhelias

src/Command/InstallRecipesCommand.php
```diff
@@ -36,15 +36,20 @@ public function __construct(/* cannot be type-hinted */ $flex, string $rootDir)
...
         ;
     }
 
     protected function execute(InputInterface $input, OutputInterface $output): int
     {
+        if (\in_array($input->getFirstArgument(), $this->getAliases(), true)) {
+            trigger_error(sprintf('"%s" is deprecated. Please use `symfony:recipes:install` instead.', $input->getFirstArgument()), E_USER_DEPRECATED);
```
nicolas-grekas: I think we don't need to trigger this: using the alias is fine, and we're not going to release flex v2 just for this.

src/Command/InstallRecipesCommand.php
```diff
@@ -36,15 +36,20 @@ public function __construct(/* cannot be type-hinted */ $flex, string $rootDir)
 
     protected function configure()
     {
-        $this->setName('symfony:sync-recipes')
-            ->setAliases(['sync-recipes', 'fix-recipes'])
+        $this->setName('symfony:recipes:install')
+            ->setAliases(['symfony:sync-recipes', 'sync-recipes', 'fix-recipes'])
             ->setDescription('Installs or reinstalls recipes for already installed packages.')
+            // TODO : Separate the force option in an update command
```
nicolas-grekas: I'm not sure about this: if we add an update command, it might be totally different than the current behavior of install --force

src/Command/InstallRecipesCommand.php
maxhelias: it was requested by @fabpot, but I can delete it, if it's good for both

nicolas-grekas: I'm in favor of removing it
/cc @fabpot for confirmation

weaverryan:
Needs a rebase now!

maxhelias:
The fabbot failing is due to the trigger_error which cannot be silent otherwise it does not appear

src/Command/InstallRecipesCommand.php
weaverryan: Also, should we add a "short" alias for `recipes:install` - most other `symfony:` commands have that short alias: https://github.com/symfony/flex/pull/562/files#diff-3354bea129aa2979af719cb4c3662f34R41

maxhelias:
It's ready to merge :)

nicolas-grekas:
Thank you @maxhelias.

PR #566, opened 2019-10-14
non-int value in the execute command will be deprecated

maxhelias:


fabpot:
Thank you @maxhelias.

PR #568, opened 2019-11-13
fix wrong bypass of parallel downloads

dmaicher:
On a big work project I noticed that with an empty composer cache the pre-fetching of packages has become really slow lately.

Here a comparison between flex and `hirak/prestissimo`:

flex:
https://blackfire.io/profiles/eb7cfd59-6d88-4035-b210-661a5f0de666/graph

prestissimo:
https://blackfire.io/profiles/07b46d2e-3912-41c5-b25c-5b39a3103de9/graph

In my case curl was not used at all :stuck_out_tongue: 

This PR aims to fix the wrong bypass condition that was added within https://github.com/symfony/flex/pull/561.

Thanks @nicolas-grekas for the help.

### DIFF
--- a/src/ParallelDownloader.php
+++ b/src/ParallelDownloader.php
@@ -244,7 +244,7 @@ class ParallelDownloader extends RemoteFilesystem
             return $result;
         }
 
-        if (!$this->downloader || !preg_match('/^https?:/', $originUrl)) {
+        if (!$this->downloader || !preg_match('/^https?:/', $fileUrl)) {
             return parent::getRemoteContents($originUrl, $fileUrl, $context, $responseHeaders);
         }
 


src/ParallelDownloader.php
```diff
@@ -244,7 +244,7 @@ protected function getRemoteContents($originUrl, $fileUrl, $context, array &$res
             return $result;
         }
 
-        if (!$this->downloader || !preg_match('/^https?:/', $originUrl)) {
+        if (!$this->downloader || !preg_match('/^https?:/', $fileUrl)) {
```
dmaicher: `$originUrl` is always `github.com` in my case.

and

`$fileUrl` is something like `https://codeload.github.com/symfony/service-contracts/legacy.zip/ffc7f5692092df31515df2a5ecf3b7302b3ddacf?access_token=60f62f9b058d01dd4c051bda82394bc5d50b5190`

dmaicher: @kevinpapst this will also still solve the issue you fixed in https://github.com/symfony/flex/pull/561, right?

kevinpapst: Yes, definitely. All that was missing was the check for https.

kevinpapst:
Can't remember what happened back then... but thanks for finding and fixing this one! 

A quick check on a current project revealed 
```
echo $fileUrl . ' # '. $originUrl . PHP_EOL;

https://repo.packagist.org/packages.json # repo.packagist.org
http://repo.packagist.org/p/provider-latest%2478290d5e9564c411f926482cc97e72347d1a3c20d7800ff867b4b05bb338a4f2.json # repo.packagist.org
https://flex.symfony.com/versions.json # https://flex.symfony.com
... and so on
```

for a local package it is always the same
```
var/packages/DemoBundle.zip # var/packages/DemoBundle.zip
```

So using `$fileUrl` looks like the right choice.

fabpot:
Thank you @dmaicher.

Issue #206, opened 2017-11-14
Support for bundle inline recipe configuration


linaori:
Now that Flex has matured a bit, I would like to bring this point up again. Currently, having dozens of internal bundles (and several open-source), having dozens of applications using those bundles, we prefer to follow the recommendations and standards as much as possible. This means that we will start using Flex from 4.0 on.

The selling point of flex, is the ability to provide an "installation" into the application by using manifests. However, this being in an external repository, poses several issues:
 - It doesn't work for closed source bundles, I can't add them.
 - Double bookkeeping, if I add a new configuration option or provide something new, I have to remember to update the recipe in another repository as well.
 - If I would use a closed source variant, I'd need to set this up and maintain this repository

To me, these are a few blockers that prevent Flex from being useful for me. I have dozens of bundles that have a mandatory configuration, but I won't be able to utilize flex to increase the Developer experience. It will actually make behavior different between the recipe'd packages and internal packages when it comes to installing, making it annoying to find out whether or not you still have to add some config file for example.

---

> Recipes must be stored on their own repositories, outside of your Composer package repository
> https://github.com/symfony/recipes/blob/master/README.rst#creating-recipes

I'm not sure if this is a technical limitation, but I would love to be able to do the exact opposite, inline the configuration in my package. This gives me several benefits:
 - My package provides optional integration into flex (if used).
 - I can control multiple version based on the current version of the files, rather than making a directory per version.
 - When I release my bundle version, I don't have to rely on the recipe repository to make it possible to install the newest config, I control the release.
 - I won't forget to update the recipe repository, nor does someone else have to do this when a new feature is introduced.
 - No need for privately hosted repositories, mirrors or proxies.

---

The directory structure of an inlined flex config could be as simple as adding a `.flex` directory or something similar.

Pierstoval:
> The directory structure of an inlined flex config could be as simple as adding a  ̀.flex ̀ directory or something similar.

If this looks like a current manifest from Flex recipes, I think it's a nice idea and can be implemented without too much problems.

Composer already provides `post-package-update` and `post-package-install` events that give the advantage of just opening the directory containing the package, instead of downloading the manifest from symfony.sh :thinking:

Grafikart:
I think this is a must have. A simple solution would be to add a `flex.json` file within a project.
Then, when we do a `composer require` the `post-package-install` hook will look for this file and run the recipe if needed. 

It's more convenient for a developer to manage the recipes from his repository. And it will allow private recipes easily.

[...]

PR #572, opened 2019-11-19
Enable support for inline recipes (local recipe inside the package itself)

drupol:
This PR:

* [x] Relates to #206 
* [x] Does not have tests yet
* [x] Does not break the actual behavior of the plugin
* [x] Needs some feedback
* [x] Needs some polishing

This PR enable support for recipe inside a package.

Basically when you install a package that doesn't have an official recipe and if the `composer.json` has ['extra']['recipe'], then that property will be used as recipe.

I personally really would like to see this in because I don't like so much the actual process of submitting new recipes and based on comments from #206, I think it would be a nice feature.

### DIFF
--- a/src/Flex.php
+++ b/src/Flex.php
@@ -711,10 +711,25 @@ class Flex implements PluginInterface, EventSubscriberInterface
 
             $noRecipe = !isset($manifests[$name]) || (isset($manifests[$name]['not_installable']) && $manifests[$name]['not_installable']);
             if ($noRecipe && 'symfony-bundle' === $package->getType()) {
-                $inlineRecipe = $this->composer->getInstallationManager()->getInstallPath($package).'/manifest.json';
-                if (file_exists($inlineRecipe)) {
+                // Get the lock data from all packages, including dev.
+                $lockDataPackages = array_merge(
+                    $this->composer->getLocker()->getLockData()['packages'],
+                    $this->composer->getLocker()->getLockData()['packages-dev']
+                );
+
+                // Get lock data of the package.
+                $lockDataPackage = current(
+                    array_filter(
+                        $lockDataPackages,
+                        static function ($lockData) use ($package) {
+                            return $lockData['name'] === $package->getName();
+                        }
+                    )
+                );
+
+                if (isset($lockDataPackage['extra']['recipe']) && \is_array($lockDataPackage['extra']['recipe'])) {
                     $manifest['origin'] = sprintf('%s:%s@inline recipe', $name, $package->getPrettyVersion());
-                    $manifest['path'] = $inlineRecipe;
+                    $manifest['recipe'] = $lockDataPackage['extra']['recipe'];
                     $recipes[$name] = new Recipe($package, $name, $job, $manifest);
                 } else {
                     $manifest = [];
--- a/src/Flex.php
+++ b/src/Flex.php
@@ -711,7 +711,7 @@ class Flex implements PluginInterface, EventSubscriberInterface
 
             $noRecipe = !isset($manifests[$name]) || (isset($manifests[$name]['not_installable']) && $manifests[$name]['not_installable']);
             if ($noRecipe && 'symfony-bundle' === $package->getType()) {
-                $inlineRecipe = $this->composer->getInstallationManager()->getInstallPath($package) . '/manifest.json';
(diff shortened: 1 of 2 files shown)


pamil:
This is an idea that we thought about in Sylius a lot when we started to craft recipes for Symfony Flex. In my opinion, code, docs and recipes should be a part of the same repository, so that they can stay synchronised and be cross-referenced easily. 

This way I could create a test Symfony application in my repo and make sure my recipes are correct within single CI build.

Another common case for me is to provide a new feature flag in bundle's configuration to keep backwards compatibility but allow for the usage of new behaviour. It looks like the following:

```yaml
my_bundle:
  new_behaviour: true # defaults to false
```

I want existing packages to use the old behaviour, but any new installation should get the new one. Currently, it requires two PRs, one to my bundle repo and one to recipes-contrib repo (which also require my library to be tagged).

The only thing I'm not sure is whether Flex recipe should stay in `manifest.json` in the same path as `composer.json` or be embedded into `composer.json`'s `extra` configs. I think I prefer the latter (so that there are fewer config files and less chance for conflict with tools that use `manifest.json` file).

bobdenotter:
For a project like Bolt, this feature would be very useful as well! 👍

Pierstoval:
I'm not sure that `manifest.json` is the right path.

Instead, I would rename the file to `flex-recipe.json` to be clearer about what it does. The `manifest.json` should stay in the recipes repository.

The reason is that there are some other tools that use the `manifest.json` convention in their settings, such as YunoHost.

nicolas-grekas:
Please don't spend too much time on this, it has been rejected in the past and will be again: a recipe is an opinionated way to use a library. The way some code is used should be totally decoupled from the code itself. That's two completely decoupled domains and they should remain decoupled technically too. Linux distributions do it right if you want a comparison: library authors ship code but don't care about packaging it for specific distributions - then each distribution has a dedicated configuration per package that fits their philosophy. That's the proper separation of concern. Same here.

:-1:

drupol:
> The only thing I'm not sure is whether Flex recipe should stay in `manifest.json` in the same path as `composer.json` or be embedded into `composer.json`'s `extra` configs. I think I prefer the latter (so that there are fewer config files and less chance for conflict with tools that use `manifest.json` file).

> I'm not sure that `manifest.json` is the right path.

I moved the content of it into `composer.json['extra']['recipe']` to avoid the extra file in the repo.

@nicolas-grekas 
Ok I understand, I just wanted to find a way to get rid of the extra work and waste of time that people have to do in order to publish a recipe. Couldn't we find a good compromise ?

I've updated the PR. Instead of relying on a file `manifest.json` at the root of the package, I'm relying on configuration found in `composer.json['extra']['recipe']`.

Sadly, as this seems to be a dead end, I'll leave this here and I won't update it from now on.

nicolas-grekas:
>  Ok I understand, I just wanted to find a way to get rid of the extra work and waste of time that people have to do in order to publish a recipe. Couldn't we find a good compromise ?

Note that I'm not specifically aware of any facts supporting the narrative about "extra work and waste of time". Most of the discussions around the topic are quite constructed to my understanding.

Anyway, thank you for trying!
I'm closing therefor.

drupol:
> Note that I'm not specifically aware of any facts supporting the narrative about "extra work and waste of time". Most of the discussions around the topic are quite constructed to my understanding.

What I meant by "extra work and waste of time" is that, in order to publish a recipe:

* Fork `symfony/flex`
* Do a PR
* Wait for the automatic validation
* Fix if necessary
* Wait for an owner to manually validate
* Fix if necessary

Maybe the wording was not the best,true.
What I'm just trying to do is to reduce that amount of time to something as fast as publishing a package on Packagist for example.

Anyway, I'll find a better alternative.

itsjavi:
This functionality can be probably replaced by composer hooks, at the cost of reinventing the wheel, kind of.

I still don't get why flex is so opinionated and you can't give the choice to the developers to use the centralised recipe server or not. This will only make us reinventing the wheel.

It's like if someone now says that all composer.json scripts will have to live in a centralised service and will not be allowed to be in the repositories themselves anymore.

I hope you are more open minded about this at some point, because it seems the community is trying to push for this for an alternative for this.

PR #577, opened 2019-11-30
Relax PHP version constraint - dont lock when run from globals

nicolas-grekas:


fabpot:
Thank you @nicolas-grekas.

PR #580, opened 2019-12-01
Fix computing the intersection of Symfony components with extra.symfony.require

nicolas-grekas:
A constraint like `">5.0.0,<5.1"` in `SYMFONY_REQUIRE` doesn't work right now, this fixes it.


nicolas-grekas:
PR ready. FYI the implementation could be a bit more robust if https://flex.symfony.com/versions.json could list the latest stable version of each branch.

fabpot:
Thank you @nicolas-grekas.

Issue #470, opened 2019-01-22
Make Flex more Granular ?


DavG:
Hello,

Flex is great, but honestly, it is a little too magic sometimes. I mean, really.

Flex is trying to force me to adopt some practices I dont want to.
For example, I use "real" environment variables, in all environment.
So, I don't use Dotenv. And I like that Symfony doesn't force me to.
But, Flex will create/update a .env.dist file.
Same with phpunit.xml.dist.
What if I have real env var ?
What if I use another environment variable naming convention ?
That was supposed to speed up things, but for me, it's a slow down.

So, firstly, I think we should be able to tell Flex "ignore this recipe part" (.env.dist)
In this special case, maybe this could be automatical ; Flex could check if Dotenv is installed.

Secondly, the recipe part, in general, should be disabled in non-dev environment.
Ok, theoricaly, if you commit well your symfony.lock , recipes will do nothing in prod. But it seems dangerous to me.

Maybe the best option would be to split the package in 2 parts : 
- symfony/flex (for all env, in "require"): symfony scripts, composer parallel downloads, etc.
- symfony/flex-dev (dev only, in "require-dev") : all about helping developers, so recipes goes here. 

I don't get why recipe code should be on the production server.

What do you think, @fabpot ?

Thanks.

Pierstoval:
> What if I have real env var ?

You can disable DotEnv lines in `config/bootstrap.php`. But even then, if you already have real env vars, DotEnv doesn't override existing vars, so you're safe anyway.

> What if I use another environment variable naming convention ?

Change var names, after all they're yours once they're added to your projects 🙂 
The current var names are just basic conventions that suit most projects, but if it's not the case for yours, you can change it.

> Maybe the best option would be to split the package in 2 parts :
> 
> * symfony/flex (for all env, in "require"): symfony scripts, composer parallel downloads, etc.
> * symfony/flex-dev (dev only, in "require-dev") : all about helping developers, so recipes goes here.

Seems like it'll introduce way too much maintenance from the team and it would not really be useful for DX. What I see here is that you need to disable "some parts" of the Flex package configuration automation, not Flex itself.

Something that could be introduced in Flex, maybe, is a `extra.symfony.disable-configurators` array option in order for you to be able to disable the `env` configurator, as well as other configurators (like `copy-from-recipe` if you want to almost completely disable the recipes system).

I don't see any more straightforward flex-related option for your case.

Let's wait for other comments 🙂

DavG:
[...]

PR #581, opened 2019-12-05
Allow disabling configurators

mnavarrocarter:
Flex is amazing, and it could be even more amazing if we give it some flexibility (no pun intended) by allowing the user to disable the configurators they want. 

For example, I often find myself in the situation of wanting all the amazing work flex does for me in terms of copying the `.yaml` config files, adding the bundle to the array, adding relevant `.gitignore` entries and environment variables, and so on and so forth. The only thing I don't like is when flex writes in my `src` folder. I like to define my own directory structure, so I don't like flex writing there.

Other users have expressed similar concerns. In #470 @DavG explains that he does not like flex writing .env vars, for example. It is reasonable to assume that other users would like flex not doing certain things. 

This PR implements a little tweak in the process of registering Configurators. It looks in the `extra/symfony/disable` composer.json path and sees if the name of the configurator is in the array. If that is the case, it does not register it. Users can fill up that array with the names of the configurators they want to disable.

So, in my case, to avoid writing files into the `src` folder:

```json
{
     "extra": {
        "symfony": {
            "allow-contrib": true,
            "require": "5.0.*",
            "disable": ["copy-from-package"]
        }
    }
}
```

Since this is a non-breaking change and fully optional, I don't see a reason why it could not work.

I don't have any idea on how to test the disabling of the configurators. Any guidance on that would be appreciated.

### DIFF
--- a/src/Configurator.php
+++ b/src/Configurator.php
@@ -31,8 +31,8 @@ class Configurator
         $this->composer = $composer;
         $this->io = $io;
         $this->options = $options;
-        // ordered list of configurators
-        $this->configurators = [
+        // We register only configurators that are not in the disabled list
+        $this->registerConfigurators([
             'bundles' => Configurator\BundlesConfigurator::class,
             'copy-from-recipe' => Configurator\CopyFromRecipeConfigurator::class,
             'copy-from-package' => Configurator\CopyFromPackageConfigurator::class,
@@ -43,7 +43,7 @@ class Configurator
             'gitignore' => Configurator\GitignoreConfigurator::class,
             'dockerfile' => Configurator\DockerfileConfigurator::class,
             'docker-compose' => Configurator\DockerComposeConfigurator::class,
-        ];
+        ]);
     }
 
     public function install(Recipe $recipe, Lock $lock, array $options = [])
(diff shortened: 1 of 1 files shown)


nicolas-grekas:
This won't have the effect you'd like to achieve: both yaml and php files are usually created by the copy-from-recipe configurator.

Overall I'm not convinced at all this would provide a better experience to anyone. A recipe that misses some ingredients is a broken one. Even if the resulting changes don't fit your way of things, it's useful to know all the changes that are needed for ending up with a consistent configuration.

Flex just provides the defaults. It doesn't mean there is nothing to do after it ran.

A better alternative IMHO would be for you to write a script that runs after flex and transforms the changes to your need if they can be automated.

mnavarrocarter:
Thanks for the feedback Nicolas. I have some points in response.

> This won't have the effect you'd like to achieve: both yaml and php files are usually created by the copy-from-recipe configurator.

Oh I see. This is terrible news for me. Not necessarily for this PR though.

> Overall I'm not convinced at all this would provide a better experience to anyone. A recipe that misses some ingredients is a broken one. Even if the resulting changes don't fit your way of things, it's useful to know all the changes that are needed for ending up with a consistent configuration.

Flexibility allows for better experience in case you need it. You still have all the flex environment working if you don't touch anything, so is a win win situation. Broken is not necessarily bad if you are aware that you are breaking something, and want to. If you would like, for example, to not copy the `.yaml` config, this would break the auto-scripts. The difference here is that you will know why it broke. The risks of breaking can be easily explained in some documentation entry.

> Flex just provides the defaults. It doesn't mean there is nothing to do after it ran.

And still will be that way if these changes merge. The only difference is that users that want to tweak things a bit more can do so now.

> A better alternative IMHO would be for you to write a script that runs after flex and transforms the changes to your need if they can be automated.

Well, better is pretty relative here. Filling up an array with some strings is better for me than writing a script. The reason for this PR in the first place is that I truly believe it is better for other people too. Not many people, granted, but at least some of us that have this problem. It's such a noninvasive change that I don't see any harm on implementing it.

nicolas-grekas:
I'm definitely :-1: here: tweaking recipes in unexpected ways (from their authors pov) can break in unexpected ways. There is no way users will understand that they opted out from working recipes and will figure out why it breaks on their own. This means this will add increased burden on this very repo - or the recipe one - while it's not in our hands.

Maybe another way could work, but this one won't happen sorry.

Issue #575, opened 2019-11-21
Documentation on Upgrading Symfony, How Flex Works


brooksvb:
There is a lack of documentation on how to properly use Symfony Flex.

For example, the behavior described in this comment is not documented anywhere:
https://github.com/symfony/flex/issues/433#issuecomment-437325751

The guide for upgrading Symfony doesn't describe how to upgrade with Flex:
https://symfony.com/doc/current/setup/upgrade_major.html#update-to-the-new-major-version-via-composer

When it comes to composer I was not very familiar with how everything worked, so trying to upgrade Symfony left me with a lot of questions that don't seem to be answered anywhere. I'm thankful that I was able to have someone explain Flex a bit better and how to modify my composer.json to upgrade by asking in Slack. But this should be available in documentation.

nicolas-grekas:
Could you please list the specific things you missed and you learned in the process?

brooksvb:
This is just going to be a brain dump so it might not be great format and it might not all be strictly relevant to Flex.

I was able to be ignorant to how flex worked for a while because all I had to do was `composer require X` for everything like usual, but now I have some nice quick aliases and it does some auto-magical config stuff for me. The problems arose when I wanted to upgrade Symfony.

* Why are all these component versions `4.3.5` or similar, but there's a couple that do not follow that scheme?
* What is this `extra` key at the bottom of my `composer.json`? What's looking at that or using that and how?
* The upgrade documentation says to update the `symfony/symfony` requirement, but in my case, what do I do? Do I have to change this `extra` key? Do I need to update every component version? What about the odd-ones-out with the unusual version numbers?
* How does flex work? How does it hook into the composer workflow?
* Why is there this `conflict: symfony/symfony` thing? (I realize this is answered briefly on [Upgrading Existing Applications to Symfony Flex](https://symfony.com/doc/current/setup/flex.html) but isn't this a weird place to find it? If I never upgraded because I started from scratch, how would I figure out what it is?
* What is this weird format in `scripts: auto-scripts` that refers to `symfony-cmd`? How does that work? I especially feel this should be explained because this is something someone might want to leverage this `symfony-cmd` more for additional tasks.

Examples of the "odd-ones-out":
```
symfony/maker-bundle
symfony/orm-pack
symfony/webpack-encore-bundle
symfony/flex
```
Some key points I learned:
[...]

PR #583, opened 2019-12-08
Allowing specific recipes to be "re-installed"

weaverryan:
Hi! 

Very simple: you can now re-install/update just one (or specific) recipes:

```
composer symfony:sync-recipes symfony/console --force
```

That's very helpful, as there are a few specific recipes (this is one of them) that really *do* need to be "updated", at the very least, when upgrading to Symfony 5 (as some important files have evolved over time).

This relates to #575 (will help with upgrading), #518 and should go along with the PRs #562 and #565

Thanks!

### DIFF
--- a/src/Command/SyncRecipesCommand.php
+++ b/src/Command/SyncRecipesCommand.php
@@ -15,6 +15,7 @@ use Composer\Command\BaseCommand;
 use Composer\DependencyResolver\Operation\InstallOperation;
 use Composer\Factory;
 use Symfony\Component\Console\Exception\RuntimeException;
+use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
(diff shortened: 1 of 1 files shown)


src/Command/SyncRecipesCommand.php
```diff
@@ -58,24 +60,47 @@ protected function execute(InputInterface $input, OutputInterface $output): int
...
             if ($force || !$symfonyLock->has($pkg['name'])) {
                 $packages[] = $pkg['name'];
             }
         }
 
+        $io = $this->getIO();
+
+        if ($targetPackages = $input->getArgument('packages')) {
+            if ($invalidPackages = array_diff($targetPackages, $totalPackages)) {
+                $io->writeError(sprintf('<warning>Invalid package name(s):</warning> %s', implode(', ', $invalidPackages)));
```
nicolas-grekas: "invalid"' means "not installed"? maybe we should tell it this way?

src/Command/SyncRecipesCommand.php
```diff
@@ -58,24 +60,47 @@ protected function execute(InputInterface $input, OutputInterface $output): int
...
+                $io->writeError('Re-run the command with <info>--force</info> to re-install the recipes.');
+                $io->writeError('');
+            }
+
+            $packages = array_diff($targetPackages, $packagesRequiringForce);
+        }
+
         if (!$packages) {
+            $io->writeError('No recipes to install.');
+            // print: nothing to install
```
nicolas-grekas: should be removed I guess

weaverryan:
This is ready to go

fabpot:
Thank you @weaverryan.

PR #585, opened 2019-12-11
Adds the URL to the exact recipe that is installed to "symfony recipes <package>"

weaverryan:
Hi!

This depends & builds off of #562 and relates to https://github.com/symfony/flex/pull/562#discussion_r355156496

Basically, I *thought* that it would be nearly impossible to get the *exact* git sha for when a recipe is installed. But with a few API calls to GitHub, it's quite possible :). This gives us the ability to (at least) give the URL to the *exact* state of the recipe when it was installed in their app. This can help them determine what's changed and if they need to update.

Giving them a "diff" URL would be even more awesome - but as the directory often changes - I don't believe there is any URL we can give to visually compare across directories. Output:

Screen Shot 2019-12-11 at 3 55 06 PM

The "installed recipe" is the line that this enables.


src/Command/RecipesCommand.php
```diff
@@ -0,0 +1,322 @@
...
+
+        $gitSha = $recipeLock['recipe']['branch'] ?? 'master';
+        if (null !== $lockRef) {
+            $gitSha = $this->findRecipeShaFromTreeRef(
+                $recipe->getName(),
+                $recipeLock['recipe']['repo'],
+                $recipeLock['recipe']['branch'],
+                $recipeLock['version'],
+                $lockRef
+            );
```
weaverryan: This is basically the new logic

maxhelias:
Great! Awesome! You impress me 😅

src/Command/RecipesCommand.php
```diff
@@ -140,16 +142,31 @@ private function displayPackageInformation(Recipe $recipe)
...
+        }
+
         $io = $this->getIO();
-        $io->write('<info>name</info>     : '.$recipe->getName());
-        $io->write('<info>version</info>  : '.$recipeLock['version']);
+        $io->write('<info>name</info>             : '.$recipe->getName());
+        $io->write('<info>version</info>          : '.$recipeLock['version']);
+        $io->write('<info>status</info>           : '.$status);
         if (!$recipe->isAuto()) {
```
maxhelias: We can add if $gitSha is not null

src/Command/RecipesCommand.php
```diff
@@ -250,4 +267,56 @@ private function writeTreeLine($line)
...
+            return null;
+        }
+
+        $parts = explode('/', $repo);
+        if (3 !== \count($parts)) {
+            return null;
+        }
+
+        $recipePath = sprintf('%s/%s', $package, $version);
+        $commitsData = $this->requestGitHubApi(sprintf(
```
maxhelias: We also have to try/catch if the api isn't available and I think after that the command should work even offline

src/Command/RecipesCommand.php
weaverryan: If we can't figure out the `$gitSha`, I default it to the branch from the lock file (e.g. `master`) so that we can at least give an approximation. So I'd like to keep this as always showing (except when the recipe is an auto recipe)

src/Command/RecipesCommand.php
```diff
@@ -0,0 +1,333 @@
...
+                $gitSha = $this->findRecipeShaFromTreeRef(
+                    $recipe->getName(),
+                    $recipeLock['recipe']['repo'],
+                    $branch,
+                    $recipeLock['version'],
+                    $lockRef
+                );
+            } catch (TransportException $exception) {
+                $io->writeError('Error downloading exact git sha for installed recipe.');
+            }
```
weaverryan: If anything goes wrong with the downloading, we fail gracefully

weaverryan:
This is ready to go. Review note: it's built on top of #562 - I've highlighted the new section.

maxhelias:
It looks good to me 👍

nicolas-grekas:
Does this work cross versions? Eg I install the recipe for bundle v1.12 and upgrade to v2.0 now which has a separate recipe?

weaverryan:
Re-opened as #588 - I think GH is having a few issues right now (so this PR never updated after the rebase). 

> Does this work cross versions? Eg I installed the recipe for bundle v1.12 and upgraded to v2.0 now which has a separate recipe?

I mention this in my new PR description. Basically, we can't show a cool "history of commits" URL for this case... so we only show the URL to the currently-installed recipe and the URL to the "new" recipe (just no URL to see any history or diff between the two).

PR #584, opened 2019-12-11
Notifying of -v option and making output better for recipe info

weaverryan:
Slightly nicer output (the indentation):

Screen Shot 2019-12-11 at 2 38 19 PM

Super minor - sorry for extra noise :)

PR #586, opened 2019-12-12
Fix travis test

maxhelias:
Since #584, the tests have failed


fabpot:
Thank you @maxhelias.

weaverryan:
Thank you - I knew about that and then completely forgot.

Issue #379, opened 2018-07-10
Symfony.lock changes during composer install


stlrnz:
I discovered a strange behaviour when deploying Symfony Flex applications on Windows using Git and Composer. Somehow the line endings in `symfony.lock` are changing during `composer install`. Maybe someone can give me a hint what I'm doing wrong or what I can do better.

**The Environment:**
- Windows based development machine
- Windows based production sever
- Git configured with `core.autocrlf=true` on both machines

**Steps to reproduce:**
1. Create a new project on dev
   `composer create-project symfony/skeleton demo`
   `symfony.lock` is created using `LF` charater as line endings :heavy_check_mark: 

2. Add everything to Git
   `cd demo`
   `git init`
   `git add -A`

3. A warning is displayed but  `symfony.lock` still has `LF` line endings in working copy  :heavy_check_mark: 

> warning: LF will be replaced by CRLF in symfony.lock.
> The file will have its original line endings in your working directory.

4. Commit & Push
   `git commit -m "init"`
   `git remote add origin https://...`
   `git push -u origin master`

5. Clone project on prod machine
   `git clone https://... demo`
   `symfony.lock` now has `CRLF` line endings (because of `core.autocrlf=true`)  :heavy_check_mark: 

6. Check status
   `cd demo`
   `git status`

> On branch master
> Your branch is up-to-date with 'origin/master'.
> nothing to commit, working tree clean

7. Install project
   `composer install`
    `symfony.lock` now has `LF` line endings :question: 

8. check status again
    `git status`

> On branch master
> Your branch is up-to-date with 'origin/master'.
> Changes not staged for commit:
>   (use "git add <file>..." to update what will be committed)
>   (use "git checkout -- <file>..." to discard changes in working directory)
> 
>         modified:   symfony.lock
> 
> no changes added to commit (use "git add" and/or "git commit -a")

So my question is why is `symfony.lock` updated during `composer install`? Is that an expected behaviour? I thougth it should be read-only during install just like `composer.lock`?

Everytime I update my project on production I will have to revert the changes to `symfony.lock` before I can do a `git pull`. Thats not very intuitive.

Kris2x:
I noticed the same problem. This is normal?

Solthun:
Issue source is line endings in different OS, so run `git config --global core.autocrlf false && git add --all` (https://stackoverflow.com/a/49324186/2591726)

stlrnz:
Im sure this is a proper workaround.
[...]

PR #479, opened 2019-03-08
only write lock when its content has changed

stlrnz:
Fixed #379 by only writing the Lockfile when something has changed (which is not the case on `composer install`).

This seems to me the best solution for the linked issue:
* no change in the Flex behaviour on `composer install` needed (still doing the same like on `composer update`)
* only update the Lockfile when needed (even the modification timestamp will stay the same)
* working for all composer actions (because of auto detection)

Thank you for your feedback.

PR #587, opened 2019-12-13
compare data before marking lock as changed

stlrnz:
In https://github.com/symfony/flex/pull/479 I tried to fix the issue https://github.com/symfony/flex/issues/379 by introducing a _changed_ flag to the Lock.

This patch worked for a few months. However, introducing https://github.com/symfony/flex/pull/576 broke the patch by [always calling](https://github.com/symfony/flex/pull/576/files#diff-46b78198ae7ea525f04268205dd782c3R273) `set` ending in the same situation as before.

Blame on me for not checking if anything differs before setting the _changed_ flag. This PR tries to do better.

### DIFF
--- a/src/Lock.php
+++ b/src/Lock.php
@@ -49,14 +49,18 @@ class Lock
 
     public function set($name, $data)
     {
-        $this->lock[$name] = $data;
-        $this->changed = true;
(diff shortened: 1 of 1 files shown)


nicolas-grekas:
Thank you @stlrnz.

PR #588, opened 2019-12-13
Adding the current recipe URL to recipes <recipe> output

weaverryan:
Hi!

This replaces #585 and builds off of #562. 

Basically, I *thought* that it would be nearly impossible to get the *exact* git sha for when a recipe is installed. But with a few API calls to GitHub, it's quite possible :). This gives us the ability to (at least) give the URL to the *exact* state of the recipe when it was installed in their app. This can help them determine what's changed and if they need to update.

Giving them a "diff" URL would be even more awesome - but as the directory often changes - I don't believe there is any URL we can give to visually compare across directories. Output:

Screen Shot 2019-12-12 at 12 06 54 PM

The "installed recipe" is the line that this enables. The "new commits" line (unfortunately) can only be shown when the version/directory of the recipe hasn't changed. There's just not good way (that I can think of) to show a decent history if they are upgrading from (for example) `symfony/framework/bundle/3.3` to `symfony/framework-bundle/4.2`.

Cheers!

### DIFF
--- a/src/Command/RecipesCommand.php
+++ b/src/Command/RecipesCommand.php
@@ -12,11 +12,13 @@
 namespace Symfony\Flex\Command;
 
 use Composer\Command\BaseCommand;
+use Composer\Downloader\TransportException;
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Flex\InformationOperation;
 use Symfony\Flex\Lock;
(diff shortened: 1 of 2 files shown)


nicolas-grekas:
Getting the tree is super nice!
I'm wondering if the links are that useful: I clicked on one for symfony/console and fine, there were symlinks + what should I do about that?

Something super useful would be to be able to display the diff. I'm sure we can checkout the files locally and use git to render the diff.

:gift_heart:

nicolas-grekas:
Thank you @weaverryan.

PR #589, opened 2019-12-13
showing commit history for entire recipe - not just specific version

weaverryan:
This allows us to always show the history - even if the version changed


nicolas-grekas:
Thank you @weaverryan.

PR #600, opened 2020-01-17
Test improvements, use Nowdoc instead of Heredoc if possible

fre5h:


Pierstoval:
Is there any particular reason for this change? I'm curious

fre5h:
@Pierstoval From the official PHP documentation
```
A nowdoc is specified similarly to a heredoc, but no parsing is done inside a nowdoc.
```
https://www.php.net/manual/en/language.types.string.php#language.types.string.syntax.nowdoc

Pierstoval:
I know this difference, but this doesn't seem to me like a _recommendation_. There's no mention of heredoc nor nowdoc in PSRs either, that's why I'm asking why you think it would be better to use one over the other in Flex.

fre5h:
It's only the micro-optimization. I found it, when explored code of Symfony Flex. I updated only those strings, which don't contain php variables inside, so heredoc could be changed with nowdoc without side effects.

Pierstoval:
Seems fair :+1:

nicolas-grekas:
Thank you. I'm going to close because we don't do CS changes when they can't be enforced by out CS checker, fabbot based on php-cs-fixer. Without a bot, these changes are arbitrary and the "target" achieved here won't last long as a future PR might break it very easily.

Thanks for proposing.

gharlan:
(php-cs-fixer has a fixer for this (`HeredocToNowdocFixer`), but symfony don't want to use it: https://github.com/FriendsOfPHP/PHP-CS-Fixer/pull/2413)

PR #602, opened 2020-01-22
Update autoload.php on POST_AUTOLOAD_DUMP

nicolas-grekas:
As reported by @stof in https://github.com/symfony/flex/pull/576#discussion_r357552402


fabpot:
Thank you @nicolas-grekas.

PR #607, opened 2020-02-06
Add option to not override autoload file

VincentLanglet:
The override made to the autoload file [here](https://github.com/symfony/flex/pull/576#discussion_r357552402) should be optional.
I made an option, with a default value to `true`.

It avoid the developper to re-edit the autoload file after a composer update/install/dump-autoload if he wants to use different php environnement.

The name of the option is debatable. And I don't know if it should be documented and where.

### DIFF
--- a/src/Flex.php
+++ b/src/Flex.php
@@ -668,6 +668,11 @@ class Flex implements PluginInterface, EventSubscriberInterface
 
     public function updateAutoloadFile()
     {
+        $strictPhpVersion = $this->composer->getPackage()->getExtra()['symfony']['strict-php-version'] ?? true;
+        if (!$strictPhpVersion) {
+            return;
+        }
+
         if (!$platform = $this->lock->get('php')['version'] ?? null) {
             return;
         }


nicolas-grekas:
I'm not convinced we should add this option. The workflow you describe looks broken to me. People that run PHP 7.1 for apps that were installed using PHP 7.4 will report parse errors as soon as a package uses 7.4-only syntax, generating a lot of wrong "support requests" to OSS maintainers, or other WTF moments when this happens.

VincentLanglet:
> I'm not convinced we should add this option. The workflow you describe looks broken to me. People that run PHP 7.1 for apps that were installed using PHP 7.4 will report parse errors as soon as a package uses 7.4-only syntax, generating a lot of wrong "support requests" to OSS maintainers, or other WTF moments when this happens.

The fact a package modify the autoload file looks weird to me. This file is generated by composer and should be only modified by composer. Even if our workflow is not the best, it worked before. And the update of a minor version of symfony/flex is now a BC-break for our project.

I'll be happy having another way to do this. But I can't say to my composer that I want :
- all the dependencies working for php >= 7.4
- all the dev-dependencies working for php >= 7.1

VincentLanglet:
@nicolas-grekas Should this PR be closed or is it possible to discuss about the non-mandatory modification of the autoload.php file ?

nicolas-grekas:
I'd recommend closing.

vnagara:
@VincentLanglet Dude, just comment the alert out from vendor/autoload.php. It works. 
For me it still wired to attach where you updated your project and not to constraints in written in the packages. It goes against the logic of packaging. It like creating a program on Debian10 with libs working on Debian8 but forbidding it to run as it was not created on Debian8 LOL. That the reason for putting requirements in a packaged/lib config. It's absurd. Because now composer remove a huge part of packaging system that allows create old-machine compatible programs using new computers/OS. It is like saying you are not allowed to create MS-DOS program on modern computer as it has different dependencies so use MS-DOS for that LOL.

xabbuh:
Composer offers an option to define a target PHP version if you run `composer update` on more recent PHP versions (and Flex will respect that version when updating the `autoload.php` file: https://getcomposer.org/doc/06-config.md#platform

vnagara:
@xabbuh Thanks. That is not obvious but should solve the problem.

PR #612, opened 2020-04-08
Enable bundles auto-discovery by default

SerheyDolgushev:
Bundles autodiscovery is now enabled for all the packages, but not only for `symfony-bundle` type packages. And it works in the same way as it worked before. 

PR description before @fabpot review:

> 1. Bundles autodiscovery is now enabled for all the packages, but not only for `symfony-bundle` type packages. And it works in the same way as it worked before. It is disabled only when there is `extra.symfony.bundles` configuration in the package (even if this configuration is an empty array).
> 2. New `extra.symfony.bundles` package configuration was introduced. It should contain the list of bundles and the environments where those bundles should be enabled. Example:
>     ```
>     "extra": {
>         "symfony": {
>             "bundles": {
>                 "VendorName\\BundleName1": ["all"],
>                 "VendorName\\BundleName2": ["dev", "test"]
>             }
>         }
>     }
>     ```
> 

Original PR description:
>I just do not see any reasons why `ezplatform-bundle` packages couldn't be enabled automatically after the installation (in the same way as `symfony-bundle` packages are).
>
>Tagging @andrerom here

### DIFF
--- a/src/Flex.php
+++ b/src/Flex.php
@@ -444,7 +444,7 @@ class Flex implements PluginInterface, EventSubscriberInterface
                         }
                         $value = strtolower($value[0]);
                         if (!\in_array($value, ['y', 'n', 'a', 'p'])) {
-                            throw new \InvalidArgumentException('Invalid choice');
+                            throw new \InvalidArgumentException('Invalid choice.');
                         }
 
                         return $value;
@@ -757,18 +757,23 @@ EOPHP
             }
 
             $noRecipe = !isset($manifests[$name]) || (isset($manifests[$name]['not_installable']) && $manifests[$name]['not_installable']);
-            if ($noRecipe && 'symfony-bundle' === $package->getType()) {
-                $manifest = [];
-                $bundle = new SymfonyBundle($this->composer, $package, $job);
+            if ($noRecipe) {
+                $bundles = [];
+
                 if (null === $devPackages) {
                     $devPackages = array_column($this->composer->getLocker()->getLockData()['packages-dev'], 'name');
                 }
                 $envs = \in_array($name, $devPackages) ? ['dev', 'test'] : ['all'];
-                foreach ($bundle->getClassNames() as $class) {
-                    $manifest['manifest']['bundles'][$class] = $envs;
+                $bundle = new SymfonyBundle($this->composer, $package, $job);
+                foreach ($bundle->getClassNames() as $bundleClass) {
+                    $bundles[$bundleClass] = $envs;
                 }
-                if ($manifest) {
-                    $manifest['origin'] = sprintf('%s:%s@auto-generated recipe', $name, $package->getPrettyVersion());
+
+                if ($bundles) {
+                    $manifest = [
+                        'origin' => sprintf('%s:%s@auto-generated recipe', $name, $package->getPrettyVersion()),
+                        'manifest' => ['bundles' => $bundles],
+                    ];
                     $recipes[$name] = new Recipe($package, $name, $job, $manifest);
                 }
             }
--- a/tests/FlexTest.php
+++ b/tests/FlexTest.php
@@ -12,6 +12,7 @@
 namespace Symfony\Flex\Tests;
 
 use Composer\Composer;
+use Composer\Config;
 use Composer\DependencyResolver\Operation\InstallOperation;
 use Composer\Factory;
 use Composer\Installer\PackageEvent;
@@ -37,10 +38,6 @@ class FlexTest extends TestCase
 {
     public function testPostInstall()
     {
-        $package = new Package('dummy/dummy', '1.0.0', '1.0.0');
-        $event = $this->getMockBuilder(PackageEvent::class)->disableOriginalConstructor()->getMock();
-        $event->expects($this->any())->method('getOperation')->willReturn(new InstallOperation($package));
-
         $data = [
             'manifests' => [
                 'dummy/dummy' => [
@@ -61,40 +58,14 @@ class FlexTest extends TestCase
             ],
         ];
 
-        $configurator = $this->getMockBuilder(Configurator::class)->disableOriginalConstructor()->getMock();
-        $configurator->expects($this->once())->method('install')->with($this->equalTo(new Recipe($package, 'dummy/dummy', 'install', $data['manifests']['dummy/dummy'], $data['locks']['dummy/dummy'])));
-
-        $downloader = $this->getMockBuilder(Downloader::class)->disableOriginalConstructor()->getMock();
-        $downloader->expects($this->once())->method('getRecipes')->willReturn($data);
-        $downloader->expects($this->once())->method('isEnabled')->willReturn(true);
-
         $io = new BufferIO('', OutputInterface::VERBOSITY_VERBOSE);
-        $locker = $this->getMockBuilder(Locker::class)->disableOriginalConstructor()->getMock();
-        $locker->expects($this->any())->method('getLockData')->willReturn(['content-hash' => 'random']);
+        $package = new Package('dummy/dummy', '1.0.0', '1.0.0');
+        $recipe = new Recipe($package, 'dummy/dummy', 'install', $data['manifests']['dummy/dummy'], $data['locks']['dummy/dummy']);
 
-        $package = $this->getMockBuilder(RootPackageInterface::class)->disableOriginalConstructor()->getMock();
-        $package->expects($this->any())->method('getExtra')->willReturn(['symfony' => ['allow-contrib' => true]]);
-
-        $lock = $this->getMockBuilder(Lock::class)->disableOriginalConstructor()->getMock();
-        $lock->expects($this->any())->method('has')->willReturn(false);
-
-        $flex = \Closure::bind(function () use ($configurator, $downloader, $io, $locker, $package, $lock) {
-            $flex = new Flex();
-            $flex->composer = new Composer();
-            $flex->composer->setLocker($locker);
-            $flex->composer->setPackage($package);
-            $flex->io = $io;
-            $flex->configurator = $configurator;
-            $flex->downloader = $downloader;
-            $flex->runningCommand = function () {
-            };
-            $flex->options = new Options(['config-dir' => 'config', 'var-dir' => 'var']);
-            $flex->lock = $lock;
-
-            return $flex;
-        }, null, Flex::class)->__invoke();
-        $flex->record($event);
-        $flex->install($this->getMockBuilder(Event::class)->disableOriginalConstructor()->getMock());
+        $rootPackage = $this->mockRootPackage(['symfony' => ['allow-contrib' => true]]);
+        $flex = $this->mockFlex($io, $rootPackage, $recipe, $data);
+        $flex->record($this->mockPackageEvent($package));
+        $flex->install($this->mockFlexEvent());
 
         $expected = [
             '',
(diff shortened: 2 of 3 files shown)


andrerom:
I was thinking of adding this some time ago as well, but I didn’t as I’m kind of expecting Fabien / Nicolas et. all to not want to add all custom bundle composer package types out there.

So I think we are lacking a convention, or something to say ezplatform-bundle is a superset of symfony-bundle


For instance composer should maybe allow type to be array (unlikely) or Flex should also look in composer.json keywords for ‘symfony-bundle’ (more likely), or something else.

SerheyDolgushev:
> I was thinking of adding this some time ago as well, but I didn’t as I’m kind of expecting Fabien / Nicolas et. all to not want to add all custom bundle composer package types out there.
> 
> So I think we are lacking a convention, or something to say ezplatform-bundle is a superset of symfony-bundle
> 
> For instance composer should maybe allow type to be array (unlikely) or Flex should also look in composer.json keywords for ‘symfony-bundle’ (more likely), or something else.

@nicolas-grekas what do you think?

alexander-schranz:
Was thinking this days to provide the same for `sulu-bundle`. Maybe we can find a convention. Or maybe use a `extra` field to provide the classes which should be added to the `bundles.php`.

```php
{
    "extra": {
        "symfony": {
             "flex-bundle-classes": ["My\Bundle\MyBundle"]
        }
    }
}
```

This would also don't need any hacks for sylius they just say which class should be added to the bundles.php.

SerheyDolgushev:
Talked to @nicolas-grekas on Symfony slack channel and he liked the idea suggested by @alexander-schranz. So this PR introduces `extra.symfony.bundle-classes` config.

When any package is installed, following bundles will be enabled automatically after the installation:
1. If package type is `symfony-bundle` all bundle classes will be extracted from it (current behavior).
2. If package has `extra.symfony.bundle-classes` configuration, all the bundle classes will be used from it (implemented in this PR).

Example usage:
```json
{
    "name": "contextualcode/ezplatform-search-binary-extractor",
    "type": "ezplatform-bundle",
    ...
    "extra": {
        "symfony": {
            "bundle-classes": ["ContextualCode\\EzPlatformSearchBinaryExtractorBundle\\EzPlatformSearchBinaryExtractorBundle"]
        }
    }
}
```

alexander-schranz:
@SerheyDolgushev I would also support in which environment the bundle should be registered:

E.g.:

```json
        "symfony": {
            "bundle-classes": {
                     "ContextualCode\\EzPlatformSearchBinaryExtractorBundle\\EzPlatformSearchBinaryExtractorBundle": {"all": true},
                     "SomeDebugBundle": {"dev": true},
           }
        }
```

or what do you think?

SerheyDolgushev:
@alexander-schranz I like this idea. @nicolas-grekas what do you think?

SerheyDolgushev:
Would change it to:
```
"symfony": {
    "bundle-classes": {
        "ContextualCode\\EzPlatformSearchBinaryExtractorBundle\\EzPlatformSearchBinaryExtractorBundle": ["all"],
        "SomeDebugBundle": ["dev"]
    }
}
```

SerheyDolgushev:
And I`m not sure if it worth adding this option for each bundle. I guess in most cases each package contains just one bundle?
So maybe `symfony. bundle-envs` would make more sense?

alexander-schranz:
@SerheyDolgushev we have a monorepo and not all our bundles will be registred in all environments of sulu so we need to define it per bundle if possible.

nicolas-grekas:
> `"SomeDebugBundle": ["dev"]`

Makes total sense yes, same as in `manifest.json` files.
I'd even suggest renaming the key to just "bundles" for consistency.

SerheyDolgushev:
So `symfony-bundle` packages are handled as they were before. But this PR provides support for:
```
    "extra": {
        "symfony": {
            "bundles": {
                "ContextualCode\\CoolBundle1": ["all"],
                "ContextualCode\\CoolBundle2": ["dev", "tests"]
            }
        }
    }
```

@alexander-schranz @nicolas-grekas

alexander-schranz:
Looks good for me 👍

src/Flex.php
```diff
@@ -757,18 +757,33 @@ public function fetchRecipes(array $operations): array
...
             $noRecipe = !isset($manifests[$name]) || (isset($manifests[$name]['not_installable']) && $manifests[$name]['not_installable']);
-            if ($noRecipe && 'symfony-bundle' === $package->getType()) {
-                $manifest = [];
-                $bundle = new SymfonyBundle($this->composer, $package, $job);
-                if (null === $devPackages) {
-                    $devPackages = array_column($this->composer->getLocker()->getLockData()['packages-dev'], 'name');
+            if ($noRecipe) {
+                $bundles = [];
+
+                if ('symfony-bundle' === $package->getType()) {
```
andrerom, line 763: @nicolas-grekas What if we also change to `if ('symfony-bundle' === $package->getType() || isset($extra['symfony'])) {` when no explicit bundles are defined? So meaning as a fallback for the explicit bundles logic

Meaning any package with symfony in extra section is assumed to be a Symfony bundle as well for the auto detect logic?

andrerom: Hmm, maybe not, symfony section is used in many packages that are not bundles right?

fabpot:
Another idea (I've not tried, so it might not work well): we already have many heuristics to find the bundle class, what about looking for a bundle class in **all* packages instead of limiting it to Symfony bundles. That would allow packages that are a lib and a bundle to be auto-detected as well.

SerheyDolgushev:
> Another idea (I've not tried, so it might not work well): we already have many heuristics to find the bundle class, what about looking for a bundle class in *_all_ packages instead of limiting it to Symfony bundles. That would allow packages that are a lib and a bundle to be auto-detected as well.

So seems like you are suggesting to enable all the bundles by default? Handle all the bundles in a way as `symfony-bundle` are handled now? I guess it would be a brilliant change, as it will not require any additional changes for existing bundles, and they will be automatically enabled after the `composer require`.

But also, I think, it make sense to add `extra.symfony.bundles`, just to being able to enable specific bundles in a specific environment.

So the final behavior would be:
1. Extract bundles from all the packages, but not just `symfony-bundle`.
2. If the current package is listed under the composers' dev packages, all its bundles will be enabled for `dev` and `test ` environments. Otherwise, they will be enabled for `all ` environments (current behavior).
3. Extract the list of bundles from `extra.symfony.bundles` and override specified environments for them. It also will work if in some cases the bundles were not extracted during #1.

So this PR will implement the following:
1. Extracted bundles will be enabled automatically for all packages, but not just `symfony-bundle`.
2. It will use `extra.symfony.bundles` to change the environments for extracted bundles and for specifying additional bundles that are not extracted automatically by some reason.

Gentlemans @nicolas-grekas @andrerom @alexander-schranz what do you think?

nicolas-grekas:
Sounds good to me, with just one change: if `extra.symfony.bundles` is found, autodiscovery should be disabled. This will provide a way to skip some bundles when appropriate.

SerheyDolgushev:
Just to summarize this PR:
1. Bundles autodiscovery is now enabled for all the packages, but not only for `symfony-bundle` type packages. And it works in the same way as it worked before. It is disabled only when there is `extra.symfony.bundles` configuration in the package.
2. New `extra.symfony.bundles` package configuration was introduced. It should contain the list of bundles and the environments where those bundles should be enabled. Example:
    ```
    "extra": {
        "symfony": {
            "bundles": {
                "VendorName\\BundleName1": ["all"],
                "VendorName\\BundleName2": ["dev", "test"]
            }
        }
    }
    ```

And that is basically all. I tested the latest changes on a few bundles and it worked as expected.

src/Flex.php
```diff
@@ -757,18 +757,33 @@ public function fetchRecipes(array $operations): array
...
+                    $extraBundles = (array) $extra['symfony']['bundles'];
+                    foreach ($extraBundles as $bundleClass => $envs) {
+                        $bundles[$bundleClass] = (array) $envs;
+                    }
                 }
-                $envs = \in_array($name, $devPackages) ? ['dev', 'test'] : ['all'];
-                foreach ($bundle->getClassNames() as $class) {
-                    $manifest['manifest']['bundles'][$class] = $envs;
+
+                if (0 === \count($bundles)) {
```
nicolas-grekas, line 771: this should be the "else" case from the previous "if": it's legit to say there are no bundles in this lib while autodiscovery might find some otherwise, isn't it?

src/Flex.php
```diff
@@ -757,18 +757,33 @@ public function fetchRecipes(array $operations): array
...
+                    $envs = \in_array($name, $devPackages) ? ['dev', 'test'] : ['all'];
+                    $bundle = new SymfonyBundle($this->composer, $package, $job);
+                    foreach ($bundle->getClassNames() as $bundleClass) {
+                        $bundles[$bundleClass] = $envs;
+                    }
                 }
-                if ($manifest) {
-                    $manifest['origin'] = sprintf('%s:%s@auto-generated recipe', $name, $package->getPrettyVersion());
+
+                if (\count($bundles)) {
```
nicolas-grekas, line 782: `if ($bundles) {` (count is useless)

src/Flex.php
SerheyDolgushev: Currently empty `symfony.bundles` should not disable the autodiscovery. With the change you propose, it will disable it. 
So it is more the question of how we want it to behave? :)

nicolas-grekas: It should skip then - we should never remove power from users.

SerheyDolgushev: But being able to disable the autodiscovery with setting an empty `symfony.bundles` might be a good option!

SerheyDolgushev: https://github.com/symfony/flex/pull/612/commits/b5aebef599fd8d0eb868a9de27d9fdc8ef1b3929

src/Flex.php
SerheyDolgushev: https://github.com/symfony/flex/pull/612/commits/390179eec930c4a94a41755c5a5f4868b544cdea

src/Flex.php
nicolas-grekas: Which is what I'm saying :)

src/Flex.php
```diff
@@ -757,18 +757,31 @@ public function fetchRecipes(array $operations): array
...
-                }
-                $envs = \in_array($name, $devPackages) ? ['dev', 'test'] : ['all'];
-                foreach ($bundle->getClassNames() as $class) {
-                    $manifest['manifest']['bundles'][$class] = $envs;
+            if ($noRecipe) {
+                $bundles = [];
+
+                $extra = $package->getExtra();
+                if (isset($extra['symfony']['bundles'])) {
+                    $extraBundles = (array) $extra['symfony']['bundles'];
```
SerheyDolgushev, line 765: @nicolas-grekas should we check if bundles is array here instead of casting it?

SerheyDolgushev: Maybe
```
$extraBundles = is_array($extra['symfony']['bundles']) ? $extra['symfony']['bundles'] : [];
```
?

SerheyDolgushev:
I can't mention I was listening to Zaz and Charles Aznavour while I was implementing @nicolas-grekas review.

src/Flex.php
nicolas-grekas: we could throw an exception if it's not an array
same in the foreach just after btw
and same if envs are not strings maybe also to be extra careful?

SerheyDolgushev: I like throwing `InvalidArgumentException` exception. With envs it is simpler, as they could be a string. In that case `env` just will be converted to `[env]`. Which should be fine?

nicolas-grekas: better be strict if we validate something imho
we should validate the content of the array also

SerheyDolgushev: Ok, so `InvalidArgumentException` in both cases?

SerheyDolgushev: Or better https://www.php.net/manual/en/class.typeerror.php?

nicolas-grekas: type errors are for internal issues, aka badly written code
this is input validation

SerheyDolgushev: Array checks have been added for extra bundles and their envs.

src/Flex.php
```diff
@@ -757,18 +757,42 @@ public function fetchRecipes(array $operations): array
...
-                foreach ($bundle->getClassNames() as $class) {
-                    $manifest['manifest']['bundles'][$class] = $envs;
+            if ($noRecipe) {
+                $bundles = [];
+
+                $extra = $package->getExtra();
+                if (isset($extra['symfony']['bundles'])) {
+                    $extraBundles = $extra['symfony']['bundles'];
+                    if (!\is_array($extraBundles)) {
+                        throw new \InvalidArgumentException('"symfony.bundles" should be an array.');
```
nicolas-grekas, line 767: `Entry "bundles" under "extra.symfony" in the composer.json of package "%s" must be an array, %s given.`

The following new exception messages should give similar added context.

SerheyDolgushev: Done.

src/Flex.php
```diff
@@ -757,18 +757,42 @@ public function fetchRecipes(array $operations): array
...
-                foreach ($bundle->getClassNames() as $class) {
-                    $manifest['manifest']['bundles'][$class] = $envs;
+            if ($noRecipe) {
+                $bundles = [];
+
+                $extra = $package->getExtra();
+                if (isset($extra['symfony']['bundles'])) {
+                    $extraBundles = $extra['symfony']['bundles'];
+                    if (!\is_array($extraBundles)) {
+                        throw new \InvalidArgumentException(sprintf('Entry "bundles" under "extra.symfony" in the composer.json of package "%s" must be an array, "%s" given.', $package->getName(), \gettype($extraBundles)));
```
nicolas-grekas, line 767: s/$package->getName()/$name

src/Flex.php
```diff
@@ -757,18 +757,42 @@ public function fetchRecipes(array $operations): array
...
+                    }
+                    foreach ($extraBundles as $bundleClass => $envs) {
+                        if (!\is_array($envs)) {
+                            throw new \InvalidArgumentException($bundleClass.' environments should be an array.');
+                        }
+                        foreach ($envs as $env) {
+                            if (!\is_string($env)) {
+                                throw new \InvalidArgumentException('Each environment should be a string.');
```
nicolas-grekas, line 775: `Bundle "%s" under "extra.symfony.bundles" should list Symfony envs as strings, %s found in the composer.json of package "%s".`

src/Flex.php
```diff
@@ -757,18 +757,42 @@ public function fetchRecipes(array $operations): array
...
+
+                $extra = $package->getExtra();
+                if (isset($extra['symfony']['bundles'])) {
+                    $extraBundles = $extra['symfony']['bundles'];
+                    if (!\is_array($extraBundles)) {
+                        throw new \InvalidArgumentException(sprintf('Entry "bundles" under "extra.symfony" in the composer.json of package "%s" must be an array, "%s" given.', $package->getName(), \gettype($extraBundles)));
+                    }
+                    foreach ($extraBundles as $bundleClass => $envs) {
+                        if (!\is_array($envs)) {
+                            throw new \InvalidArgumentException($bundleClass.' environments should be an array.');
```
nicolas-grekas, line 771: `Invalid envs for bundle "%s" under "extra.symfony.bundles": array expected, %s given in the composer.json of package "%s".`

fabpot:
Can we add some tests? iIRC, we already have quite a few tests on this feature, it might be good to add more to cover the new use cases.

SerheyDolgushev:
@fabpot Please check https://github.com/symfony/flex/blob/bfcc097fb5a7e446da3bd6a8a7440993cfcacbca/tests/FlexExtraBundlesTest.php
Also, I made minor fixes for some existing tests, as they were failing on my local.

fabpot:
I see that you've added a new test file, I was more expecting new test cases on existing tests.

SerheyDolgushev:
@fabpot Yes, I decided to use a separate file for these tests. In theory, they can be "merged" into https://github.com/symfony/flex/blob/master/tests/FlexTest.php, if it is critical.

fabpot:
I would prefer to keep all tests in the same file, yes.

SerheyDolgushev:
@fabpot done in https://github.com/symfony/flex/pull/612/commits/c7a3fa88d55d8bb229b06e256cc46f3c1b185a0f

natepage:
@fabpot @nicolas-grekas 

Hi there, I don't know if it is the right place to discuss this and I'm happy to create a specific issue if required but I think it is related.

Once this PR gets merged, are the validation rules to add recipes to `symfony/recipes-contrib` gonna be updated to accept recipes for packages that are not `symfony-bundle`?

The current PR enables autodiscovery for bundle classes and that's amazing already but if a package has some config to copy, env to create, etc. It still needs to create a recipe for all that to be automated.

Yesterday I've created a PR to `recipes-contrib` for a few packages we're working on, those packages follow the same philosophy which is to implement the core features to be framework agnostic and then create bridges for them:

```
package/
-- src/
---- Bridge/
------ Laravel/
------ Symfony/
-------- PackageBundle.php
---- ... Core files ...     
```

This allows us to implement feature solving business cases and reuse them across different projects/platforms and make sure it works as expected on each of them centralised in one place.

So yes the package itself is a `library` but it also provides a Symfony bundle and having a recipe for it would make DX amazing!

fabpot:
I think that the `extra.symfony.bundles` feature should be removed from this PR. It is orthogonal to the goal and duplicates what recipes are for, so I'm in favor of adding it.

andrerom:
@fabpot So you only want following change then to be clear? (+test adjustments)

```diff
- if ($noRecipe && 'symfony-bundle' === $package->getType()) {
+ if ($noRecipe) {
```

fabpot:
This PR should really just be about bundle auto-discovery for all packages.

SerheyDolgushev:
@fabpot requested changes have been applied. Please review them. I`ll squash commits after your confirmation.

fabpot:
Thank you @SerheyDolgushev.

alexander-schranz:
Its really sad to hear that the we can not configure it from the `composer.json`. It would be great to have everything in the package version control, but I think you decided to go with recipe repository to have all recipes under your control, I can understand that decision but from package maintainer it would be better to have a way to provide the recipe config with the package code directly.

nicolas-grekas:
Recipes are like a Linux distribution. Putting them in the hand of package authors would wrongly give them a responsibility they should not have. Defining *how* a package is used in always an opinionated decision and doesn't belong to lib authors.

dunglas:
This is subject to debate. Even for OS packages, the trend (Snap, Docker, App Store, Google Play...) is to let editors packaging software as they want. It would help a lot of packages to include their recipes in their repositories (I have Sylius plugins in mind for instance). It will also make the versioning easier. Such recipes could be considered as « contrib » recipes, so the end user has the choice to execute it or not.

alexander-schranz:
@nicolas-grekas currently it seems not yet possible to submit recipes for other types: https://github.com/symfony/recipes-contrib/issues/984 as the CI seems still check for the composer type when submitting recipes.

fabpot:
@alexander-schranz Fixed now

alexander-schranz:
@fabpot Thank you!

PR #614, opened 2020-04-15
Fix version comparison on lock files

curry684:
As previously `7.2.11` was a lower version than `7.2.2`.


stof:
This is wrong. \PHP_VERSION_ID` is **not** `7.2.11` but `70211` as an integer

curry684:
Oh f* me, you're right.

Issue #593, opened 2019-12-24
[RFC] API Endpoint to fetch historic recipe by sha? (for proper removing & better update)


weaverryan:
Hi!

This can probably only be answered by Fabien. Is it feasible for the Flex API to support fetching an *old* version of a recipe? For example, suppose we're talking about `twig/extensions`. Would it be possible to ask for the `c986e32` sha of the 1.0 version of the recipe (which would be one commit *before* HEAD - https://github.com/symfony/recipes/commits/master/twig/extensions/1.0 )?

*If* we could do this, it would allow us to do two more things slightly more accurately:

1) On package removal, we could remove using the same version of a recipe as the user originally had installed. Minor - but better.

2) On package update, we could remove using the original recipe first and THEN upgrade to the latest. That would fix the problem where a file/parameter that was added in the OLD recipe is NOT added in the new recipe. Currently, that file/parameters is NOT removed. But if we were able to remove the original recipe, it WOULD remove that old file/parameter.

So, is that possible to add to the API? 

Thank you :).

nicolas-grekas:
Just fetching recipes by tree-hash wouldn't be enough: we'd need to resolve symlinks also. But I don't think symlinks from a tree-hash can resolve to the target corresponding to the hash. Because a tree hash doesn't snapshot the target, we cannot know the state of the target.

We need to resolve this before.

One way would be to forbid symlinks - and rewrite the history of this repo to resolve them all also.
Alternatively, we could have a bot create a shadow git history with resolved symlinks, and the tree-hash would map to this shadow repo, instead of the main

maxhelias:
In fact, I think we can do it without changing the endpoint.
As we save the files added by the recipe in the .lock, we should be able to delete them and install those of the new recipe. And for the other cases, the hash is sufficient to retrieve the configuration of the manifest.

I may be wrong or forget some cases ?

nicolas-grekas:
#803 should provide all the answers to this question \o/

PR #615, opened 2020-04-21
Remove files when updating a recipe

maxhelias:
Fix issue #593 and https://twitter.com/nicolasgrekas/status/1252292768296308736

@nicolas-grekas tell me if it's ok as it's a little "hack"

### DIFF
--- a/src/Configurator/CopyFromRecipeConfigurator.php
+++ b/src/Configurator/CopyFromRecipeConfigurator.php
@@ -24,7 +24,32 @@ class CopyFromRecipeConfigurator extends AbstractConfigurator
         $this->write('Copying files from recipe');
         $options = array_merge($this->options->toArray(), $options);
 
-        $lock->add($recipe->getName(), ['files' => $this->copyFiles($config, $recipe->getFiles(), $options)]);
+        $files = $this->copyFiles($config, $recipe->getFiles(), $options);
+        $lock->add($recipe->getName(), ['files' => $files]);
+
+        // Remove extra file only on force
+        if (!($options['force'] ?? false)) {
+            return;
+        }
+
+        // Recipe not installed yet
+        if (null === $recipeLock = $options['beforeState']) {
+            return;
+        }
+
+        $lockFiles = $recipeLock['files'] ?? null;
+        if (null === $lockFiles) {
+            return;
+        }
+
+        $lockFiles = array_flip($lockFiles);
+        foreach ($files as $file) {
+            if (isset($lockFiles[$file])) {
+                unset($lockFiles[$file]);
+            }
+        }
+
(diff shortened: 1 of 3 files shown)


nicolas-grekas:
Thanks for starting this!
I had another idea that I tried to described on Twitter.
I think the current approach won't work when a file is referenced by several recipes (eg phpunit.xml.dist). We should remove the files only when they are not referenced by *any* recipe (like the "remove" operation does).

So, actually, maybe the correct approach is to *remove* recipes just before applying them again. This would mean that the current check that asks confirmation before a file is overridden should be moved before the removal happens.

I'm writing this without having opened the code, please check if it makes any sense :)

weaverryan:
> I think the current approach won't work when a file is referenced by several recipes (eg phpunit.xml.dist).

That's a really good point

> So, actually, maybe the correct approach is to remove recipes just before applying them again

I think this would be great! But to do this, wouldn't we need the ability to "fetch the original recipe" so that we can remove it? If we ran an unconfigure option on the latest version of the recipe, it wouldn't delete files that have been removed from the recipe. Right?

nicolas-grekas:
> wouldn't we need the ability to "fetch the original recipe" so that we can remove it?

For files, we don't need to, they are already listed in symfony.lock
For other things (.env, .gitignore, etc) yes - we could put the manifest in symfony.lock also when installing a recipe, that would solve this aspect also.
Note that to me, files account for 95% of the problem we want to solve, and we have everything to solve these 95% now.

weaverryan:
> For files, we don't need to, they are already listed in symfony.lock

Are you saying that we should, sort of, use the `files` from `symfony.lock` to create a `Recipe` object and then unconfigure that? That sounds reasonable to me - though there may be some work to create a `Recipe` object from this other, "incomplete" source.

> For other things (.env, .gitignore, etc) yes - we could put the manifest in symfony.lock when installing a recipe, that would solve this aspect also.

That's a really good idea :). We could also show more details when viewing a recipe with the `composer recipe` command... without needing API changes.

nicolas-grekas:
> create a Recipe object from this other, "incomplete" source

that, or we implement a dedicated file deletion logic that focuses on the side effect: refcounting files and removing those that are referenced *only* by recipes we're about to resync.

weaverryan:
Ping @maxhelias! What do you think about the proposals here? Do you have time to update the PR?

Thanks!

maxhelias:
I think a dedicated file deletion logic should be a good idea but i don't know if i have time to do it, i'll try within a week. So don't hesitate if you want to fast-forward this

weaverryan:
Friendly ping @maxhelias if you have some time :)

maxhelias:
Hi @weaverryan, i'm not sure to have time to finish that

maxhelias:
close in favor of #667

PR #616, opened 2020-04-25
Fix typo in DumpEnvCommand

Shine-neko:


nicolas-grekas:
Thank you, that's not a type but supported syntax for the console.

PR #617, opened 2020-04-25
Add support for composer 2

### LINKED ISSUE #661: Composer recipes Suggests Update from 2.2 to 1.2
If I run composer recipes, I get a suggestion to update from 2.2 to 1.2...

```
composer recipes doctrine/doctrine-migrations-bundle                   
name             : doctrine/doctrine-migrations-bundle
version          : 2.2
status           : update available
installed recipe : https://github.com/symfony/recipes/tree/4bea6a9/doctrine/doctrine-migrations-bundle/2.2
latest recipe    : https://github.com/symfony/recipes/tree/master/doctrine/doctrine-migrations-bundle/1.2
recipe history   : https://github.com/symfony/recipes/commits/master/doctrine/doctrine-migrations-bundle?since=2020-03-28T09:40:26+00:00Z
files            : 

├──config
│  └──packages
│     └──doctrine_migrations.yaml
└──migrations
   └──.gitignore

Update this recipe by running:
composer recipes:install doctrine/doctrine-migrations-bundle --force -v
```

jderusse:
This PR add compatibility with `composer-plugin-api: ^2.0` and fixes #661

Maintaining compatibility with both 1.x and 2.0 was too complex.
I think we need to release a version 2.0 of flex compatible with composer 2.

- remove `ParallelDownloader` (no more needed with composer 2)
- replace `TruncateRepository` by eventListener on `PluginEvents::PRE_POOL_CREATE`

note: I didn't change the way commands are added to the application, current plugin's contract expect a class name, while we currently inject command instances (with config and flex itself injected inside) in the application.


src/Command/RecipesCommand.php
```diff
@@ -354,8 +355,9 @@ private function findRecipeCommitDataFromTreeRef(string $package, string $repo,
 
     private function requestGitHubApi(string $path)
     {
-        $contents = $this->downloader->getContents('api.github.com', $path, false);
+        /** @var Response $response */
```
stof, line 358: Do we need that ? Isn't it already covered by the composer phpdoc ?

jderusse: Reverted, not necessary.
Not the method is not documented.

stof: if the phpdoc is missing in composer, I suggest updating composer instead :smile:

maxhelias:
@jderusse for the command instances, I think we can declare the flex class itself as the command provider. This would allow us to keep the injection of the configuration and flex

jderusse:
That's a good idea @maxhelias , I totally missed that feature I implemented ^^

Unfortunaly, it does not works for some command (require, install, update). Composer prevent overriding internal methods (https://github.com/composer/composer/blob/master/src/Composer/Console/Application.php#L180)

One solution could be to declare the command with name "symfony:require" and alias "require" (aliases are not (yet) checked, and will override the composer's command)
WDYT @nicolas-grekas 

side note: Composer improved a lots dependencies resolution. TruncatedPackage is maybe not needed anymore? Do you have good project to benchmark?

nicolas-grekas:
> Unfortunaly, it does not works for some command (require, install, update)

yep, that's why we used this technic. Then, better have the same strategy for all commands.

> TruncatedPackage is maybe not needed anymore? Do you have good project to benchmark?

it's not only for performance, but to enforce the version of nested dependencies. So yes, it's needed.

k0d3r1s:
hey, what's happening with this one? any eta?

stof:
Well, as I discussed in the issue, enforcing constraints should be doable with a listener on the pool building in Composer 2. This would remove the need for TruncatedPackageRepository.

And for commands, adding support for aliases in `require`, `update` and `remove` should be doable by manipulating the arguments using the symmfony/console `console.command` event instead.

jderusse:
> Well, as I discussed in the issue, enforcing constraints should be doable with a listener on the pool building in Composer 2. This would remove the need for TruncatedPackageRepository.

The `TruncatedPackageRepository`is already removed, and replaced by a listener on `PluginEvents::PRE_POOL_CREATE`

> And for commands, adding support for aliases in require, update and remove should be doable by manipulating the arguments using the symmfony/console console.command event instead

Hmm composer does not inject the EventDispatcher in `BaseApplication`,  (bu can be doable in the backtrace hack). And command override is not only about alias (ie. adding option `unpack` in `require`). I afraid we still have to replace original's commandes and keep the hack.


PR rebased

vasilvestre:
This is needed for PHP 8 on Symfony Flex (and so in all other versions of Symfony). https://github.com/symfony/flex/pull/624

vasilvestre:
@jderusse could you please rebase ?

fabpot:
Thank you @jderusse.

PR #621, opened 2020-05-12
Fix: Additional whitespace

OskarStark:
Otherwise it will result in:
Screenshot 2020-05-12 14 06 24

### DIFF
--- a/src/Flex.php
+++ b/src/Flex.php
@@ -405,8 +405,8 @@ class Flex implements PluginInterface, EventSubscriberInterface
         $this->operations = [];     // Reset the operation after getting recipes
 
         if (2 === $this->displayThanksReminder) {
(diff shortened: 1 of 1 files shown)


nicolas-grekas:
On other terminals, a space is eaten, leading to no space between :heart: and "by", that's why there is this double space...

OskarStark:
Hmpf, I thought about sth like this 🤔

OskarStark:
What about adding a "special"/"magic" whitespace instead a regular one?

nicolas-grekas:
How would that work?

OskarStark:
> What about adding a "special"/"magic" whitespace instead a regular one?

TBH, I don't know, I thought you know something special 😄 

Lets close here

PR #624, opened 2020-05-20
PHP 8 compatibility

vasilvestre:


nicolas-grekas:
Thank you for starting!
I took over in #625 
We'll do it :)

PR #625, opened 2020-05-21
Bump to php >= 7.1

nicolas-grekas:


fabpot:
Thank you @nicolas-grekas.

PR #628, opened 2020-05-30
avoid >100% progress

### LINKED ISSUE #468: Percentage of downloaded packages goes beyond 100%
Sometimes (I haven't been able to figure out when), the percentage of downloaded packages rapidly goes way beyond 100%:

image

There are no functional issues, everything works as expected, expect for the cosmetic issue. OS is Ubuntu 18.10.

IonBazan:
Fixes #468 
This change skips tracking the progress of downloads where `Content-Length` header is missing. Curl returns `-1` which resulted in wrong progress calculation (total file size was not added to the total bytes to download).

As soon as such file is downloaded, the progress is updated. This means that downloading one big file might cause progress to stuck until the download is done.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 /vendor/
 /build/
+.phpunit.result.cache
 .php_cs.cache
 composer.lock
--- a/.php_cs.dist
+++ b/.php_cs.dist
@@ -7,8 +7,9 @@ return PhpCsFixer\Config::create()
     ->setRules(array(
         '@Symfony' => true,
         '@Symfony:risky' => true,
+        'fopen_flags' => false,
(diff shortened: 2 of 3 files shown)


nicolas-grekas:
Thank you @IonBazan.

stof:
is this something that needs to be applied in composer 2 as well ? I know that they reused part of the curl implementation of Flex, but I don't know whether this bug happens in the code they copied or in the code they did not use.

nicolas-grekas:
@stof can you please have a look?
Or maybe @Seldaek knows?

Seldaek:
We dropped progress support entirely so should be fine but thanks for asking.

PR #632, opened 2020-06-14
Fix compatibility with semver 3

jderusse:
The `EmptyConstraint` has been renamed into `MatchAllConstraint` in semver 3.0 (minimum requirement for composer 2)


nicolas-grekas:
Fixed in #631 already, thanks :)

jderusse:
already fixed by #631

jderusse:
damn.. even on commenting PR, @nicolas-grekas is faster than me!

Issue #629, opened 2020-06-05
Improve discoverability of recipe changes


javiereguiluz:
Related to this -> https://github.com/symfony/symfony-docs/issues/13753. A user complained that it wasn't easy to discover that there were some very important changes from Symfony 5.0 to 5.1 in some recipes (mostly about the `Kernel.php` file).

It's true that developers know about `composer sync-recipes`, etc. but maybe we should be more proactive when upgrading packages to remind devs that there might be super important changes in the recipes of their dependencies. I don't know exactly how to solve this issue ... but I think this is an issue.

Thanks!

nicolas-grekas:
No changes in recipes are super important... It's never a must to sync recipes when upgrading to 5.1.

javiereguiluz:
I disagree a bit:

* In 5.0, `Kernel` loads YAML, XML and PHP routes and config ... in 5.1, it only loads YAML.
* In 5.0, the front controller, console script and tests use a config bootstrap file. In 5.1 that file doesn't even exist.
* 5.1 changes a bit how the `APP_SECRET` parameter is defined.

These changes are not important in the sense that your app breaks. Your app will keep working WITHOUT making these changes. The problem is that you are "left behind" with an "outdated way of doing things". For example, you read Symfony 5.1 docs and suddenly everything is different than your 5.1 app (because you didn't apply these recipe updates).

pbowyer:
I didn't know of `composer sync-recipes` until today, when @javiereguiluz talked about it. It has been weird when after upgrading through a few minor releases I find the docs no longer match my files.

Or when I start a new project, and realise they are different. My solution has been to create a new project every now and then and _manually diff_ the files and then update my existing project.... so yeah, I'm in favour of Flex telling me when files it would set up are different to the versions I have (minus my edits).

Adding "Run `composer sync-recipes`" to the UPGRADING files would help (unless it's been there all along and I've missed it 15+ times)

Amunak:
It would be nice if `composer sync-recipes` (or a similar command) could output the changes and do the diffing for us.

Or we could do it like package managers on Linux: create `.flexnew` files with the new contents so that the developer can diff it themselves. For files that are in the original state you could update them immediately (the developer still needs to commit so it's an affirmative action anyway) - maybe, when possible, do this even to additions to changed config files, but that could be hard. Trivially you could just try to apply a diff, and if it does apply, great. If not, create the `.flexnew` file.

[...]

PR #633, opened 2020-06-14
Improve discoverability of 'composer recipes'

maxhelias:
Discovering the `composer recipes` command during composer install/update

Fix issue #629 and move forward #518 - 8)

Capture d’écran 2020-06-15 à 17 05 28

### DIFF
--- a/src/Flex.php
+++ b/src/Flex.php
@@ -436,11 +436,17 @@ class Flex implements PluginInterface, EventSubscriberInterface
             $this->io->writeError('');
             $this->io->writeError('What about running <comment>composer global require symfony/thanks && composer thanks</> now?');
             $this->io->writeError(sprintf('This will spread some %s by sending a %s to the GitHub repositories of your fellow package maintainers.', $love, $star));
-            $this->io->writeError('');
         }
+
+        $this->io->writeError('');
 
         if (!$recipes) {
             $this->lock->write();
+
+            if ($this->downloader->isEnabled()) {
+                $this->io->writeError('Run <comment>composer recipes</> at any time to see the status of your Symfony recipes.');
+                $this->io->writeError('');
+            }
 
             return;
         }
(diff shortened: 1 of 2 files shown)


src/Flex.php
```diff
@@ -436,6 +436,11 @@ public function install(Event $event = null)
             $this->io->writeError('');
             $this->io->writeError('What about running <comment>composer global require symfony/thanks && composer thanks</> now?');
             $this->io->writeError(sprintf('This will spread some %s by sending a %s to the GitHub repositories of your fellow package maintainers.', $love, $star));
+        }
+
+        $this->io->writeError('');
+        if ($this->downloader->isEnabled()) {
+            $this->io->writeError('Run <comment>composer recipes</> at any time to see a status of your Symfony recipes.');
```
weaverryan, line 443:
```suggestion
            $this->io->writeError('Run <comment>composer recipes</> at any time to see the status of your Symfony recipes.');
```

src/Flex.php
```diff
@@ -436,6 +436,11 @@ public function install(Event $event = null)
             $this->io->writeError('');
             $this->io->writeError('What about running <comment>composer global require symfony/thanks && composer thanks</> now?');
             $this->io->writeError(sprintf('This will spread some %s by sending a %s to the GitHub repositories of your fellow package maintainers.', $love, $star));
+        }
+
+        $this->io->writeError('');
+        if ($this->downloader->isEnabled()) {
```
nicolas-grekas, line 442: When is this `true`? Displaying this on every single `composer u` or `composer i` might be too much, don't you think?

maxhelias: This is the way I found to disable the message for global update / install.
Perhaps, personally it does not shock me that it is reminded on each install / update. Otherwise we can display it in a post-installation message

maxhelias: WDYT @nicolas-grekas ?

nicolas-grekas:
Thank you @maxhelias.

PR #635, opened 2020-06-15
Set --no-suggest option to true only on Composer 1

### LINKED ISSUE #634: --no-suggest option is deprecated in Composer 2
Flex sets the `--no-suggest` option to `true` in https://github.com/symfony/flex/blob/master/src/Command/UpdateCommand.php#L35.

However, this option is deprecated in Composer 2, resulting in `You are using the deprecated option "--no-suggest". It has no effect and will break in Composer 3.` notice being printed in the output.

emodric:
Fixes #634


fabpot:
Thank you @emodric.

PR #637, opened 2020-06-16
add typehint

abdounikarim:
#636


maxhelias:
For compatibility reasons with composer 1 & 2, we cannot type this parameter.
With composer 1, we have a `Composer\Util\RemoteFilesystem` and on composer 2 we have a `Composer\Util\HttpDownloader`.

I think it's your Flex installed in global is not up to date. Run `composer global update` and it should be good.

PR #640, opened 2020-06-16
Fix downloading recipes in parallel

nicolas-grekas:


maxhelias:
Sorry, I haven't studied composer 2 well enough yet.
But I see that `wait` method can take a progress bar, that would mean we could also improve the progression in Flex?

nicolas-grekas:
I don't know. Also, I'm not sure a progress bar when downloading the recipes would provide any UX improvement.

PR #641, opened 2020-06-16
Add a way to disallow displaying the message for thanks

gmponos:


nicolas-grekas:
`extra.thanks` already has a meaning, see e.g. https://github.com/symfony/webpack-encore-bundle/blob/master/composer.json

Overall, I don't think this config option is worth it.

PR #643, opened 2020-06-16
Handle branch alias in filtered packages

### LINKED ISSUE #642: Can't upgrade or create-project with the master branch of skeletons
```sh
composer create-project symfony/skeleton test-project -s dev`
```

or

```sh
git clone git@github.com:symfony/skeleton.git test-project
cd test-project
composer update
```

both lead to the same failure:
```
Your requirements could not be resolved to an installable set of packages.

  Problem 1
    - symfony/skeleton is present at version 5.2.x-dev and cannot be modified by Composer
    - symfony/yaml 5.2.x-dev is an alias of symfony/skeleton 5.2.x-dev and thus requires it to be installed too.
    - Root composer.json requires symfony/yaml * -> satisfiable by symfony/yaml[5.2.x-dev (alias of dev-master)].
```

I've tracked down this to changes done by the new PackageFilter class.

A diff between the old (but still present) `Cache` and `PackageFilter` shows that the logic changed a lot between both. I suppose the issue lies somewhere there.

/cc @jderusse in case you'd like to have a look.

jderusse:
Fixes #642

### DIFF
--- a/src/PackageFilter.php
+++ b/src/PackageFilter.php
@@ -22,6 +22,7 @@ use Composer\Semver\VersionParser;
 class PackageFilter
 {
     private $versions;
+    private $versionParser;
     private $symfonyRequire;
     private $symfonyConstraints;
     private $downloader;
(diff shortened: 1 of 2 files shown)


src/PackageFilter.php
```diff
@@ -22,13 +22,15 @@
...
     private $symfonyRequire;
     private $symfonyConstraints;
     private $downloader;
     private $io;
 
     public function __construct(IOInterface $io, string $symfonyRequire, Downloader $downloader)
     {
+        $this->versionParser = new VersionParser();
         $this->symfonyRequire = $symfonyRequire;
         $this->symfonyConstraints = (new VersionParser())->parseConstraints($symfonyRequire);
```
nicolas-grekas, line 35: this should reuse the `$this->versionParser` var

nicolas-grekas:
@jderusse can you have a second look please? I just pushed a cleanup, the logic should remain the same.

jderusse:
LGTM. Thanks for the refactor

nicolas-grekas:
Thank you @jderusse.

Issue #645, opened 2020-06-30
Revisit the way packs work - always unpack them


nicolas-grekas:
Packs create issues like https://github.com/symfony/orm-pack/pull/25

For sure, we don't want to maintain versions of packs each time a dependent is bumped.
The current solution is like: use packs until it breaks for you, then unpack.

Instead, I'd like to propose the following:
- keep using `*` as the version constraint for dependents in packs,
- auto-unpack when installing a pack,
- auto-replace the `*` by the current version - compute this "current version" by borrowing what composer does when running compose require without a version constraint (when the package is already found in the main composer.json, skip this),
- additionally, also unpack require-dev,
- (skip "conflict" and/or "replace", that's too much work, better consider them unsupported in packs, until one contributes the logic at least?).

Related to https://github.com/symfony/orm-pack/issues/26, https://github.com/symfony/flex/issues/370, https://github.com/symfony/flex/issues/371, https://github.com/symfony/flex/issues/563

stof:
> For sure, we don't want to maintain versions of packs each time a dependent is bumped.

Why wouldn't we want to do that ? AFAIK, we don't have lots of packs involving non-Symfony packages (Symfony packages being covered by the special Symfony constraint of Flex).

Unpacking automatically in the require command when Flex is overriding it will still not prevent people from having such package in their requirement for another reason. If we keep an `*` requirement there, we would have to explicitly document that our pack packages are not following semver at all and are not covered by the Symfony BC promise.

nicolas-grekas:
Because there's zero value in doing so. We need to simplify our processes. I would dedicate zero minutes of my OSS time to maintaining this. And I'm sure that's the same for everyone.

semver/BC policy is about code, not about dependencies. That's a common misunderstanding.

What issues do you see in that proposal? Combined with flex (packs are a flex-only thing), resolving the `*` when unpacking has no drawbacks to me.

stof:
@nicolas-grekas semver is about the public API of your package, which might involve some APIs defined in dependencies (if you return one of their objects for instance). For a pack, the *only* API it has are the transitive dependencies, as that's what the pack is about.

stof:
> packs are a flex-only thing

They are not. they are registered on Packagist as normal packages.

nicolas-grekas:
There is no API in a pack. This interpretation you give is way too extensive.

> packs are a flex-only thing

Yes they are: `"type": "symfony-pack"`. If one installs a pack and doesn't use flex, they're on their own.

[...]

PR #647, opened 2020-07-07
Fork Interval classes from composer/semver

nicolas-grekas:
This is a copy/paste from https://github.com/composer/semver/ v3.

We're going to need this logic to implement #645

Forking is required because composer/semver v3 is shipped with Composer 2 but not with Composer 1.

/cc @Seldaek FYI

### DIFF
--- a/src/Semver/FullConstraint.php
+++ b/src/Semver/FullConstraint.php
@@ -1,0 +1,59 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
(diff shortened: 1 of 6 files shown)


Seldaek:
Obviously you're free to do this but IMO it might be better to make the new functionality composer 2 only than copy paste all this code. Up to you.

stof:
why do you need this logic to implement #645 ?

nicolas-grekas:
We'll need this logic when we'll want to merge version constraints to support anything else than `*`, or to support `conflict`/`replace`.

I don't think we can bind this to Composer 2. The behavior on Composer 1 would have to be defined anyway and there is no other appropriate behavior.

nicolas-grekas:
Closing because I'm not going to need this I think.
Could be reopened if needed.

PR #648, opened 2020-07-08
Support more than just "master" as the default branch

nicolas-grekas:


src/Downloader.php
```diff
@@ -110,6 +110,15 @@ public function getRecipes(array $operations): array
...
                     (isset($branchAliases[$version]) && $alias = $branchAliases[$version]) ||
+                    (isset($branchAliases['dev-main']) && $alias = $branchAliases['dev-main']) ||
+                    (isset($branchAliases['dev-trunk']) && $alias = $branchAliases['dev-trunk']) ||
+                    (isset($branchAliases['dev-develop']) && $alias = $branchAliases['dev-develop']) ||
+                    (isset($branchAliases['dev-default']) && $alias = $branchAliases['dev-default']) ||
+                    (isset($branchAliases['dev-latest']) && $alias = $branchAliases['dev-latest']) ||
+                    (isset($branchAliases['dev-next']) && $alias = $branchAliases['dev-next']) ||
+                    (isset($branchAliases['dev-current']) && $alias = $branchAliases['dev-current']) ||
+                    (isset($branchAliases['dev-support']) && $alias = $branchAliases['dev-support']) ||
+                    (isset($branchAliases['dev-tip']) && $alias = $branchAliases['dev-tip']) ||
```
stof, line 121: Why do we even have all these cases ? To match the behavior of composer, only `$branchAliases[$version]` should be relevant. Branch aliases defined for other branches are totally ignored by composer (though they are allowed, so that you don't have to remove them between branches which would be a mess)

nicolas-grekas: This looks at the branch-alias to decide which version of the recipe should be installed.
We can support more items than composer here.

stof: checking the branch alias of `dev-tip` when the installed version is not `dev-tip` looks weird to me. You want to install the recipe for the installed version, not for a random version.

nicolas-grekas: it's still pretty common when checking out a custom dev-branch without aliasing it. E.g. for Symfony components, falling back to the dev-master alias is the best way to guess the version. That's why there is logic IIUC.

PR #656, opened 2020-07-10
Unpack by default when running composer require

### LINKED ISSUE #645: Revisit the way packs work - always unpack them

nicolas-grekas:
Packs are like meta-recipes: ppl should adopt them in their root composer.json so that updating packs (like updating recipes) doesn't affect their project.

If one needs to delegate the selection of a group of packages, they should use the `meta-package` type of composer instead.

Fixes #645

### DIFF
--- a/src/Command/RequireCommand.php
+++ b/src/Command/RequireCommand.php
@@ -32,19 +32,19 @@ class RequireCommand extends BaseRequireCommand
     protected function configure()
     {
         parent::configure();
-        $this->addOption('unpack', null, InputOption::VALUE_NONE, 'Unpack Symfony packs in composer.json.');
+        $this->addOption('no-unpack', null, InputOption::VALUE_NONE, 'Disable unpacking Symfony packs in composer.json.');
+        $this->addOption('unpack', null, InputOption::VALUE_NONE, '[DEPRECATED] Unpacking is now enabled by default.');
     }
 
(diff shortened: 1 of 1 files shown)


stof:
This will make `composer require` and `composer require --unpack` being exactly the same, with only `composer require --unpack=false` skipping it. This is a confusing API IMO. A `--no-unpack` flag would be easier to understand.

nicolas-grekas:
@stof updated. This requires deprecating `--unpack`.

fabpot:
Thank you @nicolas-grekas.

PR #663, opened 2020-07-15
use static construction for static method

Gemorroj:
https://github.com/composer/composer/blob/2.0.0-alpha2/src/Composer/Package/Locker.php#L78
https://github.com/composer/composer/blob/1.10.8/src/Composer/Package/Locker.php#L73
`Locker::getContentHash` is static method.


nicolas-grekas:
Please rebase to get rid of the merge commit.

Gemorroj:
idk how.
#664

PR #666, opened 2020-07-29
Fix travis file name

maxhelias:


maxhelias:
I don't know how to fix PHP nightly stage

tests/Configurator/CopyFromPackageConfiguratorTest.php
```diff
@@ -58,7 +58,7 @@ public function testConfigureAndOverwriteFiles()
 
         $this->io->expects($this->at(0))->method('writeError')->with(['    Copying files from package']);
         $this->io->expects($this->at(2))->method('writeError')->with(['      Created <fg=green>"./public/file"</>']);
-        $this->io->method('askConfirmation')->with('File "build/public/file" has uncommitted changes, overwrite? [y/N] ')->willReturn(true);
+        $this->io->method('askConfirmation')->with('File "file" has uncommitted changes, overwrite? [y/N] ')->willReturn(true);
```
nicolas-grekas, line 61: why did this pass before and it does now?
wasn't the previous path better for the purpose?

maxhelias: I don't know, I can't explain it, could it be that the git output has changed?
Indeed, the previous path is better

maxhelias:
@nicolas-grekas It should be good, what do you think?

.travis.yml
```diff
@@ -1,16 +1,20 @@
...
+  depth: 5
+
+jobs:
   include:
     - php: 7.1
       env: COMPOSER_FLAGS='--prefer-lowest --prefer-stable'
     - php: 7.2
     - php: 7.3
     - php: 7.4
-    - php: nightly
```
fabpot, line 12: This should not be removed (right now, that disables testing on nightly).

PR #668, opened 2020-08-06
New flag to list only outdated recipes

maxhelias:
https://twitter.com/Aerendir/status/1291310746329063424 : poke @weaverryan

### DIFF
--- a/src/Command/RecipesCommand.php
+++ b/src/Command/RecipesCommand.php
@@ -16,6 +16,7 @@ use Composer\Downloader\TransportException;
 use Composer\Util\HttpDownloader;
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Input\InputOption;
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Flex\InformationOperation;
 use Symfony\Flex\Lock;
@@ -49,6 +50,7 @@ class RecipesCommand extends BaseCommand
             ->setDefinition([
                 new InputArgument('package', InputArgument::OPTIONAL, 'Package to inspect, if not provided all packages are.'),
             ])
+            ->addOption('outdated', 'o', InputOption::VALUE_NONE, 'Show only recipes that are outdated')
         ;
     }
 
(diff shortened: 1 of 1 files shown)


src/Command/RecipesCommand.php
```diff
@@ -96,35 +98,49 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
+
+            if (!$outdated) {
+                $write[] = sprintf(' * %s %s', $name, $additional);
+
+                continue;
+            }
+
+            if (null !== $additional) {
+                $write[] = sprintf(' * %s %s', $name, $additional);
+            }
```
weaverryan, line 123: This all may be correct, but it doesn’t read clearly to me. I expect the algorithm to read like this: 

* determine if the recipe is outdated (you already do this first via the $additional variable)
* then, if the recipe is NOT outdated AND the outdated flag is set, continue (skip printing)
* else, print (because the recipe is either outdated or we are showing all recipes)

src/Command/RecipesCommand.php
```diff
@@ -96,35 +98,49 @@ protected function execute(InputInterface $input, OutputInterface $output)
...
 
-        $write[] = '';
-        $write[] = 'Run:';
-        $write[] = ' * <info>composer recipes vendor/package</info> to see details about a recipe.';
-        $write[] = ' * <info>composer recipes:install vendor/package --force -v</info> to update that recipe.';
-        $write[] = '';
+        // Nothing to display
+        if (empty($write)) {
+            return 0;
+        }
```
weaverryan, line 129: We also need to return 1 if the outdated option was passed AND there are some outdated items. 

I would probably remove the return 0 here and only worry about the return 1 here:

If the outdated flag was passed AND $write is NOT empty, return 1.

weaverryan: I’m also not a huge fan of leveraging $write to know if we have at least 1 outdated package. I’d rather have a specific Boolean for that (like $hasOutdatedRecipes) that we initialize to false before the loop above and set to true inside of an outdated is found. But this is subjective.

Aerendir:
I was thinking at this and maybe a better naming would be “updated”.

There is a difference between this and composer: when a composer dep is update, you have to update it, too (so the dependency is outdated).

But when a recipe is updated, you are not forced to update it because maybe you have already done it or you have customized the config in the mean time.

So, it is useful to know which recipes are updated, but I don’t think they are outdated, because The app doesn’t have to update it necessarily,

I don’t know if my thoughts are clear to you. Please, tell me if they aren’t.

src/Command/RecipesCommand.php
maxhelias: The return here allows nothing to be displayed as presentation text.

I'm not sure I updated as you want, please tell me if it's not ok

maxhelias:
@Aerendir I'm not sur, "updated" and "outdated" are the opposites. And the option allows you to display only non-up-to-date recipes.

If you update any recipes manually, you won't have the correct refs in the symfony.lock file. So you will not be able to fully benefit from Flex such as viewing recipe information.
The notion is very close to that of composer and in this case "outdated" goes well

stof:
@maxhelias the opposite of `outdated` is `uptodate`, not `updated`.

PR #677, opened 2020-09-14
Fix unpacking nested packs into composer.json

### LINKED ISSUE #676: Unpacking pack with nested packs does not save nested packs
**Description**  
It seems that unpacking a pack with nested pack does not copy nested pack's dependencies in composer.json, leading to a remove of nested pack dependencies on a `composer update` command.

**How to reproduce**  
Try with `symfony/debug-pack` that have, in requirements, `symfony/profiler-pack` :
```bash
symfony composer req symfony/debug-pack --dev
# It will install symfony/web-profiler-bundle, symfony/profiler-pack, symfony/debug-bundle, 
# symfony/monolog-bundle, symfony/var-dumper and symfony/debug-pack then remove them (unpacking).
# But in fact, in composer.json you will see only symfony/debug-bundle, symfony/var-dumper
# and symfony/monolog-bundle
symfony composer upd
# Then, you will see that symfony/web-profiler (and its dependencies) is removed 
# since it is not required in composer.json.

```

**Additional context**  
A temporary workaround is to require first, in this case, `symfony/profiler-pack` and then `symfony/debug-pack`

nicolas-grekas:
Fix #676

### DIFF
--- a/src/Unpacker.php
+++ b/src/Unpacker.php
@@ -41,13 +41,11 @@ class Unpacker
         $this->dryRun = $dryRun;
     }
 
-    public function unpack(Operation $op, Result $result = null): Result
+    public function unpack(Operation $op, Result $result = null, &$links = []): Result
     {
         if (null === $result) {
             $result = new Result();
         }
-
-        $links = [];
 
         $localRepo = $this->composer->getRepositoryManager()->getLocalRepository();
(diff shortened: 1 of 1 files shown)


fabpot:
Thank you @nicolas-grekas.

PR #684, opened 2020-10-06
Fixed not working "--force" option if you want to reinstall a recipe

malteschlueter:
Sometimes it's necessary to reinstall a recipe but actually it's not possible if there is not update for a recipe. Also with the `--force` option Flex won't reinstall/execute the recipe. Only if there is a update for a recipe.

I figured out that the $recipes in [Flex.php#L466](https://github.com/symfony/flex/blob/115e67f76ba95d70946a6e0b15d4578bf04927c3/src/Flex.php#L466) are empty. It can't find/collects the recipe because the existing ref and the "reinstall" ref are same, see in [Flex.php#L817](https://github.com/symfony/flex/blob/115e67f76ba95d70946a6e0b15d4578bf04927c3/src/Flex.php#L817).

So i tried to fix it but to be honest i'm not sure it's a good solution.

---

What do you think?

---

### To reproduce:
1. Create a new directory
1. Execute `composer require symfony/flex`
1. Execute `composer require symfony/apache-pack`
1. Delete or truncate file `public/.htaccess`
1. Execute `composer recipes:install symfony/apache-pack --force`

### DIFF
--- a/src/Flex.php
+++ b/src/Flex.php
@@ -463,7 +463,7 @@ class Flex implements PluginInterface, EventSubscriberInterface
             copy($rootDir.'/.env.dist', $rootDir.'/.env');
         }
 
-        $recipes = $this->fetchRecipes($this->operations);
+        $recipes = $this->fetchRecipes($this->operations, $this->isUpdateEventForced($event));
         $this->operations = [];     // Reset the operation after getting recipes
 
         if (2 === $this->displayThanksReminder) {
@@ -540,7 +540,7 @@ class Flex implements PluginInterface, EventSubscriberInterface
                 case 'install':
                     $this->io->writeError(sprintf('  - Configuring %s', $this->formatOrigin($recipe->getOrigin())));
                     $this->configurator->install($recipe, $this->lock, [
-                        'force' => $event instanceof UpdateEvent && $event->force(),
+                        'force' => $this->isUpdateEventForced($event),
                     ]);
                     $manifest = $recipe->getManifest();
                     if (isset($manifest['post-install-output'])) {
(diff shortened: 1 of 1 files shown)


src/Flex.php
```diff
@@ -999,4 +999,9 @@ public static function getSubscribedEvents(): array
 
         return $events;
     }
+
+    private function isUpdateEventForced(?Event $event): bool
```
nicolas-grekas, line 1003: I don't think we need this method

nicolas-grekas: nvm...

nicolas-grekas:
I'm mixed on this one. This will create more changes than required when syncing recipes, increasing the burden of the task.
You can achieve the same by first removing symfony.lock, the running with `--force`.

nicolas-grekas:
Sorry I closed by mistake (from PR linked)

malteschlueter:
> You can achieve the same by first removing symfony.lock, then running with `--force`.

You're right in the end the results are the same but i think this is only a workaround.

For example in the help message of the `--force` option it says `Ignore the "symfony.lock" file and overwrite existing files`.

----

> I'm mixed on this one. This will create more changes than required when syncing recipes, increasing the burden of the task.
Do you mean with it will create more changes than required in the project source code or in the source code of flex?

Today i checked the recipes of a project and there was a update for the `symfony/framework-bundle` available.

If you're running `composer recipes:install symfony/framework-bundle` you get the message:
```
Recipe(s) already installed for: symfony/framework-bundle
Re-run the command with --force to re-install the recipes.

No recipes to install.
```

Then i run the command again with the option `--force` and it does exactly what I'm expecting after I read the help/info message. It re-installs the complete recipe and overrides all files or add missing ones. But that is easy to solve if you're using i.e. Git :D :)

xabbuh:
Did #756 solve this?

PR #692, opened 2020-10-08
Unpack replace and provides of packs

### LINKED ISSUE #653: Unpacking should unpack "replace" rules
If a pack lists things in its "replace" section, unpacking this pack should copy these rules into the "replace" section of the root composer.json.

I think that implementing this requires patching the `Unpacker` class only.

Anyone up to give it a try?

BackEndTea:
This fixes #653

### DIFF
--- a/src/Unpacker.php
+++ b/src/Unpacker.php
@@ -74,7 +74,7 @@ class Unpacker
 
             $versionSelector = null;
 
-            if(!$package['dev']) {
+            if (!$package['dev']) {
                 foreach ($pkg->getReplaces() as $link) {
                     $this->addLinks($link, $links, 'replace');
                 }
@@ -117,7 +117,7 @@ class Unpacker
 
                 if (isset($links[$linkName])) {
                     $links[$linkName]['constraints'][] = $constraint;
-                    if ('require' === $linkType && !in_array($links['linkName']['type'],['replace', 'provide'])) {
+                    if ('require' === $linkType && !\in_array($links['linkName']['type'],['replace', 'provide'])) {
                         $links[$linkName]['type'] = 'require';
                     }
                 } else {
--- a/src/Unpacker.php
+++ b/src/Unpacker.php
@@ -74,18 +74,13 @@ class Unpacker
 
             $versionSelector = null;
 
-            foreach ($pkg->getReplaces() as $link) {
-                if ($package['dev']) {
-                    break;
+            if(!$package['dev']) {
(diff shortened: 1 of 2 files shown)


src/Unpacker.php
```diff
@@ -60,7 +61,7 @@ public function unpack(Operation $op, Result $result = null, &$links = []): Resu
                 null === $pkg ||
                 'symfony-pack' !== $pkg->getType() ||
                 !$op->shouldUnpack() ||
-                0 === \count($pkg->getRequires()) + \count($pkg->getDevRequires())
+                0 === \count($pkg->getRequires()) + \count($pkg->getDevRequires()) + \count($pkg->getReplaces()) + \count($pkg->getReplaces())
```
stof, line 64: why counting the replaces twice ?

src/Unpacker.php
```diff
@@ -72,6 +73,21 @@ public function unpack(Operation $op, Result $result = null, &$links = []): Resu
             }
 
             $versionSelector = null;
+
+            foreach ($pkg->getReplaces() as $link) {
+                if ($package['dev']) {
```
stof, line 78: Why checking this for each iteration while it does not depend on the loop at all ? Put a `if` around the loop instead.

src/Unpacker.php
```diff
@@ -200,4 +216,20 @@ public function updateLock(Result $result, IOInterface $io): void
...
+
+    private function addLinks(Link $link, array &$links, string $linkType): void
+    {
+        $constraint = $link->getPrettyConstraint();
+        $constraint = substr($this->resolver->parseVersion($link->getTarget(), $constraint, false), 1) ?: $constraint;
+
+        $linkName = $link->getTarget();
+        $constraint = $this->versionParser->parseConstraints($constraint);
+
+        $links[$linkName] = [
```
stof, line 229: This looks wrong to me. You could have a `conflict` and a `require-dev` on the same package. Using the same `$links` indexed by package name won't work fine.

tests/UnpackerTest.php
```diff
@@ -12,6 +12,16 @@
...
 use Symfony\Flex\Unpacker;
+use function class_exists;
+use function file_get_contents;
+use function file_put_contents;
+use function getenv;
+use function json_decode;
+use function mkdir;
+use function putenv;
+use function unlink;
+use function var_dump;
```
stof, line 23: We don't import global functions in Symfony. Please don't change the coding standards of the project.

PR #693, opened 2020-10-13
Fix InformationOperation breaking on latest Composer 2.x

Seldaek:
See https://github.com/composer/composer/commit/09ef026d43fc5ac032d083241e2cab0320ddd325


nicolas-grekas:
Thank you @Seldaek.

PR #696, opened 2020-10-13
Fix support for ".x" branches

nicolas-grekas:
This also fixes support for branch `5.x`

### DIFF
--- a/src/PackageFilter.php
+++ b/src/PackageFilter.php
@@ -99,9 +99,10 @@ class PackageFilter
             foreach ($vers as $i => $v) {
                 if (!isset($okVersions[$v])) {
                     $okVersions[$v] = false;
+                    $w = '.x' === substr($v, -2) ? $versions['next'] : $v;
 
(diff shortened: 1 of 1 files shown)


src/Cache.php
```diff
@@ -83,7 +83,7 @@ public function removeLegacyTags(array $data): array
         }
 
         foreach ($symfonySymfony as $version => $composerJson) {
-            if (null !== $alias = $composerJson['extra']['branch-alias'][$version] ?? null) {
+            if (null !== $alias = $composerJson['extra']['branch-alias'][$version] ?? $composerJson['extra']['branch-version'] ?? null) {
```
stof, line 86: should we actually use `branch-version` for tagged releases ?

maxhelias (approved):

src/Cache.php
nicolas-grekas: You're right, I removed these, no need to do anything special for "branch-version"

PR #699, opened 2020-10-14
Avoid filtering out aliases of valid packages

Seldaek:
It'd be good to add tests for this ... and I am not sure if the branch-alias check at line 67/68 is still needed if aliases are handled properly, might be or might not be.. I don't fully grasp the scope of what flex does here.


nicolas-grekas:
Thank you @Seldaek.

PR #702, opened 2020-10-15
Fixed not working "--force" option if you want to reinstall a recipe

### LINKED ISSUE #777: Reinstalling recipes works only for updated ones
Trying to rollback recipe files but no files modified
```
composer recipes:install symfony/webpack-encore-bundle --force -v
```
Only message
`Files have been reset to the latest version of the recipe.`

To reproduce:
```
composer create-project symfony/skeleton issue
cd ./issue
rm ./src/Kernel.php
composer recipes:install symfony/framework-bundle --force
```

malteschlueter:
Fix #777

Sometimes it's necessary to reinstall a recipe but actually it's not possible if there is not update for a recipe. Also with the `--force` option Flex won't reinstall/execute the recipe. Only if there is a update for a recipe.

I figured out that the $recipes in [Flex.php#L466](https://github.com/symfony/flex/blob/115e67f76ba95d70946a6e0b15d4578bf04927c3/src/Flex.php#L466) are empty. It can't find/collects the recipe because the existing ref and the "reinstall" ref are same, see in [Flex.php#L817](https://github.com/symfony/flex/blob/115e67f76ba95d70946a6e0b15d4578bf04927c3/src/Flex.php#L817).

So i tried to fix it but to be honest i'm not sure it's a good solution.

---

What do you think?

---

### To reproduce:
1. Create a new directory
1. Execute `composer require symfony/flex`
1. Execute `composer require symfony/apache-pack`
1. Delete or truncate file `public/.htaccess`
1. Execute `composer recipes:install symfony/apache-pack --force`

---

_Previous PR #684 was closed automatically after changing the default branch from master to main._

### DIFF
--- a/src/Flex.php
+++ b/src/Flex.php
@@ -469,7 +469,7 @@ class Flex implements PluginInterface, EventSubscriberInterface
         }
 
         // Execute missing recipes
-        $recipes = $this->fetchRecipes($this->operations);
+        $recipes = $this->fetchRecipes($this->operations, $this->isUpdateEventForced($event));
         $this->operations = [];     // Reset the operation after getting recipes
 
         if (2 === $this->displayThanksReminder) {
@@ -549,7 +549,7 @@ class Flex implements PluginInterface, EventSubscriberInterface
                 case 'install':
                     $this->io->writeError(sprintf('  - Configuring %s', $this->formatOrigin($recipe->getOrigin())));
                     $this->configurator->install($recipe, $this->lock, [
-                        'force' => $event instanceof UpdateEvent && $event->force(),
+                        'force' => $this->isUpdateEventForced($event),
                     ]);
                     $manifest = $recipe->getManifest();
                     if (isset($manifest['post-install-output'])) {
@@ -790,7 +790,7 @@ EOPHP
         );
     }
 
-    public function fetchRecipes(array $operations): array
+    public function fetchRecipes(array $operations, bool $force = false): array
     {
         if (!$this->downloader->isEnabled()) {
             $this->io->writeError('<warning>Symfony recipes are disabled: "symfony/flex" not found in the root composer.json</>');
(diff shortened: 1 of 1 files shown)


xabbuh (approved):

nicolas-grekas:
I'm :-1: sorry.

As explained in https://github.com/symfony/flex/pull/684#issuecomment-707799669, this will make updating recipes harder, because it will force ppl to review changes that don't need to be.

As you spotted in https://github.com/symfony/flex/pull/684#issuecomment-708228981, the description of the `--force` option is wrong and is fixed by #714

You should manually remove the symfony.lock file if you want this exact behavior.

Note that this is not the last word on recipe updates, there are open discussions on the topic and things will improve I hope too.

malteschlueter:
Okay, no problem.

Maybe it would be useful to split installation and updates in 2 commands.

dunglas:
I've a use case with [Symfony Docker](https://github.com/dunglas/symfony-docker) that could require to this PR (we discussed this with @nicolas-grekas privately): I want people to be able to add the Docker configuration to existing projects. To do so, they have to copy the Docker-related files and to set `symfony.extra.docker` to `true` in their `composer.json`. As some recipes modify the `Dockerfile` and `docker-compose.*.yml`, the user also have to re-execute the recipes. Currently, it's not possible, this PR could fix this.

Maybe could we have two separate flags such as `--force` to force re-executing **all** recipes and `--force-updated` to force re-executing only the recipes of updated packages?

nicolas-grekas:
I've changed my mind on this topic. I think that the new name of the command (recipes:install) changes the expectation and that'd I'm fine changing this behavior. No need to have two flags.

@malteschlueter would you be able to rebase this PR and fix the CS issue?

dunglas:
Actually, having the ability to **not** reinstall the recipes that haven't change is very useful for another projet I maintain: API Platform. In the API Platform distribution, we made some customizations to the generated config files, but we have to sync recipes with Symfony before each release. This option is useful in this case, it reduces the maintenance burden.

nicolas-grekas:
> having the ability to not reinstall the recipes that haven't change is very useful for another projet I maintain

without `--force` does it?

dunglas:
I mean reinstalling only the recipes that have changed, and not touching to the one that haven't changed. But maybe is it the default behavior?

nicolas-grekas:
Replaced by #795, thanks for the proposal and the discussion.

PR #704, opened 2020-10-15
Don't filter locked packages

nicolas-grekas:
As hinted in https://github.com/composer/composer/issues/9290#issuecomment-708965515

### DIFF
--- a/src/PackageFilter.php
+++ b/src/PackageFilter.php
@@ -40,13 +40,22 @@ class PackageFilter
 
     /**
      * @param PackageInterface[] $data
+     * @param PackageInterface[] $lockedPackages
      *
      * @return PackageInterface[]
      */
-    public function removeLegacyPackages(array $data): array
+    public function removeLegacyPackages(array $data, array $lockedPackages): array
     {
         if (!$this->symfonyConstraints || !$data) {
             return $data;
+        }
+
+        $lockedVersions = [];
+        foreach ($lockedPackages as $package) {
+            $lockedVersions[$package->getName()] = [$package->getVersion()];
+            if ($package instanceof AliasPackage) {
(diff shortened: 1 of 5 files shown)


src/PackageFilter.php
```diff
@@ -40,15 +40,24 @@ public function __construct(IOInterface $io, string $symfonyRequire, Downloader
...
-    public function removeLegacyPackages(array $data): array
+    public function removeLegacyPackages(array $data, array $lockedPackages): array
     {
         if (!$this->symfonyConstraints || !$data) {
             return $data;
         }
 
+        foreach ($lockedPackages as $i => $package) {
+            unset($lockedPackages[$i]);
+            $lockedPackages[$package->getName()] = [$package->getVersion()];
```
stof, line 55: to convert from a `list<Package>` to an indexed array `array<string, Package[]>`, I suggest creating a new array instead of reusing the same one. It will make the code much easier to understand

tests/PackageFilterTest.php
```diff
@@ -52,7 +52,7 @@ public function testRemoveLegacyPackages(array $expected, array $packages, strin
         $expected = $configToPackage($expected);
         $packages = $configToPackage($packages);
 
-        $actual = $filter->removeLegacyPackages($packages);
+        $actual = $filter->removeLegacyPackages($packages, []);
```
stof, line 55: We should have a test covering the non-empty array.

src/PackageFilter.php
nicolas-grekas: updated

tests/PackageFilterTest.php
nicolas-grekas: test added

Issue #712, opened 2020-11-04
Really difficult to update to a dev version


Jean85:
I have a Symfony 5.1 project with Composer 2, Flex and various components installed.

I would like to upgrade to the current `5.x` branch, because there's a PR that I need that is merged there but it's not included in the latest 5.2-BETA tag. Without Flex, I would `composer require symfony/symfony:dev-5.x`, but with multiple split components I imagined that it would be more complex...

I tried multiple aliases in `extra.symfony.require`, but it always said `Nothing to update`; to make it work, I had to set `"minimum-stability": "dev"` in my `composer.json`, and do a `composer update` listing ALL the installed components, transient dependencies included, because otherwise it would update all the other dependencies too to the latest `dev` version.
Using `--prefer-stable` would stop Symfony components from updating.

This is obviously a sub-optimal DX; am I missing something? Or is this something that could be improved?

nicolas-grekas:
Do you have any suggestions? This looks like standard composer.json updates, isn't it?

wouterj:
Instead of doing `composer update` with all Symfony packages, I recommend adding `"prefer-stable": true` and then setting `extra.symfony.require` to `5.2`.

Seldaek:
@wouterj that would require you to upgrade all packages at once to 5.2-dev tho.. If you want to upgrade just one, why doesn't flex let you do it? IMO it's fairly simple, if flex sees the root package has a require/require-dev for symfony/foo in a version that does not intersect with the extra.symfony.require constraint, then it should not filter that package's versions.

Then you could leave the extra.symfony.require at 5.1, but if you explicitly require 5.x-dev for a given package it would let you install that.

wouterj:
Oh, I definitely didn't want to say that there is nothing to improve here. I was just suggesting a better possible alternative of _"[...] and do a `composer update` listing ALL the installed components, transient dependencies included, because otherwise it would update all the other dependencies too to the latest `dev` version"_
(my knowledge of Composer and Flex internals is honestly way to low to have any useful opinion on the actual matter)

Jean85:
> Instead of doing composer update listing all Symfony packages, I recommend adding "prefer-stable": true and then setting extra.symfony.require to 5.2. After this, you can run composer update without any extra arguments.

If that setting has the same efffect of `--prefer-stable`, it won't work. Prefer stable leaves the Symfony components without updates, since they are already stable, and the "suggested" contraint from `extra.symfony.require` is unstable.

[...]

PR #713, opened 2020-11-04
Use root constraints when they have no intersection with extra.symfony.require

nicolas-grekas:
Fixes https://github.com/symfony/flex/issues/712
Fixes https://github.com/composer/composer/issues/9409
Fixes https://twitter.com/scherzmut/status/1323609763125612546

### DIFF
--- a/src/Flex.php
+++ b/src/Flex.php
@@ -140,8 +140,9 @@ class Flex implements PluginInterface, EventSubscriberInterface
             $this->downloader = $downloader = new Downloader($composer, $io, $this->rfs);
             $this->downloader->setFlexId($this->getFlexId());
 
+            $rootPackage = $composer->getPackage();
             $manager = RepositoryFactory::manager($this->io, $this->config, $composer->getEventDispatcher(), $this->rfs);
-            $setRepositories = \Closure::bind(function (RepositoryManager $manager) use (&$symfonyRequire, $downloader) {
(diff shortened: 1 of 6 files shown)


src/Flex.php
```diff
@@ -863,7 +863,11 @@ public function truncatePackages(PrePoolCreateEvent $event)
         if (!$this->filter) {
             return;
         }
-        $event->setPackages($this->filter->removeLegacyPackages($event->getPackages(), $event->getRequest()->getFixedOrLockedPackages()));
+
+        $rootPackage = $this->composer->getPackage();
+        $lockedPackages = $event->getRequest()->getFixedOrLockedPackages();
```
Seldaek, line 868: This is not compatible with composer 1, so you may want to make it optional somehow, Semver\Intervals also does not exist in v1.

I guess it's already broken in the current release tho.. Or is this code path only run for composer 2?

Seldaek:
Looks good to me for v2 that is, I'd say people should upgrade anyway but I am biased.

src/Flex.php
nicolas-grekas: this codepath runs on c2 only, yes

jderusse (approved):
I'm fine with fixing this for composer 2 only.

Let people open a PR if they want fixing this issue on composer 1

nicolas-grekas:
OK, support for Composer 1 wasn't that hard, pushed :)

PR #705, opened 2020-10-16
Add support for multiple Docker Compose files

dunglas:
Allows recipes to ship configuration for multiple Docker Compose files (e.g. `docker-compose.override.yml`, `docker-compose.production.yml`). This is convenient to provide different settings for the development and the production environment. This is a companion PR for dunglas/symfony-docker#73.

PR #715, opened 2020-11-06
Improve Docker Compose file detection

dunglas:
This PR implements the rules described in the Docker Compose manual: https://docs.docker.com/compose/reference/envvars/#compose_file (see also https://github.com/symfony/flex/pull/705#issuecomment-722569904, cc @jycamier).

Basically, the directory containing the Compose definition can be configured using the `COMPOSE_FILE` environment variable, and it isn't defined Flex will look for the Compose file (as `docker-compose itself`) in the project directory and then in each parent director. 

This will be especially helpful for API Platform, where the `docker-compose.yml` file is in the parent directory of the Symfony application (https://github.com/api-platform/api-platform/), but also for other monolithic Git repositories containing several applications and a `docker-compose.yaml` file at the root (which is very common and considered a best practice).

There is a voluntary limitation regarding the `COMPOSE_FILE` environment variable: the name of the file **must** be the one defined in the recipe (with the `.yml` or `.yaml` extension). In other words, only the directory is configurable, not the file name (on purpose).

### DIFF
--- a/src/Configurator/DockerComposeConfigurator.php
+++ b/src/Configurator/DockerComposeConfigurator.php
@@ -11,7 +11,11 @@
 
 namespace Symfony\Flex\Configurator;
 
+use Composer\Composer;
+use Composer\IO\IOInterface;
+use Symfony\Component\Filesystem\Filesystem;
 use Symfony\Flex\Lock;
+use Symfony\Flex\Options;
 use Symfony\Flex\Recipe;
 
 /**
@@ -21,6 +25,15 @@ use Symfony\Flex\Recipe;
  */
 class DockerComposeConfigurator extends AbstractConfigurator
 {
+    private $filesystem;
+
+    public function __construct(Composer $composer, IOInterface $io, Options $options)
+    {
+        parent::__construct($composer, $io, $options);
+
+        $this->filesystem = new Filesystem();
+    }
+
     public function configure(Recipe $recipe, $config, Lock $lock, array $options = [])
     {
         $installDocker = $this->composer->getPackage()->getExtra()['symfony']['docker'] ?? false;
@@ -30,17 +43,14 @@ class DockerComposeConfigurator extends AbstractConfigurator
 
         $rootDir = $this->options->get('root-dir');
         foreach ($this->normalizeConfig($config) as $file => $extra) {
-            $dockerComposeFile = sprintf('%s/%s', $rootDir, $file);
-
-            // Test with the ".yaml" extension if the file doesn't end up with ".yml".
             if (
-                (!file_exists($dockerComposeFile) && !file_exists($dockerComposeFile = substr($dockerComposeFile, 0, -2).'aml')) ||
+                (null === $dockerComposeFile = $this->findDockerComposeFile($rootDir, $file)) ||
                 $this->isFileMarked($recipe, $dockerComposeFile)
             ) {
                 continue;
             }
 
-            $this->write(sprintf('Adding Docker Compose entries to "%s"', $dockerComposeFile));
+            $this->write(sprintf('Adding Docker Compose definitions to "%s"', $dockerComposeFile));
 
             $offset = 8;
             $node = null;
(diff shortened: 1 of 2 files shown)


src/Configurator/DockerComposeConfigurator.php
```diff
@@ -83,14 +93,15 @@ public function configure(Recipe $recipe, $config, Lock $lock, array $options =
...
+        $this->write('Docker Compose definitions have been modified. Please run "docker-compose up" again to apply the changes.');
     }
 
     public function unconfigure(Recipe $recipe, $config, Lock $lock)
     {
         $rootDir = $this->options->get('root-dir');
         foreach ($this->normalizeConfig($config) as $file => $extra) {
-            $dockerComposeFile = sprintf('%s/%s', $rootDir, $file);
-            if (!file_exists($dockerComposeFile) && !file_exists($dockerComposeFile = substr($dockerComposeFile, 0, -2).'aml')) {
+            if (null === $dockerComposeFile = $dockerComposeFile = $this->findDockerComposeFile($rootDir, $file)) {
```
chalasr, line 104: extra `$dockerComposeFile = `

chalasr (approved):
Nice one

src/Configurator/DockerComposeConfigurator.php
```diff
@@ -125,6 +138,45 @@ private function normalizeConfig(array $config): array
...
+            foreach ($files as $f) {
+                if ($file !== basename($f)) {
+                    continue;
+                }
+
+                if (!$this->filesystem->isAbsolutePath($f)) {
+                    $f = realpath(sprintf('%s/%s', $rootDir, $f));
+                }
+
+                return $this->filesystem->exists($f) ? $f : null;
```
jderusse, line 157: while returning null otherwise? shouldn't we fallback to other files defined in the `COMPOSE_FILE` variables?

src/Configurator/DockerComposeConfigurator.php
```diff
@@ -125,6 +138,45 @@ private function normalizeConfig(array $config): array
...
+            }
+        }
+
+        // COMPOSE_FILE not set, or doesn't contain the file we're looking for
+        $dir = $rootDir;
+        $previousDir = null;
+        do {
+            // Test with the ".yaml" extension if the file doesn't end up with ".yml".
+            if (
+                file_exists($dockerComposeFile = sprintf('%s/%s', $dir, $file)) ||
```
jderusse, line 167: `file_exists` versus `filesystem->exists` at line 157?

src/Configurator/DockerComposeConfigurator.php
dunglas: Indeed we can do that. I'll change this.

src/Configurator/DockerComposeConfigurator.php
```diff
@@ -125,6 +138,45 @@ private function normalizeConfig(array $config): array
         return $config;
     }
 
+    /**
+     * Finds the Docker Compose file according to these rules: https://docs.docker.com/compose/reference/envvars/#compose_file.
+     */
+    private function findDockerComposeFile(string $rootDir, string $file): ?string
+    {
+        if (isset($_SERVER['COMPOSE_FILE'])) {
+            $files = explode('\\' === \DIRECTORY_SEPARATOR ? ';' : ':', $_SERVER['COMPOSE_FILE']);
```
jycamier, line 147: `\DIRECTORY_SEPARATOR` is enought to know if you're working on Linux / Mac or Windows.

However, you can change the separator between files with `COMPOSE_PATH_SEPARATOR` environment variable.

https://docs.docker.com/compose/reference/envvars/#compose_path_separator

dunglas: Thanks for the hint about `COMPOSE_PATH_SEPARATOR`, using it now!

> \DIRECTORY_SEPARATOR is enought to know if you're working on Linux / Mac or Windows.

It's already what is used in Flex and it's what we used to do in Symfony.
It's good enough here because we only care about Windows. We cannot use `PHP_OS_FAMILY` yet because we still support old PHP versions and as it's a Composer plugin we cannot depend on the polyfill.

src/Configurator/DockerComposeConfigurator.php
dunglas: Changed every occurrences to use `Filesystem` for consistency.

dunglas:
This one is ready to be merged!

src/Configurator/DockerComposeConfigurator.php
jycamier: You're welcome.
Thank you too for your amazing job.

nicolas-grekas (approved):

nicolas-grekas:
Thank you @dunglas.

Issue #680, opened 2020-09-23
dump-env shouldn't require env argument in all cases


ostrolucky:
In all of our projects, when deploying, we generate .env.local file with desired target APP_ENV inside of it. However, when running `composer dump-env prod` this is ineffective, as APP_ENV defined in .env.local file is rewritten by whatever is passed as an argument.

This causes problems because people expect whatever they define as APP_ENV to be used in resulting application after build is done, but then run into situation when they change APP_ENV into something else then `prod`, but build still runs `composer dump-env prod`, resulting in broken expectations.

I realize this argument is important in cases when there are files of format `.env.{environment}` or `.env.{environment}.local`, but it should be then required only in such cases when these files exist, otherwise argument shouldn't be needed and whatever is defined in .env.local file should be used IMHO.

I think that command should check if such files exist via glob and if so, attempt to interactively ask user to choose which one do they expect to be used. If running in non-interactive mode (answer is null), then it should throw exception.  If there are no such files, just take whatever APP_ENV is defined in other, non-environment specific files (.env, .env.local)

ostrolucky:
No feedback yet, @nicolas-grekas is this welcome to implement? I could give it a try, just tell me if this proposal is sound.

nicolas-grekas:
I'd prefer no interactive mode.
I'd be fine with making the argument optional and loading it from .env.local when possible. When .env.local is missing of doesn't contain APP_ENV, an exception should be raised.
We should not load from any other .env file IMHO (especially not `.env`, which contains the default `dev` most of the time).

nicolas-grekas:
PR welcome BTW :)

PR #720, opened 2020-11-17
Make "env" argument in symfony:dump-env optional

ostrolucky:
Addresses #680


nicolas-grekas:
Thank you @ostrolucky.

PR #721, opened 2020-11-19
Fix: Remove trailing whitespace if value is empty

OskarStark:
This is the case if someone wants to add a blank line.

```diff
# IMPORTANT: Comment
-# 
+#
# FOO_BAR=baz
```

### DIFF
--- a/src/Configurator/EnvConfigurator.php
+++ b/src/Configurator/EnvConfigurator.php
@@ -51,7 +51,11 @@ class EnvConfigurator extends AbstractConfigurator
             foreach ($vars as $key => $value) {
                 $value = $this->evaluateValue($value);
                 if ('#' === $key[0] && is_numeric(substr($key, 1))) {
-                    $data .= '# '.$value."\n";
+                    if ('' === $value) {
+                        $data .= "#\n";
+                    } else {
+                        $data .= '# '.$value."\n";
+                    }
 
                     continue;
                 }


src/Configurator/EnvConfigurator.php
```diff
@@ -51,6 +51,12 @@ private function configureEnvDist(Recipe $recipe, $vars, bool $update)
             foreach ($vars as $key => $value) {
                 $value = $this->evaluateValue($value);
                 if ('#' === $key[0] && is_numeric(substr($key, 1))) {
+                    if (empty($value)) {
```
fabpot, line 54: `if ('' === $value) {` then `$data .= "#\n";`?

fabpot (changes requested):

src/Configurator/EnvConfigurator.php
```diff
@@ -51,6 +51,12 @@ private function configureEnvDist(Recipe $recipe, $vars, bool $update)
             foreach ($vars as $key => $value) {
                 $value = $this->evaluateValue($value);
                 if ('#' === $key[0] && is_numeric(substr($key, 1))) {
+                    if ('' === $value) {
+                        $data .= "#\n";
+
+                        continue;
```
fabpot, line 57: Can we remove that "continue" and do an else to avoid the double-continue? Sorry for the nitpick

OskarStark: sure, no problem

src/Configurator/EnvConfigurator.php
```diff
@@ -51,7 +51,11 @@ private function configureEnvDist(Recipe $recipe, $vars, bool $update)
             foreach ($vars as $key => $value) {
                 $value = $this->evaluateValue($value);
                 if ('#' === $key[0] && is_numeric(substr($key, 1))) {
-                    $data .= '# '.$value."\n";
+                    if ('' === $value) {
```
OskarStark, line 54:
```suggestion
                    if ('' === trim($value)) {
```

WDYT @fabpot ?

nicolas-grekas: this won't happend IRL, and if it does, I would trust the manifest did it on purpose

OskarStark: OK, ready to merge then

chalasr (approved):

fabpot:
Thank you @OskarStark.

PR #722, opened 2020-11-19
Use GithubActions instead of Travis-CI

OskarStark:
You can see the first run here: https://github.com/OskarStark/flex/pull/1

Closes https://github.com/OskarStark/flex/pull/1


.github/workflows/ci.yaml
```diff
@@ -0,0 +1,112 @@
...
+              run: vendor/bin/simple-phpunit install
+
+            - name: "PHPUnit version"
+              run: vendor/bin/simple-phpunit --version
+
+            - name: "Run tests"
+              run: vendor/bin/simple-phpunit
+
+            - name: "Lint PHP files"
+              run: find src/ -name '*.php' | xargs -n1 php -l
```
OskarStark, line 112: @nicolas-grekas do we really need to do this for each job of the matrix, do we need this for every php version or is it enough todo it once, like we do for composer validate?

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,112 @@
...
+              run: echo "::add-matcher::${{ runner.tool_cache }}/phpunit.json"
+
+            - if: matrix.php-version == '8.0'
+              run: composer config platform.php 7.4.99 || true
+
+            - if: matrix.composer-version == '1'
+              run: composer require --no-update composer/composer:^1.0.2
+
+            - if: matrix.composer-version == '2'
+              run: composer require --no-update composer/composer:^2
```
OskarStark, line 95: Can we remove these two steps, as we are now setting up the correct composer version three steps above? 🤔

.github/workflows/ci.yaml
nicolas-grekas: why not? we need it for 7.1 for sure, is it an issue to do it on all versions?

.github/workflows/ci.yaml
nicolas-grekas: not sure what you mean here
this is about putting composer/composer in the vendor/ folder, it's unrelated to which version of composer is used to install the deps

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,112 @@
...
+                    - '7.2'
+                    - '7.3'
+                    - '7.4'
+                    - '8.0'
+                dependencies:
+                    - 'lowest'
+                    - 'highest'
+                composer-version:
+                    - '1'
+                    - '2'
```
nicolas-grekas, line 71: can't we do as in travis as not make these multiply the matrix?
I having headaches with matrices multiplication, let's same some CO2 also.

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,112 @@
...
+            matrix:
+                php-version:
+                    - '7.1'
+                    - '7.2'
+                    - '7.3'
+                    - '7.4'
+                    - '8.0'
+                dependencies:
+                    - 'lowest'
+                    - 'highest'
```
nicolas-grekas, line 68: same here about the multiplication of the matrix

.github/workflows/ci.yaml
OskarStark: >  is it an issue to do it on all versions?

No but we are doing it multiple times for one php version because of the matrix

.github/workflows/ci.yaml
OskarStark: Sure, we will use composer v2 then, oK?

.github/workflows/ci.yaml
nicolas-grekas: nice, that fits with my other comments: let's kill this matrix and make it a list

.github/workflows/ci.yaml
nicolas-grekas: OK!

.github/workflows/ci.yaml
OskarStark: Done in a separate job only once per version

.github/workflows/ci.yaml
OskarStark: Ok, so only lowest test for 7.1 ?

nicolas-grekas: yes
then merge the jobs for linting with the jobs for testing, as in travis :)

.github/workflows/ci.yaml
OskarStark: I mean we can remove these steps if we install the vendors directly using composer 1 or composer 2, right?
In this case the plugin api version is 1 or 2

.github/workflows/ci.yaml
OskarStark: If we want to run them only once per php version we need to add another variable to the matrix which also increases complexity and makes maintenance harder

nicolas-grekas: only in 7.1 is fine

OskarStark: ok

.github/workflows/ci.yaml
nicolas-grekas: `composer-plugin-api": "^1.0|^2.0"` so this cannot be a selector

.github/workflows/ci.yaml
OskarStark: done

.github/workflows/ci.yaml
OskarStark: ok but you want to test it in one workflow job instead, right?

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,113 @@
...
+                    - '7.3'
+                    - '7.4'
+                    - '8.0'
+                dependencies:
+                    - 'highest'
+                allow-failures: [false]
+                include:
+                    php-version: '7.1'
+                    dependencies: 'lowest'
+                    allow-failues: false
```
nicolas-grekas, line 72: typo: failures

.github/workflows/ci.yaml
nicolas-grekas: I don't know what this means :)
Same as in travis basically :)

OskarStark:
Ready to merge from my side @nicolas-grekas 👍

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,117 @@
...
+
+            - name: "Install PHPUnit"
+              run: vendor/bin/simple-phpunit install
+
+            - name: "PHPUnit version"
+              run: vendor/bin/simple-phpunit --version
+
+            - run: vendor/bin/simple-phpunit
+
+            - run: composer self-update --snapshot
```
nicolas-grekas, line 105: maybe we can remove this one?

nicolas-grekas (approved):

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,117 @@
...
+                    coverage: "none"
+                    php-version: "${{ matrix.php-version }}"
+                    tools: "composer:v2"
+
+            -
+                name: "Validate composer.json"
+                run: "composer validate --strict --no-check-lock"
+
+            -
+                name: "Composer install"
```
stof, line 42: why running a composer install here ?

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,117 @@
...
+    composer-validate:
+        name: "Composer validate (${{ matrix.php-version }})"
+
+        runs-on: ubuntu-16.04
+
+        strategy:
+            fail-fast: false
+            matrix:
+                php-version:
+                    - '7.4'
```
stof, line 22: I don't think we even need a matrix for a single job here. You can write `7.4` directly in the `setup-php` step (and remove it from the job name, as there is no point specifying a php version to validate the composer.json)

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,117 @@
...
+            - 'main'
+
+env:
+    SYMFONY_PHPUNIT_DIR: "$HOME/symfony-bridge/.phpunit"
+
+jobs:
+    composer-validate:
+        name: "Composer validate (${{ matrix.php-version }})"
+
+        runs-on: ubuntu-16.04
```
stof, line 16: why using an old ubuntu environment ?

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,117 @@
...
+
+    tests:
+        needs:
+            - "composer-validate"
+
+        name: "PHP ${{ matrix.php-version }} + ${{ matrix.dependencies }}"
+
+        runs-on: ubuntu-16.04
+
+        continue-on-error: ${{ matrix.allow-failures }}
```
stof, line 55: why making this a matrix property when you always use `false` for it (which is even the default behavior on github). this is only making the config more confusing IMO.

thus, `continue-on-error` is *not* the equivalent of Travis's `allow-failures`, as it also continues running further steps (so if installing PHP breaks, it will still try to run composer after that for instance)

.github/workflows/ci.yaml
```diff
@@ -0,0 +1,117 @@
...
+              with:
+                  coverage: "none"
+                  php-version: ${{ matrix.php-version }}
+                  tools: composer:v2
+
+            - name: "Add PHPUnit matcher"
+              run: echo "::add-matcher::${{ runner.tool_cache }}/phpunit.json"
+
+            - if: matrix.php-version == '8.0'
+              run: composer config platform.php 7.4.99 || true
```
stof, line 89: do we still need that now that PHP 8 is stable or are the relevant dependencies properly compatible ?

.github/workflows/ci.yaml
OskarStark: Done

.github/workflows/ci.yaml
OskarStark: its just the name, internally it uses update if `highest` or `lowest` and install if `locked`, see:

https://github.com/ramsey/composer-install#dependency-versions

I removed the name of the step like down below

.github/workflows/ci.yaml
OskarStark: I use it twice, in the name of the job and the job itself. I changed the name and do not name the version there and removed the matrix entirely.

.github/workflows/ci.yaml
OskarStark: When I migrate some things I want to use nearly the same config like on travis to ensure an error is not based on the OS version.
I updated to `ubuntu-latest` now

.github/workflows/ci.yaml
OskarStark: I don't know

.github/workflows/ci.yaml
OskarStark: I am using this config all over the place and it works for me like I expect.
I do not agree here, if you check https://github.com/OskarStark/flex/runs/1426102465?check_suite_focus=true for example, composer install step is breaking and all following steps are skipped, which is what I want to achieve:
CleanShot 2020-11-27 at 13 05 59

.github/workflows/ci.yaml
stof: but there is no point running composer at all in this step, as you never use the dependencies.

.github/workflows/ci.yaml
stof: @OskarStark you are showing a case with `continue-on-error: false` (which is the default behavior on Github and the only one used in this config).

What I'm saying is that setting `allow-failures: true` in your config (not done for now, which is why the matrix is useless for it) will *not* do the equivalent of the Travis one (and will *not* do what your screenshot shows)

.github/workflows/ci.yaml
stof: I suggest trying to remove it and see if it works.

.github/workflows/ci.yaml
OskarStark: You are right I only use `false`, I removed it 👍

.github/workflows/ci.yaml
OskarStark: Done and pushed ⏳

OskarStark: No it fails: https://github.com/OskarStark/flex/pull/1/checks?check_run_id=1463674419

I reverted the commit

stof: I opened https://github.com/composer/composer/issues/9523

.github/workflows/ci.yaml
OskarStark: But how can we use phpunit bridge then? 🙄

stof: this comment was on a separate job than the one running test. You moved the validation to the test job after that

.github/workflows/ci.yaml
OskarStark: Thanks

OskarStark: @stof I removed the step and it looks like its working 👍

OskarStark:
Ready to merge @nicolas-grekas ?

nicolas-grekas:
Thank you @OskarStark.

PR #735, opened 2021-02-01
Changing to support next version of stimulus bridge with "fetch" mode

weaverryan:
Partner to https://github.com/symfony/stimulus-bridge/pull/15 and https://github.com/symfony/ux/pull/53

The tricky part about this is that, if the user has `stimulus-bridge` 1.x, then their `controllers.json` file needs `webpackMode`. If they have `stimulus-bridge` "next" (we're thinking it'll be 2.0), then their `controllers.json` file needs `fetch`. This means that Flex has a "hard break" and sorta just needs to assume that the user has the "new" version.

We made the decision - to go from `webpackMode` to `fetch` because the setting no longer is truly the same as `webpackMode`. However, if we wanted to make the upgrade path easier, we could revert that and use `webpackMode` in 2.0 as well (instead of changing to `fetch`).

### DIFF
--- a/src/PackageJsonSynchronizer.php
+++ b/src/PackageJsonSynchronizer.php
@@ -112,7 +112,7 @@ class PackageJsonSynchronizer
                 if (!isset($previousControllersJson['controllers']['@'.$phpPackage][$controllerName])) {
                     $config = [];
                     $config['enabled'] = $defaultConfig['enabled'];
-                    $config['webpackMode'] = $defaultConfig['webpackMode'];
+                    $config['fetch'] = $defaultConfig['fetch'] ?? 'eager';
 
                     if (isset($defaultConfig['autoimport'])) {
                         $config['autoimport'] = $defaultConfig['autoimport'];
@@ -128,7 +128,7 @@ class PackageJsonSynchronizer
 
                 $config = [];
                 $config['enabled'] = $previousConfig['enabled'];
-                $config['webpackMode'] = $previousConfig['webpackMode'];
+                $config['fetch'] = $previousConfig['fetch'] ?? 'eager';
 
                 if (isset($defaultConfig['autoimport'])) {
                     $config['autoimport'] = [];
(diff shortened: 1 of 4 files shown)


stof:
if you want to avoid a hard break, you could actually set both keys in Flex. stimulus-bridge will ignore the key it does not need.

nicolas-grekas (commented):
The BC break is fine to my side.
We could maybe make flex turn existing `webpackMode` entries into `fetch`?

weaverryan:
> if you want to avoid a hard break, you could actually set both keys in Flex. stimulus-bridge will ignore the key it does not need

I've just done that 👍 https://github.com/symfony/ux/pull/53

> We could maybe make flex turn existing webpackMode entries into fetch?

👍 It indeed does this :)

tests/PackageJsonSynchronizerTest.php
```diff
@@ -20,15 +20,15 @@ class PackageJsonSynchronizerTest extends TestCase
...
+    protected function setUp(): void
     {
         $this->tempDir = sys_get_temp_dir().'/flex-package-json-'.substr(md5(uniqid('', true)), 0, 6);
         (new Filesystem())->mirror(__DIR__.'/Fixtures/packageJson', $this->tempDir);
 
         $this->synchronizer = new PackageJsonSynchronizer($this->tempDir);
     }
 
-    public function tearDown(): void
+    protected function tearDown(): void
```
weaverryan, line 31: fabbot-motivated change

src/PackageJsonSynchronizer.php
```diff
@@ -149,7 +149,7 @@ private function registerWebpackResources(array $phpPackages)
             }
         }
 
-        file_put_contents($controllersJsonPath, json_encode($newControllersJson, JSON_PRETTY_PRINT | JSON_UNESCAPED_SLASHES)."\n");
+        file_put_contents($controllersJsonPath, json_encode($newControllersJson, \JSON_PRETTY_PRINT | \JSON_UNESCAPED_SLASHES)."\n");
```
weaverryan, line 152: fabbot motivated

nicolas-grekas (approved):

tests/PackageJsonSynchronizerTest.php
fabpot: fabbot is great!

tgalopin:
Thanks @weaverryan.

PR #737, opened 2021-02-05
Only warn about synchronizing if UX packages are indeed installed

tgalopin:
Currently, the `Synchronizing package.json with PHP packages` message is always displayed as soon as a package.json file is detected in your project. This can be cumbersome for people not using UX packages.

This PR proposes to only display this message if there are indeed UX packages installed locally. It does this by detecting the "symfony-ux" keyword in the composer.json file of the package.


nicolas-grekas (approved):

nicolas-grekas (approved):

src/PackageJsonSynchronizer.php
```diff
@@ -21,15 +21,37 @@
...
+                continue;
+            }
+
+            if (!file_exists($this->rootDir.'/'.$this->vendorDirname.'/'.$packageName.'/composer.json')) {
+                continue;
+            }
+
+            $package = @json_decode(file_get_contents($this->rootDir.'/'.$this->vendorDirname.'/'.$packageName.'/composer.json'), true);
+
+            if (\is_array($package) && \in_array('symfony-ux', $package['keywords'] ?? [], true)) {
```
maxhelias, line 49: Is it possible to create a "symfony-ux" package type instead of using a keyword ?

tgalopin: It was discussed (I don't remember in which issue, @nicolas-grekas do you remember) and chose to do this way.

nicolas-grekas: Right now, we use the tag, but the plan was to turn this into a package type and to require this package type to look for controllers. Something in the TODO

stof: A package type is initially meant to control the composer installer being used, and each package can have only one such type.
A package shipping a symfony-ux component is likely to also ship other things in the code (most of the time a symfony bundle). And the management of the symfony-ux logic does not rely on a composer installer. So to me, there is no benefit to using the package type rather than a keyword. It only adds more restrictions about how you can use them in packages (and as said before, it does not change the discoverability on packagist.org as both type and keywords can be used as search filters).

Side note: Flex used to enforce a `symfony-bundle` type to register bundles in the past, and this was found to be too restrictive. Why redoing the same pattern ?

src/PackageJsonSynchronizer.php
```diff
@@ -21,15 +21,37 @@
...
+        if (!$this->rootDir || !file_exists($this->rootDir.'/package.json')) {
+            return false;
+        }
+
+        foreach ($packagesNames as $packageName) {
+            if (!$this->resolveAssetsDir($packageName)) {
+                continue;
+            }
+
+            if (!file_exists($this->rootDir.'/'.$this->vendorDirname.'/'.$packageName.'/composer.json')) {
```
stof, line 43: this is compatible only with packages installed in the default location. Why not relying on Composer to get the installation path of the package instead ?

tgalopin: `$this->vendorDirname` comes from Composer already

src/PackageJsonSynchronizer.php
nicolas-grekas: Oh, I thought you were the one advocating for using package type.
Fine to me then of course! We would still require the keyword to look for controllers, isn't it?

stof: I've always been the advocating *against* using the package type.

And yes, I think requiring the keyword makes sense. It will make sure such packages can be discovered through the keyword.

src/PackageJsonSynchronizer.php
stof: But when using custom installers, not all packages are installed in `$vendorDir/$packageName` (btw, even with the default installer, Composer has the legacy `target-dir` feature which might change that too).

tgalopin: I suggest we handle this case if someone has the issue.

nicolas-grekas:
I propose #765 instead.

nicolas-grekas:
Replaced by #765 and #766
I'm going to follow up with adding the opt-in via the tag.
Thanks for opening.

PR #738, opened 2021-02-09
Allow specifying the name of the .env file to be dumped

dsech:
symfony/dotenv allows customizing the name of the  .env files. Unfortunately, the dump-env command doesn't allow changing the name.

This PR adds a new command option through which the filename can be specified:
```
> composer dump-env prod --env-file=.env.custom
Successfully dumped .env.custom files in .env.custom.local.php
```

My specific need at the moment is that I want to reserve the usage of the .env file for docker-compose

### DIFF
--- a/src/Command/DumpEnvCommand.php
+++ b/src/Command/DumpEnvCommand.php
@@ -42,6 +42,7 @@ class DumpEnvCommand extends BaseCommand
                 new InputArgument('env', InputArgument::OPTIONAL, 'The application environment to dump .env files for - e.g. "prod".'),
             ])
             ->addOption('empty', null, InputOption::VALUE_NONE, 'Ignore the content of .env files')
+            ->addOption('env-file', null, InputOption::VALUE_REQUIRED, 'The name of the .env files to be dumped, relative to the project root dir.', '.env')
         ;
     }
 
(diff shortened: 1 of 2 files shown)


src/Command/DumpEnvCommand.php
```diff
@@ -42,6 +42,7 @@ protected function configure()
                 new InputArgument('env', InputArgument::OPTIONAL, 'The application environment to dump .env files for - e.g. "prod".'),
             ])
             ->addOption('empty', null, InputOption::VALUE_NONE, 'Ignore the content of .env files')
+            ->addOption('env-file', null, InputOption::VALUE_OPTIONAL, 'The name of the .env files to be dumped, relative to the project root dir.', '.env')
```
stof, line 45: this should be `InputOption::VALUE_REQUIRED`. There is no reason to pass `--env-file` without a value in the option.

dsech: Fixed

dsech:
The code style issue reported is not related to this change.

stof:
yeah, the requested CS change is because the new rule regarding namespacing constants has been deployed in fabbot, but the new version of the fixer has not yet been applied on the project code.

nicolas-grekas:
How does this play with flex when it needs to add entries to `.env` files?
I suppose you patched your public/index.php and bin/console files accordingly?
A better option might provided by #38465, which manages an "extra.runtime.dotenv_path" for this.

dsech:
Indeed, I've manually changed the `.env` filename in `public/index.php`, `bin/console`, and `.gitignore`, but then I got stuck because I couldn't tell `dump-env` to dump the new .env files for production.

This approach definitely breaks flex by writing to the wrong .env file. But it's not a blocking issue, as I can always copy the generated lines and move them in the correct place.

And, of course, if the new symfony/runtime component allows configuring the .env filename everywhere (including the dump env command), then it's definitely a nicer approach than what I'm trying to do here.

Should I close this PR?

PR #742, opened 2021-02-16
Require autoload.php in dump-env

### LINKED ISSUE #741: Upcoming Composer version (now in snapshot) gives errors at DotEnv dump
After running `composer self-update --snapshot`  (which gives d3ba447b79e45d1130e5523224700b97541b16de) and then running `composer dump-env dev`  Im presented with

```
  [RuntimeException]
  Please run "composer require symfony/dotenv" to load the ".env" files configuring the application.
```

After downgrading to `composer self-update --stable` (which gives 2.0.9) it then works again.

Just a heads up that its going to stop working soon.

I debugged it to the point the autoloader.php, exists, but is not actually loaded so `if (!class_exists(Dotenv::class)) {` in `vendor/symfony/flex/src/Command/DumpEnvCommand.php:88` fails.

nicolas-grekas:
Fixes #741

Let's wait for confirmation by @Seldaek before merging.

### DIFF
--- a/src/Command/DumpEnvCommand.php
+++ b/src/Command/DumpEnvCommand.php
@@ -71,7 +71,7 @@ class DumpEnvCommand extends BaseCommand
 return $vars;
 
 EOF;
-        file_put_contents($path.'.local.php', $vars, LOCK_EX);
+        file_put_contents($path.'.local.php', $vars, \LOCK_EX);
 
         $this->getIO()->writeError('Successfully dumped .env files in <info>.env.local.php</>');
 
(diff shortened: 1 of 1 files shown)


Seldaek:
OK yes go ahead and merge, I don't think we can reasonably fix this in Composer without risking major interference between Composer deps and project deps. This will still cause interference risk, but at least only for flex users..

As a side-note, it would be much cleaner if this dump-env command was part of the symfony console as `bin/console dump-env`, you wouldn't have any of these issues nor risk interfering with Composer deps.

nicolas-grekas:
released as v1.12.2

Issue #745, opened 2021-02-26
[RFC] Allow packages to bundle their respective recipes as source code instead of forcing a separate repo


gggeek:
Wouldn't it be easier both to maintain and discover recipes within the original source code, than keeping them on a separate repo/server?
Imho separate recipes should only be used for packages which are abandoned or for which the original author is against the usage of Flex...

icanhazstring:
Actually, I was looking for this as well.
Since `symfony/flex` only checks for `"type": "symfony-bundle"` this will give developers the ability to ship the recipe with the code. This is especially useful for proprietary bundle development.

For the requirements of a package to have a recipe the following criteria should be met:
- The `type` of the package is `symfony-bundle`
- A `recipe` folder must be present in the root path of the package
- All other requirements for symfony/flex must be met (like an existing `manifest.json`)

As for the priority I could think of that
1. If `<package>/recipe/manifest.json` is present use this
2. Fallback to `github.com/symfony/recipe` or `symfony/recipes-contrib`

nicolas-grekas:
See https://github.com/symfony/flex/pull/753#issuecomment-924109963

PR #753, opened 2021-04-14
PoC: Load manifest.json from local bundle

icanhazstring:
This is a PoC for #745.

This will add the ability to receive the `manifest.json` for `symfony/flex` from an installed symfony-bundle.
I did this to be able to provide a recipe for a proprietary bundle without the need to setup a recipe server.

Each bundle should have its own (already versioned) manifest.json inside a `<package>/recipe/` folder.
This should resemble the same specs as for `github.com/symfony/recipe`.

To reproduce that this is working:
1. Setup a symfony application (`symfony new flex-test`)
2. Create a small local symfony bundle with this `composer.json`

    ```
    {
      "name": "flex-test/symfony-bundle",
      "type": "symfony-bundle",
      "require": {
        "php": "^8.0",
      }
    }
    ```
    And this `manifest.json` in `<package-path>/recipe/manifest.json`

    ```
    {
      "env": {
        "SYMFONY_TEMPLATE": "magic"
      }
    }
    ```

3. Add repositories inside your `flex-test` application `composer.json` to receive the bundles and the poc
    ```
    {
      "repositories": [
        {
          "type": "path",
          "url": "<path-to-bundle>/symfony-bundle"
        },
        {
          "type": "vcs",
          "url": "https://github.com/icanhazstring/flex"
        }
      ]
    }
    ```
4. Require `symfony/flex:dev-main` and `flex-test/symfony-bundle:dev-main`
5. The `.env` file should have the following content:

    ```
    ###> flex-test/symfony-bundle ###
    SYMFONY_TEMPLATE=magic
    ###< flex-test/symfony-bundle ###
    ```

---

Feel free to give some feedback for this.

### DIFF
--- a/src/Downloader.php
+++ b/src/Downloader.php
@@ -31,6 +31,7 @@ class Downloader
     private static $MAX_LENGTH = 1000;
     private static $versions;
 
+    private $composerConfig;
     private $io;
     private $sess;
     private $cache;
@@ -49,9 +50,9 @@ class Downloader
 
         $this->endpoint = rtrim(getenv('SYMFONY_ENDPOINT') ?: ($composer->getPackage()->getExtra()['symfony']['endpoint'] ?? self::$DEFAULT_ENDPOINT), '/');
         $this->io = $io;
-        $config = $composer->getConfig();
+        $this->composerConfig = $composer->getConfig();
         $this->rfs = $rfs;
-        $this->cache = new ComposerCache($io, $config->get('cache-repo-dir').'/'.preg_replace('{[^a-z0-9.]}i', '-', $this->endpoint));
+        $this->cache = new ComposerCache($io, $this->composerConfig->get('cache-repo-dir').'/'.preg_replace('{[^a-z0-9.]}i', '-', $this->endpoint));
         $this->sess = bin2hex(random_bytes(16));
     }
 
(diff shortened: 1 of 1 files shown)


icanhazstring:
@fabpot maybe you have some feedback if we could implement this and maybe how to improve it.

This current implementation is obviously not perfect and "just a hack" to get the idea how this could work.

icanhazstring:
Thanks for the update. Not in favor of my idea, but a valid and reasonable decision 👍

MichaelBrauner:
So from a bundle builder's point of view, that would be a killer feature.

I think that just by writing a symfony bundle and not a simple PHP package, you commit yourself to a certain extent. Some bundles have a PHP library included and use it as a symfony bundle. Is this not enough?

When I install a bundle, I'm asked anyway if I want to run the associated recipe or not. With official recipes, this is even done without my intervention. So what difference does it make if the recipe is outside or inside the bundle repository?

How many bundles are there that may want to have their code used in different ways that require multiple external recipes?

Is it wise to withhold such an absolute killer feature that would obviously be easy to implement?

icanhazstring:
With flex now being able to use another source for recipes, I would recon one could write it so, that is looks inside the package and then on a server?

Maybe another PoC would be feasible now.

PR #754, opened 2021-04-17
Package JSON sync alters array formatting

HypeMC:
Currently, each time `composer install` is run, it changes the formatting of arrays keys in `package.json`:

```diff
{
-   "browserslist": [
-      "defaults"
-   ]
+   "browserslist": ["defaults"]
}
```

This is especially annoying since `npm install` actually reverts it to the previous formatting, so it's a constant "battle" between the two package managers.


nicolas-grekas (approved):

nicolas-grekas:
Thank you @HypeMC.

Issue #755, opened 2021-04-23
Existing files are not overwritten when directory is copied with CopyFromPackage


mnocon:
Hi!

I believe there's an issue with how `copy-from-package` works in recipes when directory is copied.

Given following recipe:
```
{
      "copy-from-package": {
        "testSourceDir/": "testTargetDir/",
      }
}
```

Once files exist in testTargetDir they will never be updated, even when `--force` is used when running recipes.

The behaviour I'd expect is as follows:
1) When running `composer symfony:recipes:install <package>` without `--force` the files are not overwritten (their state is not updated).
2) When running `composer symfony:recipes:install <package>` with `--force` the files are overwritten (their state is updated).

I believe this behaviour (overwriting existing files when force is used) is consistent with what's written here: https://symfony.com/doc/current/setup/upgrade_major.html#updating-recipes 
```
Instead, the updates process re-installs the latest version of the recipe which means that your custom code will be overridden completely. After updating a recipe, you need to carefully choose which changes you want, and undo the rest.
```

I've created a PR here: https://github.com/symfony/flex/pull/756

PR #756, opened 2021-04-23
Fix copy from package with directories

mnocon:
Potential fix for https://github.com/symfony/flex/issues/755

I've:
1) added a test that uncovers the issue - even when `force` is used the existing files are not overwritten
2) added a test that makes sure the behaviour does not change when force is not used
3) fixed the issue by respecting the `force` option in `copyDir` method.


nicolas-grekas:
Thank you @mnocon.

PR #758, opened 2021-04-27
Fix CS

fabpot:


stof:
I think you need to bump the min version of `symfony/phpunit-bridge`

PR #763, opened 2021-05-13
Don't try to update package.json on composer install

nicolas-grekas:
composer install should be read-only, so this doesn't make much sense and this spams the console with unneeded messages.

### DIFF
--- a/src/Flex.php
+++ b/src/Flex.php
@@ -483,7 +483,9 @@ class Flex implements PluginInterface, EventSubscriberInterface
         $this->io->writeError('');
 
         if (!$recipes) {
-            $this->synchronizePackageJson($rootDir);
+            if (ScriptEvents::POST_UPDATE_CMD === $event->getName()) {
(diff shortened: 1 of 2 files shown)


fabpot (approved):

fabpot:
Thank you @nicolas-grekas.

src/Flex.php
```diff
@@ -483,7 +483,9 @@ public function install(Event $event = null)
         $this->io->writeError('');
 
         if (!$recipes) {
-            $this->synchronizePackageJson($rootDir);
+            if (ScriptEvents::POST_UPDATE_CMD === $event->getName()) {
```
Seldaek, line 486: This is borked, $event can be null when called via https://github.com/symfony/flex/blob/ff85b54092f70c866709cd97f0f1ffa1ddca2b24/src/Flex.php#L376

I'm not sure how to best fix this sorry so will leave that to you :)

nicolas-grekas: fixed in 4961515

PR #764, opened 2021-05-13
Copy peerDependencies to the root package.json

nicolas-grekas:
As spotted in https://github.com/symfony/ux/pull/93

/cc @weaverryan

### DIFF
--- a/src/PackageJsonSynchronizer.php
+++ b/src/PackageJsonSynchronizer.php
@@ -13,6 +13,9 @@ namespace Symfony\Flex;
 
 use Composer\Json\JsonFile;
 use Composer\Json\JsonManipulator;
+use Composer\Semver\Constraint\ConstraintInterface;
+use Composer\Semver\Intervals;
+use Composer\Semver\VersionParser;
 
 /**
  * Synchronize package.json files detected in installed PHP packages with
@@ -40,6 +43,8 @@ class PackageJsonSynchronizer
         foreach ($packagesNames as $packageName) {
             $this->addPackageJsonLink($packageName);
         }
+
+        $this->registerPeerDependencies($packagesNames);
 
         // Register controllers and entrypoints in controllers.json
         $this->registerWebpackResources($packagesNames);
@@ -98,11 +103,8 @@ class PackageJsonSynchronizer
                 continue;
             }
 
-            if (!file_exists($packageJsonPath = $this->rootDir.'/vendor/'.$phpPackage.$assetsDir.'/package.json')) {
-                continue;
-            }
-
             // Register in config
+            $packageJsonPath = $this->rootDir.'/vendor/'.$phpPackage.$assetsDir.'/package.json';
             $packageJson = (new JsonFile($packageJsonPath))->read();
 
             foreach ($packageJson['symfony']['controllers'] ?? [] as $controllerName => $defaultConfig) {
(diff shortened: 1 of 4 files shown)


src/PackageJsonSynchronizer.php
```diff
@@ -150,6 +149,43 @@ private function registerWebpackResources(array $phpPackages)
...
+            return;
+        }
+
+        $manipulator = new JsonManipulator(file_get_contents($this->rootDir.'/package.json'));
+        $content = json_decode($manipulator->getContents(), true);
+        $devDependencies = $content['devDependencies'];
+
+        foreach ($peerDependencies as $peerDependency => $constraints) {
+            if (!isset($content['dependencies'][$peerDependency]) && !isset($content['devDependencies'][$peerDependency])) {
+                // FIXME: we should compute the intersection of all constraints, including any pre-existing one
```
nicolas-grekas, line 179: this should not be a blocker for this PR (as this is hard to implement and is not mandatory to provide a nice DX improvement)

src/PackageJsonSynchronizer.php
```diff
@@ -98,11 +100,8 @@ private function registerWebpackResources(array $phpPackages)
                 continue;
             }
 
-            if (!file_exists($packageJsonPath = $this->rootDir.'/vendor/'.$phpPackage.$assetsDir.'/package.json')) {
-                continue;
```
stof, line 102: Why removing this ?

src/PackageJsonSynchronizer.php
```diff
@@ -150,6 +149,41 @@ private function registerWebpackResources(array $phpPackages)
...
+    {
+        $peerDependencies = [];
+
+        foreach ($phpPackages as $phpPackage) {
+            if (!$assetsDir = $this->resolveAssetsDir($phpPackage)) {
+                continue;
+            }
+
+            $packageJsonPath = $this->rootDir.'/vendor/'.$phpPackage.$assetsDir.'/package.json';
+            $packageJson = (new JsonFile($packageJsonPath))->read();
```
weaverryan, line 162: These 2 lines could be moved into a private method - they're repeated elsewhere

src/PackageJsonSynchronizer.php
```diff
@@ -150,6 +149,41 @@ private function registerWebpackResources(array $phpPackages)
...
+            }
+        }
+
+        if (!$peerDependencies) {
+            return;
+        }
+
+        $manipulator = new JsonManipulator(file_get_contents($this->rootDir.'/package.json'));
+        $content = json_decode($manipulator->getContents(), true);
+        $devDependencies = $content['devDependencies'];
```
weaverryan, line 175:
```suggestion
        $devDependencies = $content['devDependencies'] ?? [];
```

Just to be safe

tests/PackageJsonSynchronizerTest.php
```diff
@@ -70,6 +70,8 @@ public function testSynchronizeExistingPackage()
             [
                 'name' => 'symfony/fixture',
                 'devDependencies' => [
+                    '@hotcookies' => '^1.1',
+                    '@hotdogs' => '^2',
```
weaverryan, line 74: 🌭

weaverryan (commented):
This is a really nice idea - thank you!

src/PackageJsonSynchronizer.php
nicolas-grekas: Because it's useless, we're sure the file exists if we reach this line.

tests/PackageJsonSynchronizerTest.php
nicolas-grekas: 🍪

src/PackageJsonSynchronizer.php
nicolas-grekas: I'm going to keep them as is, at least for now. The private method would need to return two variables, that's not worth the dedup IMHO.

nicolas-grekas: Now part of #767

PR #766, opened 2021-05-18
Use Composer vendor-dir

nicolas-grekas:
Extracted from #737


stof:
Things could be even better by using the installation dir of each package instead of hardcoding this as `$this->vendorDirname.'/'.$phpPackage`. this way, it would be compatible with custom installers too.

nicolas-grekas:
Indeed. I'm still going to merge as is to move forward as I won't be able to work more on this.
If anyone hits the limitation, PR welcome as usual!

nicolas-grekas:
Thank you @tgalopin.

PR #776, opened 2021-06-10
feat: allow to enable Docker support using an env var

dunglas:
Currently, it's not possible to execute the Docker-related configurators when running `composer create-project` because of a chicken and egg problem: the `composer.json` file doesn't exist yet, so it's not possible to set `extra.symfony.docker` to `true`. This PR allows to enable the Docker support by setting an environment variable called `SYMFONY_DOCKER`.

Needed by https://github.com/dunglas/symfony-docker/pull/152.

### DIFF
--- a/src/Configurator/DockerComposeConfigurator.php
+++ b/src/Configurator/DockerComposeConfigurator.php
@@ -36,7 +36,7 @@ class DockerComposeConfigurator extends AbstractConfigurator
 
     public function configure(Recipe $recipe, $config, Lock $lock, array $options = [])
     {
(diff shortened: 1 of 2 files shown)


maxhelias (approved):

nicolas-grekas (changes requested):
Might not be needed: create-project with --no-plugins, then patching, then `composer i` might do it in the Dockerfile.
Let us know if yes :)

dunglas:
Good idea @nicolas-grekas! It works.

PR #779, opened 2021-08-02
do not try to write a cache entry if there is no cache key

themasch:
Writing to the cache with an empty key will fail with "failed to open stream: Is a directory", so do not try to do that.

We noticed this when cloudflare - or the backend service - responded with a "last-modified" header for our CI servers (AWS) but not for our local system. This triggered the condition to become true and it tries to write a cache file without a filename.

I hope this is the correct place to fix it, but from what I see, this code should not try to write a cache with am empty key, correct?

Here is how a `composer install` failed for us (because of this?): 
```
  [ErrorException]                                                                                              
  file_put_contents(/tmp/composer/cache/repo/https---flex.symfony.com/): failed to open stream: Is a directory  
```

### DIFF
--- a/src/Downloader.php
+++ b/src/Downloader.php
@@ -293,7 +293,7 @@ class Downloader
         }
 
         $response = new Response($data, $lastHeaders);
-        if ($response->getHeader('last-modified')) {
+        if ($response->getHeader('last-modified') && $cacheKey) {
             $this->cache->write($cacheKey, json_encode($response));
         }
 


skripka:
We are seeing same issue. @themasch thanks for the fix!

tmarly:
same pb on my side, thanks @themasch 

As a workaround, waiting this PR to be included in master or for older versions, I'm using the `--no-cache` option with `composer install`

themasch:
Can confirm that disabling the cache (--no-cache or COMPOSER_CACHE_DIR=/dev/null) works as a workaround.

fabpot:
Thank you @themasch.

fabpot:
Just released a new version with this fix.

themasch:
Thanks for the quick release!

PR #781, opened 2021-08-20
Install recipes for packs before any others

nicolas-grekas:


fabpot (approved):

fabpot:
Thank you @nicolas-grekas.

PR #788, opened 2021-08-28
Automatically create Docker Compose files when needed

fabpot:
If you want to use Docker support in Flex, you must configure it explicitly in a project (via `symfony composer config extra.symfony.docker true` for instance). But that's not enough. You must also create the `docker-compose.yaml` file. And if you do not also create the `docker-compose.override.yaml` file, you will end up with half of the configuration.

In this PR, I propose to create those 2 files automatically if they don't exist when a service should be added by a recipe.

That way, using Docker support is "only" a matter of opting-in via `composer.json`.

/cc @dunglas

### DIFF
--- a/src/Configurator/DockerComposeConfigurator.php
+++ b/src/Configurator/DockerComposeConfigurator.php
@@ -43,16 +43,20 @@ class DockerComposeConfigurator extends AbstractConfigurator
 
         $rootDir = $this->options->get('root-dir');
         foreach ($this->normalizeConfig($config) as $file => $extra) {
-            if (
-                (null === $dockerComposeFile = $this->findDockerComposeFile($rootDir, $file)) ||
-                $this->isFileMarked($recipe, $dockerComposeFile)
-            ) {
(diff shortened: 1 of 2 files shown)


src/Configurator/DockerComposeConfigurator.php
```diff
@@ -43,16 +43,20 @@ public function configure(Recipe $recipe, $config, Lock $lock, array $options =
...
+                $this->write(sprintf('  Created <fg=green>"%s"</>', $dockerComposeFileName));
+            }
+            if ($this->isFileMarked($recipe, $dockerComposeFile)) {
                 continue;
             }
 
             $this->write(sprintf('Adding Docker Compose definitions to "%s"', $dockerComposeFile));
 
-            $offset = 8;
+            $offset = 2;
```
fabpot, line 59: The default indentation for Docker Compose file is 2 spaces. This was not caught before as tests always used an existing file.

dunglas (approved):

PR #794, opened 2021-09-06
Skip doing anything on "install"

### LINKED ISSUE #669: composer install changes symfony.lock on deploy without vendor directory
We've noticed that a deployment process from scratch (i.e. without a vendor directory) with `composer install` adds new data to our symfony.lock. I was under the impression that lock-files (composer.lock and symfony.lock) were regarded as read-only during a `composer install` - is the fact that the symfony.lock is being changed a bug or a feature? And if the latter, where would I find documentation for that behaviour?

nicolas-grekas:
Fix #669

### DIFF
--- a/src/Flex.php
+++ b/src/Flex.php
@@ -407,7 +407,7 @@ class Flex implements PluginInterface, EventSubscriberInterface
         }
 
         // Execute missing recipes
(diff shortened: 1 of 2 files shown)


fabpot:
Thank you @nicolas-grekas.

PR #795, opened 2021-09-06
Add --updated flag to recipes:install, default to reinstalling all recipes

nicolas-grekas:
Replaces #702, see discussion there.


fabpot:
Too bad it has been merged without taking my comment into account. Adding one more random flag name on top of something that is already unclear is not going to help. Just my 2cts.

nicolas-grekas:
I thought replacing the previous `--updated` (which changed the default behavior) with `--reset` was what we discussed (to keep the default behavior, which makes sense when updating)...

nicolas-grekas:
(If you mean changing the name of `--force` for something better, I'd prefer waiting for more work to happen on a future `recipes:update` before annoying ppl with a deprecation).

PR #797, opened 2021-09-06
Fixed directory separator trimming in vendors path

### LINKED ISSUE #796: Vendor's directory separator is not trimmed on Windows
Affected line:
https://github.com/symfony/flex/blob/main/src/Flex.php#L534

Relative path to the vendor is outputed as `vendor/`, and should be trimmed from `/` but `DIRECTORY_SEPARATOR` on Windows is defined as `\`. 

So when `package.json` is synchronized, resulted dependency path has duplicated slash like:
`file:vendor//symfony/ux-turbo/Resources/assets`

skmedix:
Fixes #796 

We don't need to use `\DIRECTORY_SEPARATOR` here, as [**symfony/filesystem**](https://github.com/symfony/filesystem/blob/4.4/Filesystem.php#L451) already normalizes windows directory separator to use forward slash.

### DIFF
--- a/src/Flex.php
+++ b/src/Flex.php
@@ -531,7 +531,7 @@ class Flex implements PluginInterface, EventSubscriberInterface
     private function synchronizePackageJson(string $rootDir)
     {
         $rootDir = realpath($rootDir);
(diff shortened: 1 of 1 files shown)


nicolas-grekas:
Thank you @skmedix.

PR #798, opened 2021-09-13
removed unnecessary code

### LINKED ISSUE #775: Remove runtime check of the PHP version from autoload
These checks are now part of Composer (since https://github.com/composer/composer/pull/8546), so we can revert #576 from Flex.

For ref: https://github.com/composer/composer/issues/9967#issuecomment-858462117

Relates to #708 as well

EhEhRon91:
Fixes #775

Removed code that was implemented in #576 as this is no longer needed as per the issue reported.

First PR in this repository. Hoping to contribute to more! If there are any issues, let me know.

### DIFF
--- a/src/Flex.php
+++ b/src/Flex.php
@@ -307,17 +307,6 @@ class Flex implements PluginInterface, EventSubscriberInterface
         return $backtrace;
     }
 
-    public function lockPlatform()
-    {
-        if (!$this->downloader->isEnabled()) {
-            return; // "symfony/flex" not found in the root composer.json - don't create the symfony.lock file
-        }
-
-        $this->lock->set('php', [
-            'version' => $this->config->get('platform')['php'] ?? (\PHP_MAJOR_VERSION.'.'.\PHP_MINOR_VERSION),
-        ]);
-    }
-
     public function configureProject(Event $event)
     {
         if (!$this->downloader->isEnabled()) {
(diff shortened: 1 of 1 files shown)


src/Flex.php
```diff
@@ -1010,14 +965,13 @@ public static function getSubscribedEvents(): array
             ScriptEvents::POST_INSTALL_CMD => 'install',
             ScriptEvents::PRE_UPDATE_CMD => 'configureInstaller',
             ScriptEvents::POST_UPDATE_CMD => 'update',
-            ScriptEvents::POST_AUTOLOAD_DUMP => 'updateAutoloadFile',
-            'auto-scripts' => 'executeAutoScripts',
```
stof, line 1014: this change looks wrong to me. It will probably break things like `composer run auto-scripts`

src/Flex.php
```diff
@@ -1010,14 +965,13 @@ public static function getSubscribedEvents(): array
...
+            ScriptEvents::POST_AUTOLOAD_DUMP => [['auto-scripts'], ['executeAutoScripts']],
```
stof, line 968: This change looks wrong to me. There is no `auto-scripts` method. And we were not executing auto scripts on `POST_AUTOLOAD_DUMP`

EhEhRon91: ```auto-scripts``` should now map to ```executeAutoScripts``` method and removed ``` ScriptEvents::POST_AUTOLOAD_DUMP``` as this was not there prior to the change in #576. If there are any issues with this, let me know!

nicolas-grekas (approved):

PR #807, opened 2021-09-21
Avoid reverting the composer.json when an auto-script fails

### LINKED ISSUE #751: If the `cache:clear` command fails after an installation, the content of `composer.json` is reverted
**Steps to reproduce**

```
symfony new test-hwi
cd test-hwi
composer require hwi/oauth-bundle symfony/http-client nyholm/psr7 guzzlehttp/promises php-http/httplug-bundle
```

Execute the community recipes when asked.
The `cache:clear` commands fails as expected (`hwi/oauth-bundle` requires extra manual configuration). The recipes files are correctly generated, however the installed packages are removed from `composer.json`.
Flex will then remove the related config files when the next `composer require` or `composer update` command will occur.

**Possible fix**

An error occurring during `cache:clear` shouldn't revert the content of `composer.json`.

Similar to https://github.com/symfony/flex/issues/142.

nicolas-grekas:
Fix #751

### DIFF
--- a/src/Flex.php
+++ b/src/Flex.php
@@ -20,6 +20,7 @@ use Composer\DependencyResolver\Operation\UpdateOperation;
 use Composer\DependencyResolver\Pool;
 use Composer\Downloader\FileDownloader;
 use Composer\EventDispatcher\EventSubscriberInterface;
+use Composer\EventDispatcher\ScriptExecutionException;
 use Composer\Factory;
 use Composer\Installer;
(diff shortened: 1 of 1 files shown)


stof:
shouldn't we at least display the error message ?

nicolas-grekas:
The error says that the command failed, but that's not really useful. I expect failing commands to display something on their own instead.

nicolas-grekas:
Closing in favor of https://github.com/composer/composer/pull/10118

PR #810, opened 2021-09-24
Use the extra.runtime.test_envs composer variabels in the symfony:dump-env command

cmodijk:
Added the option to read the `extra.runtime.test_envs` to load the test envs during the environment dump.

### DIFF
--- a/src/Command/DumpEnvCommand.php
+++ b/src/Command/DumpEnvCommand.php
@@ -13,6 +13,7 @@ namespace Symfony\Flex\Command;
 
 use Composer\Command\BaseCommand;
 use Composer\Config;
+use Composer\Factory;
 use Symfony\Component\Console\Input\InputArgument;
 use Symfony\Component\Console\Input\InputInterface;
 use Symfony\Component\Console\Input\InputOption;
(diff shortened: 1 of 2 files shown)


src/Command/DumpEnvCommand.php
```diff
@@ -112,7 +113,13 @@ private function loadEnv(string $path, ?string $env): array
             }
 
             if (method_exists($dotenv, 'loadEnv')) {
-                $dotenv->loadEnv($path);
+                $config = @json_decode(file_get_contents(Factory::getComposerFile()), true);
+                $dotenv->loadEnv(
```
nicolas-grekas, line 117: should be on one line

nicolas-grekas (commented):
Can you please have a look at failures? I guess we need to feature test to add the argument only when a recent enough version of Dotenv is installed.

tests/Command/DumpEnvCommandTest.php
```diff
@@ -164,6 +164,46 @@ public function testDoesNotRequireToSpecifyEnvArgumentWhenLocalFileIsPresent()
...
+        file_put_contents($env, 'APP_ENV=dev');
+        $envContent = <<<EOF
+APP_ENV=test
+APP_SECRET=abcdefgh123456789
+EOF;
+        file_put_contents($envLocal, $envContent);
+
+        copy($composer, FLEX_TEST_DIR.'/composer-backup.json');
+        $composerContent = @json_decode(file_get_contents($composer), true);
+        $composerContent['extra']['runtime']['test_envs'] = [];
```
nicolas-grekas, line 185: I think this test case tests nothing right now.
I order for it to do anything, a `.env.test` file should be added to me.

nicolas-grekas: oh nvm, `.env.local` is the one

PR #817, opened 2021-10-01
Use default endpoints as fallbacks if Composer endpoint is specified as string

### LINKED ISSUE #816: Sync behaviour of extra.symfony.endpoint with SYMFONY_ENDPOINT env var
PR #803 says the following:

    The default endpoints can be changed by setting the `extra.symfony.endpoint` entry in the root
    composer.json file. The special `flex://defaults` endpoint can be used in that list to reference
    the default endpoints.

    On top of that, the `SYMFONY_ENDPOINT` env var can be defined to a single endpoint. This endpoint
    will be used on top of (with higher priority than) the default endpoints of the app.

It would be nice if the `extra.symfony.endpoint` config would behave the same as the env var when using a string endpoint. Specifically, it would be nice if config like this:

```
"extra": {
    "symfony": {
        "endpoint": "https://api.github.com/...",
        "allow-contrib": true,
        "require": "5.3.*"
    }
}
```

would automatically use `flex://defaults` as the fallback endpoint.

This would help eliminate potential unintended disabling of the default Flex endpoint if one would configure Flex like this.

In that case, specifying the `endpoint` config as an array would be used to disable usage of the default endpoint if needed.

emodric:
Fixes #816

### DIFF
--- a/src/Downloader.php
+++ b/src/Downloader.php
@@ -65,6 +65,9 @@ class Downloader
             }
         } elseif (\is_array($endpoint) || false !== strpos($endpoint, '.json') || 'flex://defaults' === $endpoint) {
             $this->endpoints = array_values((array) $endpoint);
+            if (\is_string($endpoint) && false !== strpos($endpoint, '.json')) {
+                $this->endpoints[] = 'flex://defaults';
+            }
         } else {
             $this->legacyEndpoint = rtrim($endpoint, '/');
         }


src/Downloader.php
```diff
@@ -59,12 +59,17 @@ public function __construct(Composer $composer, IoInterface $io, $rfs)
...
+            if (!$this->isServerlessEnabled()) {
                 $this->endpoints = null;
                 $this->legacyEndpoint = 'https://flex.symfony.com';
             }
         } elseif (\is_array($endpoint) || false !== strpos($endpoint, '.json') || 'flex://defaults' === $endpoint) {
             $this->endpoints = array_values((array) $endpoint);
+            if (false !== strpos($endpoint, '.json')) {
+                $this->endpoints[] = 'flex://defaults';
+            }
+        } elseif ($this->isServerlessEnabled() && \is_string($endpoint)) {
```
nicolas-grekas, line 71: I think this case should be removed since when this is a string and it doesn't contain `.json`, the it means a legacy endpoint is configured.
If confirmed, then let's remove the `isServerlessEnabled` method.

emodric: It does mean that, but this PR is exactly about that.

If we use a string which does not end in `.json` in `extra.symfony.endpoint` like this:

```
"extra": {
    "symfony": {
        "endpoint": "https://api.github.com/...",
        "allow-contrib": true,
        "require": "5.3.*"
    }
}
```

Flex will consider it a legacy endpoint and serverless will not work. My goal is to have the above config work even in serverless mode (with the fallback to `flex://defaults`), which is activated by the env var. If the env var is present, only then should this be considered a legacy endpoint.

I hope I'm making myself clear enough :)

emodric: We can remove the condition if you do not wish to support any string (even ones without `.json`) as serverless endpoints, however, this will in time be supported implicitly once you remove the legacy mode completely.

nicolas-grekas: we're going to remove the FLEX_SERVERLESS env var in a few weeks, so better drop this now

emodric: Done!

nicolas-grekas (approved):

nicolas-grekas:
Thank you @emodric.

PR #821, opened 2021-10-13
feat: enable Docker support by default

dunglas:
Docker and Docker Compose configurators now have been extensively tested, and there are many recipes with Docker support available. I think that it's time to enable this feature by default.

Note: the configurations are executed only if the `Dockerfile` and the `docker-compose.*` files are present in the project.


fabpot:
Thank you @dunglas.

PR #832, opened 2021-11-22
Allow Symfony 6

nicolas-grekas:


composer.json
```diff
@@ -12,14 +12,14 @@
     "minimum-stability": "dev",
     "require": {
         "php": ">=7.1",
-        "composer-plugin-api": "^1.0|^2.0"
+        "composer-plugin-api": ">=1.0"
```
stof, line 15: Do we actually want to do this ? This seems dangerous regarding changes in the composer plugin API (and using an incompatible plugin in composer can break you ability to upgrade...)

tests/UnpackerTest.php
```diff
@@ -40,6 +40,9 @@ public function testDoNotDuplicateEntry(): void
         file_put_contents($composerJsonPath, '{}');
 
         $originalEnvComposer = getenv('COMPOSER');
+        $originalEnvComposer = $_SERVER['COMPOSER'];
```
stof, line 43: should it be a different variable name ?

composer.json
nicolas-grekas: ok, reverting, we might decide to widen later on if needed

tests/UnpackerTest.php
nicolas-grekas: no need to, esp for a test file

composer.json
nicolas-grekas: Reverted in https://github.com/symfony/flex/commit/e2ea724773cc7c391e0d163967675acd31ee04c7

tests/UnpackerTest.php
stof: then the previous line is dead code.

PR #833, opened 2021-11-22
Bump flex to v2, composer >= 2.1, php >= 8.0

nicolas-grekas:
Let's drop the legacy since Symfony 6.0 will require composer >= 2.1 anyway.


Nyholm:
Except for "remove legacy code", is there reason to release a new major version?

weaverryan:
Should the branch be called `2.x`?

nicolas-grekas:
Since Symfony 6 is already requiring composer >= 2.1, I think it's just the right timing to fade out this legacy code yes. Eg I don't want to maintain the CurlDownloader anylonger, and keeping it alive in maintained version is a :bomb: waiting to explode on my free time :) The earlier the better.

nicolas-grekas:
> Should the branch be called 2.x?

then we'd drop `main`? could be

src/Command/RecipesCommand.php
```diff
@@ -30,10 +30,10 @@ class RecipesCommand extends BaseCommand
     /** @var \Symfony\Flex\Flex */
     private $flex;
 
-    private $symfonyLock;
-    private $downloader;
+    private Lock $symfonyLock;
```
OskarStark, line 33: No property constructor promotion?

nicolas-grekas: No intention to do this in this PR: I prefer keeping the diff focused.
Could come later on as part of a code modernization PR.

weaverryan:
> then we'd drop main? could be

That's what I was thinking. The `main` branch feels more and more of an "odd ball" to me compared to `1.x`, `2.x` branches anyways.

nicolas-grekas:
`main` dropped, welcome `2.x` :)

src/Command/RecipesCommand.php
OskarStark: Fine for me 👌🏻

GromNaN:
How long do you think the branch 1.x should be supported? PHP 7.4 EOL is in 1 year, and Symfony 5.4 EOL (with PHP 7.2 support) is in 4 years.

If we want to push adoption in order to not maintain legacy code, I think PHP requirements should stay to 7.4 max for now. PHP 8 has several breaking changes that make some projects hard to update.

src/Flex.php
```diff
@@ -267,19 +191,12 @@ class_exists(__NAMESPACE__.str_replace('/', '\\', substr($file, \strlen(__DIR__)
                 BasePackage::$stabilities['dev'] = 1 + BasePackage::STABILITY_STABLE;
             }
 
-            if ($populateRepoCacheDir && isset(self::$repoReadingCommands[$command]) && ('install' !== $command || (file_exists($composerFile = Factory::getComposerFile()) && !file_exists(substr($composerFile, 0, -4).'lock')))) {
-                $this->populateRepoCacheDir();
-            }
-
             $app->add(new Command\RequireCommand($resolver, \Closure::fromCallable([$this, 'updateComposerLock'])));
             $app->add(new Command\UpdateCommand($resolver));
```
stof, line 178: Should those registration be changed to use the CommandProvider of the Composer 1.1+ plugin API instead of relying on the backtrace hack to access the Application ?

nicolas-grekas:
> How long do you think the branch 1.x should be supported? PHP 7.4 EOL is in 1 year, and Symfony 5.4 EOL (with PHP 7.2 support) is in 4 years.

With 5.4 the latest, earlier if possible. [Stats](https://packagist.org/packages/symfony/flex/stats#major/1) might help decide.
We should continue merging new features in `1.x` of course, since flex is the way to push better DX to our ecosystem.

> If we want to push adoption in order to not maintain legacy code, I think PHP requirements should stay to 7.4 max for now. PHP 8 has several breaking changes that make some projects hard to update.

Symfony 6 is already for PHP8, there is no point in having flex use different versions than both 5.4 and 6.0 to me.

src/Flex.php
nicolas-grekas: I just had a look and I think there are little benefits in doing so. Limiting merge conflicts with `1.x` looks more important to me right now.

PR #835, opened 2021-11-25
Remove unnecessary space after love / star character

herndlm:
Please excuse my fussiness but this started to annoy me after I noticed it :D


nicolas-grekas:
Haha good catch
Unfortunately this is on purpose because some terminals don't display correctly wide emojis and the space is for them.

herndlm:
> Haha good catch Unfortunately this is on purpose because some terminals don't display correctly wide emojis and the space is for them.

Nooooo :/

I thought it was strange that nobody "noticed" this for 4 years. Ok, thx for the update and sorry for the noise then :)

PR #838, opened 2021-11-26
[Flex] Fix Composer update exception when flex://defaults missing

### LINKED ISSUE #836: Composer update exception when flex://defaults missing
When the `extra.symfony.endpoint` array in `composer.json` does not include `flex://defaults`, `composer update` crashes with:

> [ErrorException] Undefined array key "splits"

This happens in [line 125 of PackageFilter.php](https://github.com/symfony/flex/blob/99e3323cca175690ec6b03353b346006298b5a0d/src/PackageFilter.php#L125).

I'm using Flex v1.17.5.

gnito-org:
Q | A
-- | --
Branch? | 1.x
Bug fix? | yes
New feature? | no
Deprecations? | no
Tickets | Fix #836
License | MIT
Doc PR | -

### DIFF
--- a/src/Cache.php
+++ b/src/Cache.php
@@ -130,6 +130,9 @@ class Cache extends BaseCache
         $this->downloader = null;
         $okVersions = [];
 
+        if (!isset($versions['splits'])) {
+            throw new \LogicException('The Flex index is missing a "splits" entry. Did you forget to add "flex://defaults" in the "extra.symfony.endpoint" array of your composer.json?');
+        }
         foreach ($versions['splits'] as $name => $vers) {
             foreach ($vers as $i => $v) {
                 if (!isset($okVersions[$v])) {
--- a/src/PackageFilter.php
+++ b/src/PackageFilter.php
@@ -122,6 +122,9 @@ class PackageFilter
         $this->downloader = null;
         $okVersions = [];
 
+        if (!isset($versions['splits'])) {
+            throw new \LogicException('The Flex index is missing a "splits" entry. Did you forget to add "flex://defaults" in the "extra.symfony.endpoint" array of your composer.json?');
+        }
         foreach ($versions['splits'] as $name => $vers) {
             foreach ($vers as $i => $v) {
                 if (!isset($okVersions[$v])) {


nicolas-grekas:
Hmm, on the other side, an endpoint that doesn't declare splits is likely broken as that means the `extra.symfony.require` entry will not work.
Maybe we should throw instead?

gnito-org:
Which specific exception would be the most appropriate to throw?

nicolas-grekas:
A LogicException would be fine to me.

gnito-org:
I changed the PR to throw a LogicException.

src/Cache.php
```diff
@@ -130,6 +130,9 @@ private function getVersions(): array
         $this->downloader = null;
         $okVersions = [];
 
+        if (!isset($versions['splits'])) {
+            throw new \LogicException('It appears that the flex://defaults entry is missing in the extra.symfony.endpoint array of your composer.json.');
```
nicolas-grekas: The message should tell that the endpoint is missing a "splits", which is required to make the `extra.symfony.require` config option work.

gnito-org: I'm not sure I understand why you're saying that.

As per [#836](https://github.com/symfony/flex/issues/836), the error occurs when the extra.symfony.endpoint array in composer.json does not include flex://defaults. My ``composer.json`` has never had a ``splits``.

If I have a ``flex://defaults`` array entry, then the ``$versions['splits']`` entry is set. Remove the ``flex://defaults`` array entry and ``$versions['splits']`` is not set.

gnito-org: If you want to give me the exact wording of the exception, then I will change the message accordingly. 

Unfortunately I cannot formulate the message when I don't understand what you're referring to.

nicolas-grekas: Yeah sure, I was surprised to read your sentence also, but now I get why you phrased it this way. Here is my proposal:

`The Flex index is missing a "splits" entry. Did you forget to add "flex://defaults" in the "extra.symfony.endpoint" array of your composer.json?`

gnito-org: Thank you! I rephrased the excepetion message as recommended.

nicolas-grekas:
Thank you @gnito-org.

PR #841, opened 2021-11-30
Refactor PackageJsonSynchronizer to prevent unintentional duplicate dependencies

### LINKED ISSUE #840: Duplicate package.json dependencies
👋

I've been experimenting with creating my own Symfony UX-like bundle ([this one](https://github.com/onlinq/form-collection-bundle)). Although the Flex documentation is limited it's working great.

One problem I've been encountering is that on every update of Composer, Flex tries to add my bundle's autoimport libraries to the `devDependencies` section of the package.json. Awesome, of course, but in this case I already added it to the `dependencies` section myself.

My request is that Flex checks if the library is already defined under the `dependencies` section before it adds it to the `devDependencies` section. I can make a PR myself if it helps!

codedmonkey:
Fixes https://github.com/symfony/flex/issues/840

Changes how `PackageJsonSynchronizer` handles dependency resolving. If a dependency is already defined under the `dependencies` section of package.json, Flex won't add it again under `devDependencies`. If multiple UX bundles require the same dependency, no action is performed.

Because the way `PackageJsonSynchronizer` was written, I had to change the order of the steps in the synchronization process. While it used to update the package.json for each UX dependency individually, it now resolves a list of dependencies and updates package.json at the end.

I also changed how it handles incompatible peer dependencies, no action is performed by the synchronizer if multiple ux package require incompatible peer dependencies.

### DIFF
--- a/src/PackageJsonSynchronizer.php
+++ b/src/PackageJsonSynchronizer.php
@@ -13,9 +13,6 @@ namespace Symfony\Flex;
 
 use Composer\Json\JsonFile;
 use Composer\Json\JsonManipulator;
-use Composer\Semver\Constraint\ConstraintInterface;
-use Composer\Semver\Intervals;
-use Composer\Semver\VersionParser;
 use Seld\JsonLint\ParsingException;
 
 /**
@@ -40,13 +37,16 @@ class PackageJsonSynchronizer
 
     public function synchronize(array $phpPackages): bool
     {
-        // Remove all links and add again only the existing packages
         try {
-            $didAddLink = $this->removePackageJsonLinks((new JsonFile($this->rootDir.'/package.json'))->read());
+            JsonFile::parseJson(file_get_contents($this->rootDir.'/package.json'));
         } catch (ParsingException $e) {
             // if package.json is invalid (possible during a recipe upgrade), we can't update the file
             return false;
         }
+
+        $didChangePackageJson = $this->removeObsoletePackageJsonLinks();
+
+        $dependencies = [];
 
         foreach ($phpPackages as $k => $phpPackage) {
             if (\is_string($phpPackage)) {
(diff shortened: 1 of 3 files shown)


src/PackageJsonSynchronizer.php
```diff
@@ -76,34 +84,77 @@ private function removePackageJsonLinks(array $packageJson): bool
...
         if (!$packageJson = $this->resolvePackageJson($phpPackage)) {
-            return false;
+            return $dependencies;
         }
 
-        $manipulator = new JsonManipulator(file_get_contents($this->rootDir.'/package.json'));
-        $manipulator->addSubNode('devDependencies', '@'.$phpPackage['name'], 'file:'.substr($packageJson->getPath(), 1 + \strlen($this->rootDir), -13));
+        $dependencies['@'.$phpPackage['name']] = 'file:'.substr($packageJson->getPath(), 1 + \strlen($this->rootDir), -13);
+
+        $versionParser = new VersionParser();
```
stof: Relying on the Composer version parser to parse npm constraints looks wrong to me. The syntax is not the same between both ecosystems (some parts of the syntax are common, but not everything)

stof: and worse than that. Some syntax is the same with different meaning (so the parser would accept it but produce a different result)

codedmonkey: I agree it's a problem. In the current version it simply generates an invalid package.json the developer has to fix themselves. This hasn't been a common use case for Flex yet so it rarely occurs. What is the intended behavior?

weaverryan: Can you explain the problem a bit more that this is solving? Is this if you install 2 packages that both have `chart.js` has a peer dependency... but they one requires v2 and another requires v3?

If so, yea... it's super rare. Well, probably not a reality at all right now. I would vote to simply "do nothing". What I mean is, if you already had `chart.js` in your package.json, leave it with whatever version was there. This is a problem the user will need to resolve and they will see peer dependency warnings from yarn/npm.

codedmonkey: I wasn't trying to solve it but while writing the PR I noticed it simply generated invalid JSON with 2 conflicting peer dependencies.

Apperantly I forgot the use case for peer dependencies in NPM so awesome suggestion. Thanks :)

codedmonkey:
Apologies for the delay, I had a very persistent cold the last few weeks.

nicolas-grekas:
Can you please rebase?
There is a conflict to resolve after https://github.com/symfony/flex/pull/845/files#diff-58c55c0fe84f0796c800aca4a49066dc4b12ce3bfb512ce88fe541fad0fe6209

nicolas-grekas (approved):

nicolas-grekas:
Thank you @codedmonkey.

Issue #830, opened 2021-11-21
File name too long


cjunge:
Just updated to Symfony Flex 1.17.2 and trying to run `composer recipes` I keep getting:

```bash
$ composer recipes
                                                                                                                                                                                                                                                           
  [ErrorException]                                                                                                                                                                                                                                         
  file_put_contents(/home/cameron/.cache/composer/repo/flex/https---api.github.com-repos-symfony-recipes-contrib-contents-friendsofsymfony.elastica-bundle.5.0.json-ref-flex-main.619ac205ee82c4.90304062.tmp): failed to open stream: File name too long 
```

Xubuntu 21.04, ext4, composer 2.1.12

The cache-key is generated from the URL, which needs to be shortened.

```bash
$ composer recipes -vv

                                                                                                                                                                                                                                                           
  [ErrorException]                                                                                                                                                                                                                                         
  file_put_contents(/home/cameron/.cache/composer/repo/flex/https---api.github.com-repos-symfony-recipes-contrib-contents-friendsofsymfony.elastica-bundle.5.0.json-ref-flex-main.619ac3b00e5ba8.11438895.tmp): failed to open stream: File name too long  
                                                                                                                                                                                                                                                           

Exception trace:
 () at phar:///home/cameron/bin/composer/src/Composer/Cache.php:152
 Composer\Util\ErrorHandler::handle() at n/a:n/a
 file_put_contents() at phar:///home/cameron/bin/composer/src/Composer/Cache.php:152
 Composer\Cache->write() at /home/cameron/Projects/weirdly/vendor/symfony/flex/src/Downloader.php:364
 Symfony\Flex\Downloader->parseJson() at /home/cameron/Projects/weirdly/vendor/symfony/flex/src/Downloader.php:304
 Symfony\Flex\Downloader->Symfony\Flex\{closure}() at n/a:n/a
 call_user_func() at phar:///home/cameron/bin/composer/vendor/react/promise/src/React/Promise/FulfilledPromise.php:20
 React\Promise\FulfilledPromise->then() at phar:///home/cameron/bin/composer/vendor/react/promise/src/React/Promise/Deferred.php:74
 React\Promise\Deferred->React\Promise\{closure}() at n/a:n/a
 call_user_func() at phar:///home/cameron/bin/composer/vendor/react/promise/src/React/Promise/Deferred.php:166
[...]

Issue #630, opened 2020-06-05
Cleanup .gitignore after recipe sync


ro0NL:
I try to run monthly recipe syncs or so.

Everytime i have to clean up the same .gitignore files, would be nice if these are automatically cleaned after a recipe install/update  ¯\_(ツ)_/¯

e.g. https://github.com/symfony/recipes/blob/master/doctrine/doctrine-bundle/1.6/src/Entity/.gitignore

javiereguiluz:
I also find these files annoying and I remove them ... but when I'm working in existing applications.

For new applications is great that directories are created for you.

So I guess this is tricky.

ro0NL:
Well it's OK to generate empty dirs, which is what the .gitignore currently is for.

But then after the file can be removed, while remaining the empty dir until the user commits it.

7ochem:
I agree. Maybe Symfony Flex could check if there are files in that directory already and then don't sync the `.gitignore` file for that directory?

B.t.w. in my company we use `.gitkeep` files to keep empty directories in Git. As a `.gitignore` file suggest the intention that the file has been put there with the purpose of ignoring certain files in the directory. With `.gitkeep` it is clear that it's purpose is to _keep_ the empty dir in Git. I do know it is not documented and it is not officially supported by Git, but it is an industry standard.

ro0NL:
> .gitkeep isn’t documented, because it’s not a feature of Git.
> Git cannot add a completely empty directory.

hence we use an empty .gitignore

nicolas-grekas:
`.gitignore` is the industry standard. Check https://stackoverflow.com/questions/7229885/what-are-the-differences-between-gitignore-and-gitkeep for the background behind this claim.

@ro0NL PR welcome.

ro0NL:
WFM as-is.

PR #845, opened 2021-12-16
Merge-powered recipe update system

weaverryan:
Hi!

tl;dr; Updating recipes was a pain. Now its easy: the new `recipes:update` command performs a smart "patch" to your project of what actually changed in the recipe.

[asciicast](https://asciinema.org/a/456707)

If you run simply `composer recipes:update`, then it will ask you what to update, from a list of only the out-dated recipes (thanks to @shadowc for the idea!):

Screen Shot 2021-12-17 at 11 18 09 AM

As you can see, if there are conflicts, you fix them like normal git conflicts. In addition to the unit tests, I've tried this on a fairly out-dated project and it worked *brilliantly*. This was NOT possible until @nicolas-grekas open sourced the Flex server - so a HUGE thanks 🎆 

### How It Works

A) Whenever `symfony/recipes` (or contrib) has a new commit, a GitHub action *already* stores each recipe as JSON, which becomes the "recipe API": https://github.com/symfony/recipes/tree/flex/main. A change to that action (https://github.com/symfony/recipes/pull/1037) and the recipes-checker (https://github.com/symfony-tools/recipes-checker/pull/2) will add an "archived/" directory that contains EVERY version of every recipe. You can see an example in my fork: https://github.com/weaverryan/recipes/tree/flex/main/archived

B) When you run `recipes:update <package/name>`, we fetch the "original" recipe and "new" recipe and then pass both to each "configurator". Its job is to say what files would exist and what they would look if the "original" recipe were installed now and if the "new" recipe were installed now.

C) We use these "original files" and "new files" to generate a patch file.

D) We apply that patch file to the user's actual project. There are a few tricks here, like temporarily inserting the git "blob" for what the "original" version of the file would look like. That allows for a 3-way merge.

So, while it was a bit of work, it's a fairly straightforward process. It makes keeping your recipes up to date both easy and rewarding.

## TODOS

* [x] Merge recipes-checker PR https://github.com/symfony-tools/recipes-checker/pull/2
* [x] Merge symfony/recipes PR https://github.com/symfony/recipes/pull/1037
* [x] Update `Downloader` in this PR to remove 2x TODOs that the above PR's will enable
* [x] Make sure the `archived/` directory is populated for both `recipes` and `recipes-contrib` - see https://github.com/symfony-tools/recipes-checker/pull/3 and https://github.com/symfony/recipes/pull/1038
* [x] Document that this now exists https://github.com/symfony/symfony-docs/pull/16301

Cheers!

### DIFF
--- a/src/Command/UpdateRecipesCommand.php
+++ b/src/Command/UpdateRecipesCommand.php
@@ -1,0 +1,423 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Flex\Command;
+
+use Composer\Command\BaseCommand;
+use Composer\IO\IOInterface;
+use Composer\Util\ProcessExecutor;
+use Symfony\Component\Console\Exception\RuntimeException;
+use Symfony\Component\Console\Formatter\OutputFormatterStyle;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Flex\Configurator;
+use Symfony\Flex\Downloader;
+use Symfony\Flex\Flex;
+use Symfony\Flex\GithubApi;
+use Symfony\Flex\InformationOperation;
+use Symfony\Flex\Lock;
+use Symfony\Flex\Recipe;
+use Symfony\Flex\Update\RecipePatcher;
+use Symfony\Flex\Update\RecipeUpdate;
+
+class UpdateRecipesCommand extends BaseCommand
+{
+    /** @var Flex */
+    private $flex;
+    private $downloader;
+    private $configurator;
+    private $rootDir;
+    private $githubApi;
+    private $processExecutor;
+
+    public function __construct(/* cannot be type-hinted */ $flex, Downloader $downloader, $httpDownloader, Configurator $configurator, string $rootDir)
+    {
+        $this->flex = $flex;
+        $this->downloader = $downloader;
+        $this->configurator = $configurator;
+        $this->rootDir = $rootDir;
+        $this->githubApi = new GithubApi($httpDownloader);
+
+        parent::__construct();
+    }
+
+    protected function configure()
+    {
+        $this->setName('symfony:recipes:update')
+            ->setAliases(['recipes:update'])
+            ->setDescription('Updates an already-installed recipe to the latest version.')
+            ->addArgument('package', InputArgument::OPTIONAL, 'Recipe that should be updated.')
+        ;
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output): int
+    {
+        $win = '\\' === \DIRECTORY_SEPARATOR;
+        $runtimeExceptionClass = class_exists(RuntimeException::class) ? RuntimeException::class : \RuntimeException::class;
+        if (!@is_executable(strtok(exec($win ? 'where git' : 'command -v git'), \PHP_EOL))) {
+            throw new $runtimeExceptionClass('Cannot run "recipes:update": git not found.');
+        }
+
+        $io = $this->getIO();
+        if (!$this->isIndexClean($io)) {
+            $io->write([
+                '  Cannot run <comment>recipes:update</comment>: Your git index contains uncommitted changes.',
+                '  Please commit or stash them and try again!',
+            ]);
+
+            return 1;
+        }
+
+        $packageName = $input->getArgument('package');
+        $symfonyLock = $this->flex->getLock();
+        if (!$packageName) {
+            $packageName = $this->askForPackage($io, $symfonyLock);
+
+            if (null === $packageName) {
+                $io->writeError('All packages appear to be up-to-date!');
+
+                return 0;
+            }
+        }
+
+        if (!$symfonyLock->has($packageName)) {
+            $io->writeError([
+                'Package not found inside symfony.lock. It looks like it\'s not installed?',
+                sprintf('Try running <info>composer recipes:install %s --force -v</info> to re-install the recipe.', $packageName),
+            ]);
+
+            return 1;
+        }
+
+        $packageLockData = $symfonyLock->get($packageName);
+        if (!isset($packageLockData['recipe'])) {
+            $io->writeError([
+                'It doesn\'t look like this package had a recipe when it was originally installed.',
+                'To install the latest version of the recipe, if there is one, run:',
+                sprintf('  <info>composer recipes:install %s --force -v</info>', $packageName),
+            ]);
+
+            return 1;
+        }
+
+        $recipeRef = $packageLockData['recipe']['ref'] ?? null;
+        $recipeVersion = $packageLockData['recipe']['version'] ?? null;
+        if (!$recipeRef || !$recipeVersion) {
+            $io->writeError([
+                'The version of the installed recipe was not saved into symfony.lock.',
+                'This is possible if it was installed by an old version of Symfony Flex.',
+                'Update the recipe by re-installing the latest version with:',
+                sprintf('  <info>composer recipes:install %s --force -v</info>', $packageName),
+            ]);
+
+            return 1;
+        }
+
+        $originalRecipe = $this->getRecipe($packageName, $recipeRef, $recipeVersion);
+
+        if (null === $originalRecipe) {
+            $io->writeError([
+                'The original recipe version you have installed could not be found, it may be too old.',
+                'Update the recipe by re-installing the latest version with:',
+                sprintf('  <info>composer recipes:install %s --force -v</info>', $packageName),
+            ]);
+
+            return 1;
+        }
+
+        $newRecipe = $this->getRecipe($packageName);
+
+        if ($newRecipe->getRef() === $originalRecipe->getRef()) {
+            $io->write(sprintf('This recipe for <info>%s</info> is already at the latest version.', $packageName));
+
+            return 0;
+        }
+
+        $io->write([
+            sprintf('  Updating recipe for <info>%s</info>...', $packageName),
+            '',
+        ]);
+
+        $recipeUpdate = new RecipeUpdate($originalRecipe, $newRecipe, $symfonyLock, $this->rootDir);
+        $this->configurator->populateUpdate($recipeUpdate);
+        $originalComposerJsonHash = $this->flex->getComposerJsonHash();
+        $patcher = new RecipePatcher($this->rootDir, $io);
+
+        try {
+            $patch = $patcher->generatePatch($recipeUpdate->getOriginalFiles(), $recipeUpdate->getNewFiles());
+            $hasConflicts = !$patcher->applyPatch($patch);
+        } catch (\Throwable $throwable) {
+            $io->writeError([
+                '<bg=red;fg=white>There was an error applying the recipe update patch</>',
+                $throwable->getMessage(),
+                '',
+                'Update the recipe by re-installing the latest version with:',
+                sprintf('  <info>composer recipes:install %s --force -v</info>', $packageName),
+            ]);
+
+            return 1;
+        }
+
+        $symfonyLock->add($packageName, $newRecipe->getLock());
+        $this->flex->finish($this->rootDir, $originalComposerJsonHash);
+
+        // stage symfony.lock, as all patched files with already be staged
+        $cmdOutput = '';
+        $this->getProcessExecutor()->execute('git add symfony.lock', $cmdOutput, $this->rootDir);
+
+        $io->write([
+            '  <bg=blue;fg=white>                      </>',
+            '  <bg=blue;fg=white> Yes! Recipe updated! </>',
+            '  <bg=blue;fg=white>                      </>',
+            '',
+        ]);
+
+        if ($hasConflicts) {
+            $io->write([
+                '  The recipe was updated but with <bg=red;fg=white>one or more conflicts</>.',
+                '  Run <comment>git status</comment> to see them.',
+                '  After resolving, commit your changes like normal.',
+            ]);
+        } else {
+            if (!$patch->getPatch()) {
+                // no changes were required
+                $io->write([
+                    '  No files were changed as a result of the update.',
+                ]);
+            } else {
+                $io->write([
+                    '  Run <comment>git status</comment> or <comment>git diff --cached</comment> to see the changes.',
+                    '  When you\'re ready, commit these changes like normal.',
+                ]);
+            }
+        }
+
+        if (0 !== \count($recipeUpdate->getCopyFromPackagePaths())) {
+            $io->write([
+                '',
+                '  <bg=red;fg=white>NOTE:</>',
+                '  This recipe copies the following paths from the bundle into your app:',
+            ]);
+            foreach ($recipeUpdate->getCopyFromPackagePaths() as $source => $target) {
+                $io->write(sprintf('  * %s => %s', $source, $target));
+            }
+            $io->write([
+                '',
+                '  The recipe updater has no way of knowing if these files have changed since you originally installed the recipe.',
+                '  And so, no updates were made to these paths.',
+            ]);
+        }
+
+        if (0 !== \count($patch->getRemovedPatches())) {
+            if (1 === \count($patch->getRemovedPatches())) {
+                $notes = [
+                    sprintf('  The file <comment>%s</comment> was not updated because it doesn\'t exist in your app.', array_keys($patch->getRemovedPatches())[0]),
+                ];
+            } else {
+                $notes = ['  The following files were not updated because they don\'t exist in your app:'];
+                foreach ($patch->getRemovedPatches() as $filename => $contents) {
+                    $notes[] = sprintf('    * <comment>%s</comment>', $filename);
+                }
+            }
+            $io->write([
+                '',
+                '  <bg=red;fg=white>NOTE:</>',
+            ]);
+            $io->write($notes);
+            $io->write('');
+            if ($io->askConfirmation('  Would you like to save the "diff" to a file so you can review it? (Y/n) ')) {
+                $patchFilename = str_replace('/', '.', $packageName).'.updates-for-deleted-files.patch';
+                file_put_contents($this->rootDir.'/'.$patchFilename, implode("\n", $patch->getRemovedPatches()));
+                $io->write([
+                    '',
+                    sprintf('  Saved diff to <info>%s</info>', $patchFilename),
+                ]);
+            }
+        }
+
+        if ($patch->getPatch()) {
+            $io->write('');
+            $io->write('  Calculating CHANGELOG...', false);
+            $changelog = $this->generateChangelog($originalRecipe);
+            $io->write("\r", false); // clear current line
+            if ($changelog) {
+                $io->write($changelog);
+            } else {
+                $io->write('No CHANGELOG could be calculated.');
+            }
+        }
+
+        return 0;
+    }
+
+    private function getRecipe(string $packageName, string $recipeRef = null, string $recipeVersion = null): ?Recipe
+    {
+        $installedRepo = $this->getComposer()->getRepositoryManager()->getLocalRepository();
+        $package = $installedRepo->findPackage($packageName, '*');
+        if (null === $package) {
+            throw new RuntimeException(sprintf('Could not find package "%s". Try running "composer install".', $packageName));
+        }
+        $operation = new InformationOperation($package);
+        if (null !== $recipeRef) {
+            $operation->setSpecificRecipeVersion($recipeRef, $recipeVersion);
+        }
+        $recipes = $this->downloader->getRecipes([$operation]);
+
+        if (0 === \count($recipes['manifests'] ?? [])) {
+            return null;
+        }
+
+        return new Recipe(
+            $package,
+            $packageName,
+            $operation->getOperationType(),
+            $recipes['manifests'][$packageName],
+            $recipes['locks'][$packageName] ?? []
+        );
+    }
+
+    private function generateChangelog(Recipe $originalRecipe): ?array
+    {
+        $recipeData = $originalRecipe->getLock()['recipe'] ?? null;
+        if (null === $recipeData) {
+            return null;
+        }
+
+        if (!isset($recipeData['ref']) || !isset($recipeData['repo']) || !isset($recipeData['branch']) || !isset($recipeData['version'])) {
+            return null;
+        }
+
+        $currentRecipeVersionData = $this->githubApi->findRecipeCommitDataFromTreeRef(
+            $originalRecipe->getName(),
+            $recipeData['repo'],
+            $recipeData['branch'],
+            $recipeData['version'],
+            $recipeData['ref']
+        );
+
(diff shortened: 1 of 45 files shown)


nicolas-grekas:
It's Christmas before Christmas! :gift_heart:

nicolas-grekas:
> an "archived/" directory that contains EVERY version of every recipe

Did you consider using git instead of maintaining this directory? cloning the recipe repo locally and using git to checkout the tree-ref? Not saying this is would be better. This is just what I had in mind :)

maxhelias:
I just read some part and it an amazing work! 👏

weaverryan:
> Did you consider using git instead of maintaining this directory? cloning the recipe repo locally and using git to checkout the tree-ref? Not saying this is would be better. This is just what I had in mind :)

That definitely seems like a valid approach. I guess I was trying to (A) minimize the amount of work needed on the user's local computer (e.g. git cloning the recipes repo) and (B) loading the "old" recipe data in the exact same way as normal (recipes are normally loaded from JSON files... so why not do the exact same thing to load the old recipes)?

So, I think my approach for that part is the correct one. However, that represents just a small part of the PR: if we needed to change that, it wouldn't be a huge thing.

nicolas-grekas:
With longer URLs to cache, we might hit #830 soon.
Either we should add some logic to `generateCacheKey()` to make it never generate file names longer than 143 chars or we should skip caching calls to "archived" jsons.
Maybe both :)

weaverryan:
> With longer URLs to cache, we might hit #830 soon.

I tested this out - here's an example cache key for an archive:

> weaverryan-recipes-flex-main-archived-symfony.framework-bundle-ffcb76bab779154c0a7d8cddb9f448c2c45a4e50.json

That's length 108. So we have 35 more chars to work with, which seems unlikely to hit. UNLESS someone uses a non-Github endpoint, in which case the current `generateCacheKey()` won't shorten it nearly as much.

I've just pushed a little tweak where, if the key is longer than 140 (this lives a little wiggle room for eCryptfs), we `sha($url)`. That leaves some inconsistency... but it will leave pretty much all filenames as readable in the cache dir, which is nice.

src/Update/RecipePatcher.php
```diff
@@ -0,0 +1,214 @@
...
+        }
+
+        return $addedBlobs;
+    }
+
+    private function generateBlobs(array $originalFiles, string $originalFilesRoot): array
+    {
+        $addedBlobs = [];
+        foreach ($originalFiles as $filename => $contents) {
+            // if the didn't originally exist, no blob needed
```
nicolas-grekas: the?

src/Update/RecipePatcher.php
```diff
@@ -0,0 +1,214 @@
...
+        $this->rootDir = $rootDir;
+        $this->filesystem = new Filesystem();
+        $this->io = $io;
+        $this->processExecutor = new ProcessExecutor($io);
+    }
+
+    /**
+     * Applies the patch. If it fails unexpectedly, an exception will be thrown.
+     *
+     * @return bool returns true of fully successful, false if conflicts were encountered
```
nicolas-grekas: s/of/if

src/Downloader.php
```diff
@@ -173,6 +182,26 @@ public function getRecipes(array $operations): array
                     if (version_compare($version, $v, '>=')) {
                         $data['locks'][$package->getName()]['version'] = $version;
                         $data['locks'][$package->getName()]['recipe']['version'] = $v;
+
+                        if (null !== $recipeRef) {
+                            // TODO: add a new archived_recipe_template to _links
```
nicolas-grekas: done :)

src/Downloader.php
```diff
@@ -29,7 +29,8 @@
 class Downloader
 {
     private const DEFAULT_ENDPOINTS = [
-        'https://raw.githubusercontent.com/symfony/recipes/flex/main/index.json',
+        // TODO: REVERT THIS
```
nicolas-grekas: can be done

nicolas-grekas:
> if the key is longer than 140, we sha($url)

Nice thanks.

While you're on this topic, I know we have an issue right now related to files that are removed from recipes. Eg when `config/package/test/framwork.yaml` moved to `when@dev`. Will this be able to spot the removal? If not, is it too much to ask for this very PR? :angel:

weaverryan:
> While you're on this topic, I know we have an issue right now related to files that are removed from recipes. Eg when config/package/test/framwork.yaml moved to when@dev. Will this be able to spot the removal? If not, is it too much to ask for this very PR? 👼

Yup! It catches those! You can even see it at the end of the video in the PR description: a `dev` and `test` files are deleted because the `when@dev` has been patched into the main file. 🎄 

I don't see a specific issue or PR about that, but I know it has been an issue (there is #706, but it's about `unconfigure()`.

shadowc:
I've tested on Windows updating a few recipes and it worked without issues for me!

nicolas-grekas:
Did you consider submitting against `1.x`?

stof:
> That's length 108. So we have 35 more chars to work with, which seems unlikely to hit. UNLESS someone uses a non-Github endpoint,

or just a longer package name than `symfony/framework-bundle`, which is likely to happen at some point

weaverryan:
> Did you consider submitting against 1.x?

I was hoping you wouldn't ask ;). But strictly speaking, I CAN change to that if you want. It would mostly involve removing types and a few other things. Annoying, but doable. WDYT?

Comments so far have been addressed. Only 1 TODO left in the code, pending the full `archived/` history being added to `symfony/recipes`.

nicolas-grekas:
Haha, then please target `1.x` :)

src/Update/RecipePatcher.php
```diff
@@ -0,0 +1,220 @@
...
+            if (!isset($newFiles[$file])) {
+                $newFiles[$file] = null;
+            }
+        }
+
+        // If a file is being modified, but does not exist in the current project,
+        // it cannot be patched. Ignore these files and leave them deleted.
+        $modifiedFiles = array_intersect_key(array_keys($originalFiles), array_keys($newFiles));
+        foreach ($modifiedFiles as $modifiedFile) {
+            if (!file_exists($this->rootDir.'/'.$modifiedFile)) {
```
jderusse: Would be great to not lose the patch when the file does not exist!

My used case: I switched from .yaml to .PHP configuration file. But I would appreciate having a diff of changes. In that way, I can implement the diff in PHP.

Logic could be something like:
if file exists in both `$originalFiles` and `$newFiles` BUT does not exists on file system (it means the USER deleted it) then dump a `.patch` somewhere (in file system, or stdout)

src/Update/RecipePatcher.php
```diff
@@ -0,0 +1,220 @@
...
+                    unlink($path);
+                }
+
+                continue;
+            }
+
+            if (!file_exists(\dirname($path))) {
+                $this->filesystem->mkdir(\dirname($path));
+            }
+            file_put_contents($path, $contents);
```
jderusse, line 160: if content is null (case when file exists in `original` but not in `new`) should it delete the file instead of putting an empty content in it?

jderusse (commented):
thank you so much for this :heart:

src/Update/RecipePatcher.php
weaverryan: Hmm, that’s an interesting idea! I cannot apply the patch - it just doesn’t work - but I could dump a patch file.

weaverryan:
The `archived/` directory is now populated on `recipes` and `recipes-contrib`! I've updated this PR to use the true endpoints 🎆 .

I'll re-work this for the 1.x branch, but just because you asked nicely ;)

src/Update/RecipePatcher.php
weaverryan: It does this a few lines up - there's a check for null `$contents` and an `unlink` (if the file exists - because if this is the starting file, then there is nothing yet to delete). So unless you're seeing something that looks funny to you, it already deletes on null :).

W2AlharbiMe:
Niiice, well done 👏👏

src/Update/RecipePatcher.php
weaverryan: Screen Shot 2021-12-18 at 4 01 46 PM

How about this :)

weaverryan:
Ok, this is now against the 1.x branch, tested locally, tests are green!

src/Update/RecipePatcher.php
jderusse: that's perfect ❤️

7ochem:
I've setup an outdated project to use your fork+branch of symfony/flex and tried to run `composer symfony:recipes:update`. I got a list of outdated recipes that I could update and my first thought was "I want to updated them all", but maybe this could be a next step after this PR has been merged. So I just picked the first one and got this error:

```
There was an error applying the recipe update patch
Command "git commit -m "original files"" failed: "error: gpg failed to sign the data
fatal: failed to write commit object
". Output: "".
```

A quick Google gave me some suggestions. I think you should probably add `git config commit.gpgsign false` (or first check the setting with the output of `git config commit.gpgsign`)

In src/Update/RecipePatcher.php:
```diff
@@ -107,6 +107,7 @@ class RecipePatcher
 
         try {
             $this->execute('git init', $tmpPath);
+            $this->execute('git config commit.gpgsign false', $tmpPath);
             $this->execute('git config user.name "Flex Updater"', $tmpPath);
             $this->execute('git config user.email ""', $tmpPath);
```

7ochem:
The next thing I'm running into is that I've got the habit of removing those `.gitignore` files that recipes place in empty directories to keep the directories under VCS once I've added actual files in them (mentioned before in #630). I think something is missing in the process to take deleted files into account. I'm thinking of any other situation where you would delete files that have been placed by recipes (because you don't want them), but cannot think of any other than those `.gitignore` files. I'm also fine if the answer will be: "Don't delete them", because the advantage of having an "recipe update" system is way better 💪🏻 

This happened when running recipe update for `doctrine/doctrine-bundle`:
```
Executing command (/tmp/_flex_recipe_update199562016461c04b5485c2a3.63558769): git init
Executing command (/tmp/_flex_recipe_update199562016461c04b5485c2a3.63558769): git config commit.gpgsign false
Executing command (/tmp/_flex_recipe_update199562016461c04b5485c2a3.63558769): git config user.name "Flex Updater"
Executing command (/tmp/_flex_recipe_update199562016461c04b5485c2a3.63558769): git config user.email ""
Executing command (/tmp/_flex_recipe_update199562016461c04b5485c2a3.63558769): git add -A
Executing command (/tmp/_flex_recipe_update199562016461c04b5485c2a3.63558769): git commit -m "original files"
Executing command (/tmp/_flex_recipe_update199562016461c04b5485c2a3.63558769): git hash-object 'config/bundles.php'
Executing command (/tmp/_flex_recipe_update199562016461c04b5485c2a3.63558769): git hash-object 'config/packages/doctrine.yaml'
Executing command (/tmp/_flex_recipe_update199562016461c04b5485c2a3.63558769): git hash-object 'config/packages/prod/doctrine.yaml'
Executing command (/tmp/_flex_recipe_update199562016461c04b5485c2a3.63558769): git hash-object 'src/Entity/.gitignore'
Executing command (/tmp/_flex_recipe_update199562016461c04b5485c2a3.63558769): git hash-object 'src/Repository/.gitignore'
Executing command (/tmp/_flex_recipe_update199562016461c04b5485c2a3.63558769): git hash-object '.env'
Executing command (/tmp/_flex_recipe_update199562016461c04b5485c2a3.63558769): git hash-object 'phpunit.xml.dist'
Executing command (/tmp/_flex_recipe_update199562016461c04b5485c2a3.63558769): git add -A
Executing command (/tmp/_flex_recipe_update199562016461c04b5485c2a3.63558769): git diff --cached
There was an error applying the recipe update patch
Could not find file "src/Entity/.gitignore" in the patch.
```

After `touch src/Entity/.gitignore` and `touch src/Repository/.gitignore` everything ran just fine 👍🏻 

Great job @weaverryan !

weaverryan:
@7ochem THANK YOU for thoroughly testing this - that awesome!

> error: gpg failed to sign the data

I've added your suggested fix for this - thank you!

> Could not find file "src/Entity/.gitignore" in the patch.

This was a bug I introduced in the latest feature I added - very good catch! It's now fixed - it will work without needing to `touch` those files manually (it's totally ok to delete files from recipes like this - the `recipe:update` system handles it just fine, now that this bug is gone).

So, tests are still green and all comments have been addressed!

7ochem:
One more thing... When trying this out, I had changed my composer.json (and composer.lock) file because I had to target your fork and this branch, so I had uncommitted changes in those files when running `composer symfony:recipes:update`. Then I got this error:

```
There was an error applying the recipe update patch
Command "git update-index --refresh" failed: "". Output: "composer.json: needs update
composer.lock: needs update
".
```

Maybe you could show a warning at the start that there are uncommitted changes in files and advise to commit any changes before running `recipe:update` (because it is using Git to process and apply the necessary changes)

weaverryan:
> Maybe you could show a warning at the start that there are uncommitted changes in files and advise to commit any changes before running recipe:update (because it is using Git to process and apply the necessary changes)

Done!

7ochem (approved):
Nice! 👍🏻

src/Command/UpdateRecipesCommand.php
```diff
@@ -0,0 +1,407 @@
...
+        ;
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output): int
+    {
+        $win = '\\' === \DIRECTORY_SEPARATOR;
+        $runtimeExceptionClass = class_exists(RuntimeException::class) ? RuntimeException::class : \RuntimeException::class;
+        if (!@is_executable(strtok(exec($win ? 'where git' : 'command -v git'), \PHP_EOL))) {
+            throw new $runtimeExceptionClass('Cannot run "recipes:update": git not found.');
+        }
```
wouterj, line 68: I think it's better to use `ExecutableFinder` here and then use the found executable for all git commands? This allows best compatibility (e.g. when someone on Windows has `git.cmd` instead of `git`) and safety (e.g. when using this on a host with `open_basedir` set).

src/Command/UpdateRecipesCommand.php
```diff
@@ -0,0 +1,407 @@
...
+        if ('' !== trim($output)) {
+            return false;
+        }
+
+        $processExecutor->execute('git diff --name-only', $output, $this->rootDir);
+        if ('' !== trim($output)) {
+            return false;
+        }
+
+        return true;
```
wouterj:
```suggestion
        $processExecutor = new ProcessExecutor($io);

        return 0 === $processExecutor->execute('git diff-index --quiet HEAD');
```

(taken from https://unix.stackexchange.com/a/394674/381461)

wouterj: Optionally, we can also use `git status --porcelain` instead.

These 2 commands allow to check cached and non-cached index trees at the same time, and have more stability guarantees across git versions.

src/Command/UpdateRecipesCommand.php
```diff
@@ -0,0 +1,407 @@
...
+            ]);
+        } else {
+            if (!$patch->getPatch()) {
+                // no changes were required
+                $io->write([
+                    '  No files were changed as a result of the update.',
+                ]);
+            } else {
+                $io->write([
+                    '  Run <comment>git status</comment> the changes then commit like normal.',
```
wouterj:
```suggestion
                    '  Run <comment>git status</comment> to see the changes then commit like normal.',
```

wouterj: Maybe also tell about `git diff --cached`. It was a bit unexpected to me when I ran the first time that this command added all changes to the index already.

src/Command/UpdateRecipesCommand.php
```diff
@@ -0,0 +1,407 @@
...
+        }
+
+        $newRecipe = $this->getRecipe($packageName);
+
+        if ($newRecipe->getRef() === $originalRecipe->getRef()) {
+            $io->write(sprintf('This recipe for <info>%s</info> is already at the latest version.', $packageName));
+
+            return 0;
+        }
+
```
wouterj, line 144:
```suggestion

    $io->write(sprintf('  Updating recipe for <info>%s</info>...', $packageName));

```

The first message that appears when running a happy-flow is "Yes, the recipe is updated". While this gives you a "yay" feeling, it also feels like the common "I'm going to do stuff for you" initial message was missing.

<details>
<summary>Besides, happy-flow output also didn't mention the package name I was updating once in the output</summary>

```
   Yes! Recipe updated!


  Run git status the changes then commit like normal.

  * #916 - Routing: don't require to install doctrine/annotations to use attributes (@dunglas)
      https://github.com/symfony/recipes/pull/916
```

</detail>

wouterj (commented):
A very quick test run of this PR (and added some suggestions about the command class). I love how easy it is to get the references to the changes and to fix conflicts with local modifications! Great job!
It comes with a small price of requiring to have a clean git index, etc. But that small price is worth the benefits this gives us.

One thing that might be improved is the final git state after running this command. As the changes are applied, they are now part of the git index. But the `symfony.lock` changes aren't. I would suggest to either make sure everything modified by the command is part of the index, or nothing.

src/Command/UpdateRecipesCommand.php
weaverryan: That sounds good to me! But I copied this from `InstallRecipesCommand`, which was added by @nicolas-grekas https://github.com/symfony/flex/commit/d3adcac0fa1749403b4010f14af73438908c5d4b

Nicolas - did you purposely avoid `ExecutableFinder`?

src/Command/UpdateRecipesCommand.php
weaverryan: I had some back and forth on this, and I think I missed the obvious solution - so thanks :). I've changed to:

```
git status --porcelain --untracked-files=no
```

Which seems to work well.

weaverryan:
Thank you for the thoughtful review Wouter!

> But the symfony.lock changes aren't. I would suggest to either make sure everything modified by the command is part of the index, or nothing.

To make sure I understand, are you referring to the fact that the `symfony.lock` file contains a `files` key under some recipes which tracks the added files... but that it doesn't (inside of this file) store information about the other files that were updated/added? If so, it's not strictly related to this PR, but that "struck" me as odd also. I think this was originally done to help with the `composer recipes <package/name>` command, so it could show which files it added. But it does completely ignore all the other files that might be added or modified via other configurator (beyond "copy from recipe"). Probably, but in a different PR, we should clean this up. I'd recommend removing things from symfony.lock because it's not very easy - at any time - to query the Flex API and get the full details of what the recipe looked like when you originally installed it. We could probably use that info for a visualization in the `composer recipes` command.

wouterj:
> To make sure I understand, are you referring to the fact that the symfony.lock file contains a files key under some recipes which tracks the added files... but that it doesn't (inside of this file) store information about the other files that were updated/added? 

I'm sorry for the confusion, this is way over my head for flex internals, but I'm sure your answer will feed ideas to other flex contributors :)

What I was talking about was the git index. After running `recipe:update` for a recipe, `git status` gives me this:

```
On branch main
Changes to be committed:
  (use "git restore --staged <file>..." to unstage)
        modified:   config/routes.yaml

Changes not staged for commit:
  (use "git add <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
        modified:   symfony.lock
```

I would love if either all changed files and `modified: symfony.lock` would be in the "Changes to be committed" section, or all changes would be under "Changes not staged for commit". Currently, I find it a bit weird to see some of the changes done by this command to be auto-staged, and others don't (ofc, I understand the internal reasons behind this, but looking this is looking at it from a user perspective).

weaverryan:
> I would love if either all changed files and modified: symfony.lock would be in the "Changes to be committed" section

Oh, of course! Nice idea - I've just made that change :)

src/Command/UpdateRecipesCommand.php
nicolas-grekas: > e.g. when someone on Windows has git.cmd instead of git

Looking at Composer's code (which always uses `git`), I don't think we should care about this.

Here, we are checking if git is installed. `ExecutableFinder` might do the job for sure, but since we're relying on the OS discovery, the current logic is consistent to me.

No strong opinion.

nicolas-grekas:
Having to select a package in the list doesn't feel like a real choice. Aka how can I know which package I want to update?
Based on this, I'm wondering if there is a way to upgrade all recipes at once? Eg iterate until a conflict happens? Since conflicts are going to be quite common, continue applying updates until two conflicts happen on the same file?

(The simple way to take my concern into account is to set 0 as the default choice in the list of choices and tell ppl to run the command as many time as needed, but I'm wondering if we can do better)

fabpot:
The good thing about doing it one by one is the ease of review: the diff is smaller and easier to understand I suppose.

weaverryan:
You’re not the first person to ask about updating all of them at once. I agree with Fab that it’s nice to review them one-by-one, including the CHANGELOG that’s generated for each.

If we were to update all at once, from a practical perspective, I think we would need to commit automatically because the patch system needs the working tree and index to be clean… though i can’t remember the exact reason why (I did add a check for this on purpose).

So I’d like to default to the first choice and take the easy way out (just default the 0 choice) - but I’m open to either, which would include either (A) committing for the user or (B) more likely, investigating if we can apply changes to a dirty tree.

wouterj:
> but I’m open to either, which would include either (A) committing for the user or (B) more likely, investigating if we can apply changes to a dirty tree.

From a usage point of view, that last point is very interesting (I also preferred applying on a dirty tree when testing the current feature). But to keep the scope small, I would do it outside this big PR :)

weaverryan:
New look, with the default :) 

Screen Shot 2022-01-17 at 10 24 58 AM

nicolas-grekas (approved):
LGTM!

Some note about the changelog that is displayed at the end of an update:
- because it relies on calls to api.github.com, it won't be compatible with private recipes (but we don't have to care until someone sends a PR to improve this), but the update itself will still work if I understood correctly since updates rely only on the JSONs.
- I'd prefer if we could use one-liners:

```
  * #1012 - Add missing return type (@dreadnip)
      https://github.com/symfony/recipes/pull/1012
```

could be
```
  * #1012 - Add missing return type - https://github.com/symfony/recipes/pull/1012
```
we could even make the PR number clickable and hide the full URL:
```
  * Add missing return type (#1012)
```

- it'd be nice to display the same changelog when running the `recipes foo/bar` command (this command is a bit outdated now :) )

wouterj:
> we could even make the PR number clickable and hide the full URL:

For what's it worth: I don't think all shells have support for this

fabpot:
> > we could even make the PR number clickable and hide the full URL:
> 
> For what's it worth: I don't think all shells have support for this

I think the vast majority does. Let's make things work best for most people. I agree that the current changelog is quite "heavy".

weaverryan:
Changelog shortened :)

Screen Shot 2022-01-18 at 8 47 08 AM

> because it relies on calls to api.github.com, it won't be compatible with private recipes (but we don't have to care until someone sends a PR to improve this), but the update itself will still work if I understood correctly since updates rely only on the JSONs.

Correct! As long as the private recipe API implements the same structure as the main one (the same JSON files, etc), then the update will work. And (as you know) https://github.com/symfony-tools/recipes-checker has the tools that private recipe API's can use (the `generate:flex-endpoint` command creates the archived files and `generate:archived-recipes` can be used once time to generate a full history for all recipes).

fabpot (approved):

fabpot:
Thank you @weaverryan.

Issue #73, opened 2017-05-07
Flex tries to inspect vendor/ after Composer has already removed the package


dkarlovi:
Basically this:

    $ SYMFONY_ENDPOINT=https://symfony.sh/r/contrib/15 composer remove dkarlovi/symfony-flex-docker
    Loading composer repositories with package information
    Updating dependencies (including require-dev)
    Package operations: 0 installs, 0 updates, 1 removal
      - Removing dkarlovi/symfony-flex-docker (0.1.3)
        Auto-unconfiguring "dkarlovi/symfony-flex-docker"
        Removing configuration and files
    
    Removal failed, reverting ./composer.json to its original content.
    
                                                                                                                                             
      [UnexpectedValueException]                                                                                                             
      RecursiveDirectoryIterator::__construct(/home/dado/Development/symfony-skeleton/vendor/dkarlovi/symfony-flex-docker/.infra/): failed   
      to open dir: No such file or directory

fabpot:
Thinking about this issue a bit more, I'm wondering if copying files from a package is really a feature we want to keep.

nicolas-grekas:
Closing as staled. Please provide a PR ideally if you can.

tperunsky:
@nicolas-grekas 
I would like to reopen this issue to discuss options. 

The problem is described in more detail here: https://github.com/symfony/flex/issues/604 
Although it seems to me that the root cause is that `\Symfony\Flex\Configurator\CopyFromPackageConfigurator::removeFilesFromDir()` iterates over files in the package to figure out what to delete from the project. This is probably to make sure Flex doesn't delete files/folders which weren't copied from the package by the recipe in the first place. However, the package no longer exists in `/vendor` when this code runs.

In short, 
1. on composer install the `copy-from-package` stuff is copied (eg FE assets) 
2. on composer remove there's an error and as a result the composer package is removed, but the recipe is still in symfony.lock (bundle.php entry and `copy-from-recipe` files get removed correctly)

Unless I'm missing something, the `copy-from-package` functionality is broken at the moment. 

@fabpot mentioned in a comment above that this functionality is in question, but 4 years later it's still there, although not fully working, as evidenced by this issue. 
IMO it should either be removed or fixed - at the very least removing a package shouldn't result in an error. Maybe instead of the error, there should be a message explaining that files copied from the package have to be removed manually. 

I can provide a PR if you think it makes sense to pursue this.

nicolas-grekas:
[...]

PR #847, opened 2022-01-04
Skipping missing source directories during copy-from-package unconfiguration

tperunsky:
https://github.com/symfony/flex/issues/73

This PR makes the output look like this: 
image

### DIFF
--- a/src/Configurator/CopyFromPackageConfigurator.php
+++ b/src/Configurator/CopyFromPackageConfigurator.php
@@ -145,6 +145,9 @@ class CopyFromPackageConfigurator extends AbstractConfigurator
 
     private function removeFilesFromDir(string $source, string $target)
     {
+        if (!is_dir($source)) {
+            return;
+        }
         $iterator = $this->createSourceIterator($source, \RecursiveIteratorIterator::CHILD_FIRST);
         foreach ($iterator as $item) {
             $targetPath = $this->path->concatenate([$target, $iterator->getSubPathName()]);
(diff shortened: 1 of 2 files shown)


src/Configurator/CopyFromPackageConfigurator.php
```diff
@@ -110,6 +110,10 @@ public function copyFile(string $source, string $target, array $options)
 
     private function removeFilesFromDir(string $source, string $target)
     {
+        if (!is_dir($source)) {
+            $this->write(sprintf('  Skipped directory <fg=green>"%s"</>: unable to determine which files to remove', $this->path->relativize($target)));
```
nicolas-grekas: I'd suggest removing the message as I don't see what it would be useful for.

tperunsky: I removed it. I still think it would be nice to have it at least when running flex with `-vvv`, but that would require even more code changes, so perhaps it's best to keep it simple at this point. From my point of view the use case is developers setting up a private flex recipe. The documentation about doesn't go into too much detail about how flex works and I think that having extra information on why a directory was not removed during unconfiguration would save users from unnecessary confusion (and frustration). Alternatively, the documentation could explain in more detail what unconfiguration does and doesn't do. The naive expectation is that it completely cleans up, but that's not the reality (for good reasons). Either way I don't feel strongly about it.

nicolas-grekas:
Thank you @tperunsky.

PR #846, opened 2022-01-04
ReturnTypeWillChange to suppress PHP8.1 Deprecation notice.

PhilETaylor:
Pretty quick and standard change for PHP 8.1

Surpress:
```
Deprecation Notice: Return type of Symfony\Flex\Response::jsonSerialize() should either be compatible with JsonSerializable::jsonSerialize(): mixed, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice in /Users/phil/Sites/manage.mysites.guru/vendor/symfony/flex/src/Response.php:68
```
Screenshot 2022-01-05 at 10 50 58

Alternative solution to #848

PR #848, opened 2022-01-04
jsonSerialize return type to suppress PHP8.1 Deprecation notice.

PhilETaylor:
jsonSerialize return type to suppress PHP8.1 Deprecation notice. 

Alternative solution to #846


fabpot:
Closing in favor of #846

PR #854, opened 2022-01-19
Adding a SYMFONY_DOCKER env var to permanently activate docker support

weaverryan:
The partner PR to: https://github.com/symfony-cli/symfony-cli/pull/53

Will allow for a `symfony new --docker` flag, which will start a project with the Docker configuration pre-activated.

### DIFF
--- a/src/Configurator/DockerComposeConfigurator.php
+++ b/src/Configurator/DockerComposeConfigurator.php
@@ -111,34 +111,14 @@ class DockerComposeConfigurator extends AbstractConfigurator
             return false;
         }
 
-        $warning = $io->isInteractive() ? 'WARNING' : 'IGNORING';
-        $io->writeError(sprintf('  - <warning> %s </> %s', $warning, $recipe->getFormattedOrigin()));
-        $question = '    The recipe for this package contains some Docker configuration.
+        if (!isset($_SERVER['SYMFONY_DOCKER'])) {
+            $answer = self::askDockerSupport($io, $recipe);
+        } elseif (filter_var($_SERVER['SYMFONY_DOCKER'], \FILTER_VALIDATE_BOOLEAN)) {
+            $answer = 'p';
+        } else {
+            $answer = 'x';
+        }
 
-    This may create/update <comment>docker-compose.yml</comment> or update <comment>Dockerfile</comment> (if it exists).
-
-    Do you want to include Docker configuration from recipes?
-    [<comment>y</>] Yes
-    [<comment>n</>] No
-    [<comment>p</>] Yes permanently, never ask again for this project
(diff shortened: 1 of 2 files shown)


src/Configurator/DockerComposeConfigurator.php
```diff
@@ -111,34 +111,12 @@ public static function shouldConfigureDockerRecipe(Composer $composer, IOInterfa
...
-            $question,
-            function ($value) {
-                if (null === $value) {
-                    return 'y';
-                }
-                $value = strtolower($value[0]);
-                if (!\in_array($value, ['y', 'n', 'p', 'x'], true)) {
-                    throw new \InvalidArgumentException('Invalid choice.');
-                }
+        if ($_SERVER['DOCKER_SUPPORT'] ?? false) {
```
nicolas-grekas: let's use `filter_var(..., \FILTER_VALIDATE_BOOLEAN)` and allow enabling or disabling via the env var?

nicolas-grekas: what about SYMFONY_DOCKER for the name?

fabpot: Re: enabling/disabling, we want Docker disables by default, and enable it when the env var is "set". We don't really care about the value. If it's false-ish, it's like not having it.

weaverryan: Name updated.

> Re: enabling/disabling, we want Docker disables by default, and enable it when the env var is "set". We don't really care about the value. If it's false-ish, it's like not having it.

And so `SYMFONY_DOCKER=0` would work if you, for some reason, wanted to use this env var but explicitly set it to false.

nicolas-grekas: it's strange to be able to SYMFONY_DOCKER=0 to enable it

It's easy cheap to implement better now isn't it?

fabpot: SYMFONY_DOCKER=0 must disable it, that's what I mentioned before.

nicolas-grekas: oh, right about SYMFONY_DOCKER=0
my point stands for SYMFONY_DOCKER=false :)
using `filter_var(..., \FILTER_VALIDATE_BOOLEAN)` is the way to check boolean envvars (or php.ini settings) in Symfony to me

dunglas: So `SYMFONY_DOCKER=1` to permanently enable Docker support and `SYMFONY_DOCKER=0` to permanently disable it (without being asked)?

+1 to use `filter_var()`.

fabpot:
Thank you @weaverryan.

PR #874, opened 2022-02-21
Add some native return types

PhilETaylor:
### Symfony version(s) affected

symfony/flex: v1.18.5

### Description

Composer version 2.3-dev (2.3-dev+c9baeda95b2da59a8c2d0bb7350b18d009aa891b) 2022-02-21 12:57:32

Symfony Project with `symfony/flex: v1.18.5` 

```
PHP Fatal error:  Declaration of Symfony\Flex\ParallelDownloader::copy($originUrl, $fileUrl, $fileName, $progress = true, $options = []) must be compatible with Composer\Util\RemoteFilesystem::copy($originUrl, $fileUrl, $fileName, $progress = true, $options = []): bool in /Users/phil/Sites/example.com/vendor/symfony/flex/src/ParallelDownloader.php on line 132

Fatal error: Declaration of Symfony\Flex\ParallelDownloader::copy($originUrl, $fileUrl, $fileName, $progress = true, $options = []) must be compatible with Composer\Util\RemoteFilesystem::copy($originUrl, $fileUrl, $fileName, $progress = true, $options = []): bool in /Users/phil/Sites/example.com/vendor/symfony/flex/src/ParallelDownloader.php on line 132
```

### How to reproduce

```

~/Sites/symfony-test composer -V
Composer version 2.3-dev (2.3-dev+c9baeda95b2da59a8c2d0bb7350b18d009aa891b) 2022-02-21 12:57:32
~/Sites/symfony-test composer selfup --snapshot
You are already using the latest available Composer version c9baeda95b2da59a8c2d0bb7350b18d009aa891b (snapshot channel).
~/Sites/symfony-test composer require symfony/flex 1.18.5
./composer.json has been created
Running composer update symfony/flex
Loading composer repositories with package information
Updating dependencies
Lock file operations: 1 install, 0 updates, 0 removals
  - Locking symfony/flex (v1.18.5)
Writing lock file
Installing dependencies from lock file (including require-dev)
Package operations: 1 install, 0 updates, 0 removals
  - Installing symfony/flex (v1.18.5): Extracting archive
symfony/flex contains a Composer plugin which is currently not in your allow-plugins config. See https://getcomposer.org/allow-plugins
Do you trust "symfony/flex" to execute code and wish to enable it now? (writes "allow-plugins" to composer.json) [y,n,d,?] y
[...]

PR #876, opened 2022-02-22
Fixes Fatal Error: Declaration of Symfony\Flex\ParallelDownloader

silentgecko:
Fatal Error: Declaration of Symfony\Flex\ParallelDownloader::copy must be compatible with Composer\Util\RemoteFilesystem::copy

Add return types to be compatible with composer changes

Newest version of composer added return types to these functions (@see https://github.com/composer/composer/commit/a16ed3d0ed6a137d11231ed77d3545998584d2e8). Now you get Fatal Errors when using new composer:
Declaration of Symfony\Flex\ParallelDownloader::copy($originUrl, $fileUrl, $fileName, $progress = true, $options = Array) must be compatible with Composer\Util\RemoteFilesystem::copy($originUrl, $fileUrl, $fileName, $progress = true, $options = Array): bool

### DIFF
--- a/src/ParallelDownloader.php
+++ b/src/ParallelDownloader.php
@@ -103,7 +103,7 @@ class ParallelDownloader extends RemoteFilesystem
         }
     }
 
-    public function getOptions()
+    public function getOptions(): array
     {
         $options = array_replace_recursive(parent::getOptions(), $this->nextOptions);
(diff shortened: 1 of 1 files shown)


nicolas-grekas:
Duplicate of #874 ;)

PR #878, opened 2022-02-22
Fix parsing docker-composer files with no "version"

### LINKED ISSUE #877: composer update is giving an error
https://github.com/symfony/flex/blob/10e438f53a972439675dc720706f0cd5c0ed94f1/src/Configurator/DockerComposeConfigurator.php#L262
access to $lines[$i - 1] is giving me an error `Undefined array key -1`. $i is 0 because it is taken from https://github.com/symfony/flex/blob/10e438f53a972439675dc720706f0cd5c0ed94f1/src/Configurator/DockerComposeConfigurator.php#L239 `file()` and the first key value of the array this function returns is always 0.

this happens only if `version` isn't specified in docker-compose config file and `services` goes first.

CountZero1981:
Fix #877

### DIFF
--- a/src/Configurator/DockerComposeConfigurator.php
+++ b/src/Configurator/DockerComposeConfigurator.php
@@ -259,7 +259,7 @@ class DockerComposeConfigurator extends AbstractConfigurator
                 }
 
                 // Keep end in memory (check break line on previous line)
-                $endAt[$node] = '' !== trim($lines[$i - 1]) ? $i : $i - 1;
+                $endAt[$node] = !$i || '' !== trim($lines[$i - 1]) ? $i : $i - 1;
                 $node = $matches[1];
                 if (!isset($nodesLines[$node])) {
                     $nodesLines[$node] = [];


src/Configurator/DockerComposeConfigurator.php
```diff
@@ -259,7 +259,7 @@ private function configureDockerCompose(Recipe $recipe, array $config, bool $upd
                 }
 
                 // Keep end in memory (check break line on previous line)
-                $endAt[$node] = '' !== trim($lines[$i - 1]) ? $i : $i - 1;
+                $endAt[$node] = !$i || '' !== trim($lines[$i - 1]) ? $i : $i - 1;
```
nicolas-grekas, line 262: I change your fix a bit because the previous code was putting a `-1` into `$endAt[$node]`.
Can you please confirm this works for you?

CountZero1981: yes, it works

nicolas-grekas:
Thank you @CountZero1981.

PR #883, opened 2022-03-11
Fix problem of overwriting envs

### LINKED ISSUE #879: Flex changing bundles.php environments
We have a special bundles.php. Since some changes now flex seems to overwrite our `bundles.php` configurations which ends up in a none working version of @sulu as example we requires fixtures bundle be enabled in all environments and has other envs configured:

Bildschirmfoto 2022-03-03 um 14 10 54

I'm not sure which change in flex did this change. But it seems flex is ignoring exist entries in the `bundles.php` and there configuration and just overwrites them now. Which was not the case in the past.

alexander-schranz:
Fixes problem of flex overwriting bundles.php envs.

Fixes #879 which currently breaks @sulu skeleton installation.

### DIFF
--- a/src/Configurator/BundlesConfigurator.php
+++ b/src/Configurator/BundlesConfigurator.php
@@ -69,12 +69,11 @@ class BundlesConfigurator extends AbstractConfigurator
             unset($classes[$fwb]);
         }
         foreach ($classes as $class => $envs) {
-            // if the class already existed, clear so we can update the envs
-            if (isset($registered[$class])) {
-                $registered[$class] = [];
-            }
-            foreach ($envs as $env) {
-                $registered[$class][$env] = true;
+            // do not override existing configured envs for a bundle
+            if (!isset($registered[$class])) {
+                foreach ($envs as $env) {
+                    $registered[$class][$env] = true;
+                }
             }
         }
 
--- a/tests/Configurator/DockerComposeConfiguratorTest.php
+++ b/tests/Configurator/DockerComposeConfiguratorTest.php
@@ -13,7 +13,7 @@ namespace Symfony\Flex\Tests\Configurator;
 
 use Composer\Composer;
 use Composer\IO\IOInterface;
-use Composer\Package\Package;
+use Composer\Package\RootPackage;
 use PHPUnit\Framework\TestCase;
 use Symfony\Component\Filesystem\Filesystem;
 use Symfony\Flex\Configurator\DockerComposeConfigurator;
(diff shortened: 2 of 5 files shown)


alexander-schranz:
Not sure how to fix:

>  Method getPackage may not return value of type Composer\Package\Package, its declared return type is "Composer\Package\RootPackageInterface"

tests/Configurator/DockerComposeConfiguratorTest.php
```diff
@@ -114,7 +115,7 @@ class DockerComposeConfiguratorTest extends TestCase
     /** @var DockerComposeConfigurator */
     private $configurator;
 
-    /** @var Package */
```
alexander-schranz, line 117: Trying here to fix: https://github.com/symfony/flex/runs/5509940830?check_suite_focus=true

tests/Configurator/DockerComposeConfiguratorTest.php
```diff
@@ -114,7 +115,7 @@ class DockerComposeConfiguratorTest extends TestCase
...
+    /** @var PackageInterface */
```
alexander-schranz: Trying here to fix: https://github.com/symfony/flex/runs/5509940830?check_suite_focus=true

But still errors this way: https://github.com/symfony/flex/runs/5510160271?check_suite_focus=true

src/Configurator/BundlesConfigurator.php
```diff
@@ -69,12 +69,11 @@ private function configureBundles(array $bundles): array
             unset($classes[$fwb]);
         }
         foreach ($classes as $class => $envs) {
-            // if the class already existed, clear so we can update the envs
-            if (isset($registered[$class])) {
-                $registered[$class] = [];
-            }
-            foreach ($envs as $env) {
-                $registered[$class][$env] = true;
+            // do not override exist configured envs for a bundle
```
stof: typo: `existing`

tests/Configurator/DockerComposeConfiguratorTest.php
stof: the error message you see here are coming from the mock configuration. So that's unrelated to that phpdoc

stof: I think the right fix is to make it use a RootPackage instead of a Package for this property (but also for the actual value, not just for the comment)

alexander-schranz: Thx that seems to fix the tests 👍 Had todo the same for LockArrayRepository in FlexTest.

alexander-schranz: For backwards compatibility to composer 1.0.2 there a class_exists had to be added.

nicolas-grekas (approved):

fabpot (approved):

fabpot:
Thank you @alexander-schranz.

PR #895, opened 2022-04-09
[recipes:update] Fixing bug where files failed to delete that were modified previously

weaverryan:
Hi!

This fixes TWO `recipes:update` bugs:

## Bug 1️⃣ : sometimes deleted files caused patch to fail

Small bug fix. The mystery is how I didn't catch this before... and how nobody seems to have hit this. The problem is fairly simple:

A) The user gets a file (a long time ago) from a recipe (e.g. `config/bootstrap.php`).
B) The user modifies (and commits) some change.
C) A recipe update *deletes* that file.

This, oddly, fails because the patch can't be applied. For example, when `config/packages/dev/framework.yaml` is deleted in `symfony/framework-bundle` recipe, this patch is correctly generated

```diff
diff --git a/config/routes/dev/framework.yaml b/config/routes/dev/framework.yaml
deleted file mode 100644
index bcbbf13..0000000
--- a/config/routes/dev/framework.yaml
+++ /dev/null
@@ -1,3 +0,0 @@
-_errors:
-    resource: '@FrameworkBundle/Resources/config/routing/errors.xml'
-    prefix: /_error
```

However, if the user's `framework.yaml` doesn't look EXACTLY like this, the patch will fail (not with a conflict like you might expect, it just completely fails to apply). 

The fix is quite simple: if a recipe update is *deleting* a file, instead of generating a "delete patch" for it, we run `git rm <filename>`. The downside is that the user won't get a nice "file conflict" if they ever modified the file... but apparently that is not possible. And the user will still review this change before they commit.

## Bug 2️⃣  : `bundles.php` environments didn't change

If an upgraded recipe changed the environments that a bundle is configured in (e.g. https://github.com/symfony/recipes/pull/940/files), this was previously not taken into account: the update did not update the environments. Fixed now.

Tested locally on a fairly complex project.

Thanks!

### DIFF
--- a/src/Update/RecipePatcher.php
+++ b/src/Update/RecipePatcher.php
@@ -38,40 +38,15 @@ class RecipePatcher
      */
     public function applyPatch(RecipePatch $patch): bool
     {
-        if (!$patch->getPatch()) {
-            // nothing to do!
-            return true;
-        }
+        $withConflicts = $this->_applyPatchFile($patch);
 
-        $addedBlobs = $this->addMissingBlobs($patch->getBlobs());
-
-        $patchPath = $this->rootDir.'/_flex_recipe_update.patch';
-        file_put_contents($patchPath, $patch->getPatch());
-
-        try {
-            $this->execute('git update-index --refresh', $this->rootDir);
-
-            $output = '';
-            $statusCode = $this->processExecutor->execute('git apply "_flex_recipe_update.patch" -3', $output, $this->rootDir);
-
-            if (0 === $statusCode) {
-                // successful with no conflicts
-                return true;
-            }
-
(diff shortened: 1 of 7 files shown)


src/Update/RecipePatcher.php
```diff
@@ -223,4 +200,42 @@ private function getBlobPath(string $hash): string
...
+
+            throw new \LogicException('Error applying the patch: '.$this->processExecutor->getErrorOutput());
+        } finally {
+            unlink($patchPath);
+            // clean up any temporary blobs
+            foreach ($addedBlobs as $filename) {
+                unlink($filename);
+            }
+        }
+    }
```
weaverryan, line 242: Method just refactored from above, with no changes

fabpot (approved):

fabpot:
Thank you @weaverryan.

PR #903, opened 2022-04-29
Fix relative URLs in index.json not working

### LINKED ISSUE #897: remote artifactory repository for flex
Hello,
I'am wondering if we could use Symfony Flex with a remote repository in Artifactory to work as a proxy/cache for the main recipes repository in Github.
I'am using flex 1.18.5, Symfony 4 and composer 2.3.4.
I'have tried different configurations, but without success :

- I have created a generic repository in Artifacory with this URL : https://raw.githubusercontent.com/symfony/recipes/flex/main
- I have added "endpoint": ["${MyArtiRepo}/index.json"] to the composer file, but Flex was still trying to get recipes directly from Github.  
- setting SYMFONY_ENDPOINT = ${MyArtiRepo} or ${MyArtiRepo}/index.json doesn't help either.

I found this ticket in Jfrog's Jira: https://www.jfrog.com/jira/browse/RTFACT-19746 which relates to the exact same problem.

Thanks for your help.
Best Regards,
Salmen

salmen:
Hello, 
This PR fixes #897.  It relates also to #872. @nicolas-grekas  
I found out that in [index.json](https://raw.githubusercontent.com/symfony/recipes-contrib/flex/main/index.json), the name of the attribute is _recipe_template_relative_, and not _recipes_template_relative_ as expected in the source code.
And we should also set _recipe_template_ , instead of _recipes_template_.
Regards,

### DIFF
--- a/src/Downloader.php
+++ b/src/Downloader.php
@@ -224,8 +224,8 @@ class Downloader
                         break;
                     }
 
-                    if (isset($links['recipes_template_relative'])) {
-                        $links['recipes_template'] = preg_replace('{[^/\?]*+(?=\?|$)}', $links['recipes_template_relative'], $endpoint, 1);
+                    if (isset($links['recipe_template_relative'])) {
+                        $links['recipe_template'] = preg_replace('{[^/\?]*+(?=\?|$)}', $links['recipe_template_relative'], $endpoint, 1);
                     }
 
                     $urls[] = strtr($links['recipe_template'], [


nicolas-grekas:
Good catch!
I fixed the index instead, thanks for the PR.

nicolas-grekas:
Oops: https://github.com/symfony/flex/issues/904

nicolas-grekas:
Good catch, thanks @salmen.

salmen:
Glad to help @nicolas-grekas,
It seems that we have the same issue in 2.x branch, could you fix it as well ?
Are you planning to make a new release shortly ? have you any target date for 1.18.6 ? because we really need this fix asap.
Thanks

PR #908, opened 2022-05-02
Allow to disable every configurator

### LINKED ISSUE #881: Possibility to use only as symfony version resolver
Hello,

it seems like this is currently not possible to use Flex only for the symfony package version resolving (limiting? filtering?)

Would it be possible/welcomed to add a feature switch for every configurator, so you can control what parts of Flex you are interested in?

At the moment it is possible, for example, to turn off docker(-compose) files configurator:
https://github.com/symfony/flex/blob/dd7dafe3bfebf8441e167b76d8a18974585467f3/src/Configurator/DockerComposeConfigurator.php#L103

TL;DR; Looking for a solution to restrict symfony dependencies of symfony dependencies and only that and Flex looks like a good candidate for that.

marmichalski:
Howdy, @nicolas-grekas.  
As suggested, here comes a basic PR allowing to disable configurators. I am opening this in a current state to see if this approach is valid and to happily work on your feedback :sweat_smile:.

Closes #881.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,4 +2,5 @@
 /build/
 .phpunit.result.cache
 .php_cs.cache
+.php-cs-fixer.cache
 composer.lock
--- a/composer.json
+++ b/composer.json
@@ -26,6 +26,11 @@
             "Symfony\\Flex\\": "src"
         }
     },
+    "autoload-dev": {
+        "psr-4": {
+            "Symfony\\Flex\\Tests\\": "tests"
+        }
+    },
     "extra": {
         "class": "Symfony\\Flex\\Flex"
     }
--- a/src/Configurator/AbstractConfigurator.php
+++ b/src/Configurator/AbstractConfigurator.php
@@ -12,7 +12,10 @@
 namespace Symfony\Flex\Configurator;
 
 use Composer\Composer;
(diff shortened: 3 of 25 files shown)


lzakrzewski (approved):
LGTM, thanks for addressing this :+1:

nicolas-grekas:
I didn't expect this approach. I don't think it makes sense to disable configurators case by case as that's going to provide broken experiences, where only part of a recipe is applied. Docker support is a bit special since that's not about the app itself.

Instead, I was expecting a single flag to disable all recipes.

(Note that this can already be achieved without patching flex by creating a recipe repo that ships no recipes.)

marmichalski:
> I didn't expect this approach. I don't think it makes sense to disable configurators case by case as that's going to provide broken experiences, where only part of a recipe is applied. Docker support is a bit special since that's not about the app itself.

Okay :cry: 

> Instead, I was expecting a single flag to disable all recipes.

Sure, I'll change this PR into single-flag-to-rule-them-all then, thanks.

maxhelias:
I don't see the added value even to disable all the recipes

fabpot:
Closing as there has been no activity for a long time.

PR #916, opened 2022-05-27
check for AbstractBundle

tacman:


src/SymfonyBundle.php
```diff
@@ -107,6 +107,9 @@ private function isBundleClass(string $class, string $path, bool $isPsr4): bool
         }
 
         // heuristic that should work in almost all cases
-        return false !== strpos(file_get_contents($classPath), 'Symfony\Component\HttpKernel\Bundle\Bundle');
+        $classContents = file_get_contents($classPath);
+
+        return strpos($classContents, 'Symfony\Component\HttpKernel\Bundle\Bundle')
```
stof: This should still use `false !== strpos()` as the way to check if the string contains a substring, to be safe (even though a PHP file will not actually contain this at index `0`, it makes the intent less clear)

tacman: so 

(false !== strpos($classContents, 'Symfony\Component\HttpKernel\Bundle\Bundle') || 
    false !== strpos($classContents, 'Symfony\Component\HttpKernel\Bundle\AbstractBundle')) 

Too bad we're not on php8 yet, str_contains is easier to read.

Or we could check for just the root of the namespace, e.g. just look for Symfony\Component\HttpKernel\Bundle.

Really it feels like we should be checking if the class implements BundleInterface, but not sure if that's worth the effort.

stof: yeah, use the 2 `false !== strpos()`.
And yes, I agree `str_contains` is easier to read. This can be cleaned when merging up in the 2.x branch. But 1.x needs to support PHP 7.

tacman: Would this work:

        return is_subclass_of($class, 'Symfony\Component\HttpKernel\Bundle\BundleInterface');

tacman: I think my PR is (finally) right, but tests are failing on php 7.3:

https://github.com/symfony/flex/runs/6644280992?check_suite_focus=true

nicolas-grekas (approved):

nicolas-grekas:
Thank you @tacman.

src/SymfonyBundle.php
stof: The `is_subclass_of` would not work because this code runs before the autoloader is generated, so we cannot reliably load the app code.

PR #917, opened 2022-05-27
[POC][WIP] Pack recipes

kbond:
Proposal for "scaffolds" as recipes as discussed with @nicolas-grekas and @weaverryan as a possible alternative to https://github.com/symfony/maker-bundle/pull/1085.

- Scaffolds make sense to be packs as we want their dependencies installed but not the "empty package".
- Since scaffolds contain both require/require-dev dependencies, I had to modify the unpacker to allow for this.
- We want pack recipes installed but not unconfigured when the pack is uninstalled.
- Pack/Scaffold recipes cannot be updated/uninstalled. I think this makes sense as they will likely be heavily modified.

Recipe PR: https://github.com/symfony/recipes/pull/1088

**Test:**

```bash
symfony new project
cd project
export SYMFONY_ENDPOINT=https://raw.githubusercontent.com/symfony/recipes/flex/pull-1088/index.json
composer config repositories.homepage vcs git@github.com:kbond/homepage-scaffold
composer config repositories.flex vcs git@github.com:kbond/flex
composer require symfony/flex:dev-pack-recipes
composer require kbond/homepage-scaffold # installs the "scaffold"
vendor/bin/phpunit # 1 test should run & pass
```

**TODO:**
- [ ] If installing a scaffold with `--dev`, you get all the deps in require-dev.
- [ ] Way to determine if a recipe is a pack on unconfigure.
- [ ] Higher priority for pack recipes. This will allow files in pack recipes to take precedence over their dependency recipes (like `metapackages` do now).

### DIFF
--- a/src/Configurator.php
+++ b/src/Configurator.php
@@ -72,6 +72,10 @@ class Configurator
 
     public function unconfigure(Recipe $recipe, Lock $lock)
     {
+        if ('kbond/homepage-scaffold' === $recipe->getPackage()->getName()) {
+            return;
+        }
+
         $manifest = $recipe->getManifest();
         foreach (array_keys($this->configurators) as $key) {
             if (isset($manifest[$key])) {
--- a/src/Unpacker.php
+++ b/src/Unpacker.php
@@ -73,14 +73,6 @@ class Unpacker
             }
             $devRequires = $pkg->getDevRequires();
 
-            foreach ($devRequires as $i => $link) {
-                if (!isset($requires[$link->getTarget()])) {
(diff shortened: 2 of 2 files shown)


src/Configurator.php
```diff
@@ -72,6 +72,10 @@ public function populateUpdate(RecipeUpdate $recipeUpdate): void
 
     public function unconfigure(Recipe $recipe, Lock $lock)
     {
+        if ('kbond/homepage-scaffold' === $recipe->getPackage()->getName()) {
```
kbond, line 75: I could not figure out how to determine if the recipe is a pack or not here. `$recipe->getPackage()->getType === null`.

kbond:
Closing in favour of #923.

PR #924, opened 2022-06-02
Allow packs to embed their recipe

nicolas-grekas:
By making packs able to embed their recipe, we enable using them for scaffolding feature-full applications.

Note that such local recipes cannot be managed with any `composer recipes:*` command, and this is a feature :)

At the community scale, this might allow ppl to work around the central repository of recipes if authors maintain a pack next to their main lib. But this looks fair to me, especially with the noted limitation of packs' recipes. We do want to allow ppl to create scaffolds, even if it's for one lib.

(Note also that since #923, recipes bound to packs via the central repository are now update-able, so we don't need the `*-meta` repos anymore.)

On a technical level, this works as is:
1. `composer require the/pack` installs the pack and all its deps using the regular composer flow,
2. flex unpacks `the/pack` into the root `composer.json`,
3. flex turns the files in the `vendor/the/pack/` directory into a recipe and installs it,
4. flex tells composer to remove `the/pack`, thus removing `vendor/the/pack/`

:magic_wand:

One last note: embedded recipes escape the CI we have in place on the central repo. We might want to provide a callable GHA to make it easier for authors to verify them. This GHA could also check conflicts between the central repo and packs' recipes (overridden files typically), to ease propagate changes from the central repo to the overriding packs.

### DIFF
--- a/src/Downloader.php
+++ b/src/Downloader.php
@@ -18,6 +18,7 @@ use Composer\DependencyResolver\Operation\UninstallOperation;
 use Composer\DependencyResolver\Operation\UpdateOperation;
 use Composer\IO\IOInterface;
 use Composer\Json\JsonFile;
+use Composer\Package\PackageInterface;
 use Composer\Util\Http\Response as ComposerResponse;
 use Composer\Util\HttpDownloader;
 use Composer\Util\Loop;
@@ -51,6 +52,7 @@ class Downloader
     private $caFile;
     private $enabled = true;
     private $composer;
+    private $vendorDir;
 
     public function __construct(Composer $composer, IoInterface $io, $rfs)
     {
@@ -94,6 +96,7 @@ class Downloader
         $this->cache = new ComposerCache($io, $config->get('cache-repo-dir').'/flex');
         $this->sess = bin2hex(random_bytes(16));
         $this->composer = $composer;
+        $this->vendorDir = $config->get('vendor-dir');
     }
 
     public function getSessionId(): string
@@ -155,6 +158,7 @@ class Downloader
             $this->conflicts = [];
         }
 
+        $localRecipes = [];
         $data = [];
         $urls = [];
         $chunk = '';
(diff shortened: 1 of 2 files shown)


lyrixx:
Why allowing to install a recipes from a pack but not from a regular composer package (a lib for instance)?

nicolas-grekas:
For the same reasons I gave in #572 and #753

MichaelBrauner:
So if I then write a blog bundle, I would need an external pack repository instead of an external bundle repository? Have I understood that correctly?

nicolas-grekas:
@MichaelBrauner this is already the situation yes. And while the "symfony distribution" argument doesn't stand for bundles I agree, there are other reasons to keep the central repo I mentioned there. Eg the validation of recipes, the review steps to ensure a consistent ecosystem, and more recently the `recipes:update` command which needs the recipe API. For bundles specifically, note that we encourage them NOT requiring any recipe. That's not always possible of course, but that should still be a principle: bundles should provide sane defaults that make a recipe not needed for them.

MichaelBrauner:
> Eg the validation of recipes, the review steps to ensure a consistent ecosystem

Are the areas of application not already limited by the range of functions flex offers? 
Is it a good idea to control and limit bundle creators at the expense of simplicity?

As an example. I don't see any use-case for me, but someone would maybe just copy some Entities and FormTypes. When this is a good idea in a specific recurring scenario. Why should it be restricted?

When I use composer packs now - I can do that without control and restriction, isn't it? 
So why restrict and control at all, when there is already a workaround with this new PR.

> and more recently the recipes:update command which needs the recipe API.

Yes, that's a good one 👍 ! 
I think the longer a system exists in its form, the more depends on it. But is that a reason not to think about fundamental improvements when it would make that big of an impact on the community.

> For bundles specifically, note that we encourage them NOT requiring any recipe. That's not always possible of course, but that should still be a principle: bundles should provide sane defaults that make a recipe not needed for them.

I can only agree with that. 
But should recommendations be encouraged by making recipe creation more complicated than it could be?

What would offer more to the Symfony ecosystem and make it even more attractive? 
Control over the quality and purposes of the recipes - or freedom and flexibility for creators?

And what does the community want? Are there many developers who would like to keep Flex as it is now - or are there mostly developers who would like it to be more direct and simple - without decoupled Git repositories?

nicolas-grekas:
On second thought and after talking with a few others, we don't need this for the scaffolding feature we'd like to work on.
Closing therefore.

MichaelBrauner:
Ok. Thank you for your wonderful work .

PR #925, opened 2022-06-02
Remove overriding native composer commands

nicolas-grekas:
That doesn't work since Composer v2.1.6 and this commit anyway:
https://github.com/composer/composer/commit/9727adf63b0a6a7f082d1da572eff42e52e2dab1

For `update` and `remove`, these were no-op already, because alias resolution happens early anyway (see removed comment, which says this is for composer 1.x, while it's also why 2.x works)

For `require`, this removes #783, aka `require-dev` is now removed by composer if it's empty, while we'd prefer keeping it so that's its position is preserved when adding dev-deps again.

/cc @Seldaek FYI and in case you have any thoughts related to these behaviors.


Seldaek:
Less hackery 👍👍

nicolas-grekas:
We might be fine with that but for the record I just figured out that this broke resolving package aliases when using composer 2.0.0 to 2.1.5 included.

PR #930, opened 2022-06-08
Add configurator for "composer-commands" entries in recipes

### LINKED ISSUE #901: Real composer scripts
While using in the manifest `composer-scripts` I just assumed its the `scripts` part of the `composer.json` and thats currently not 😅 .

It would be nice to be able to define them like 

```json
{
"composer-commands": {
"build-js": "npm run build"
}
}
```

and then I can run them with `composer run build-js`. This works already out of the box with Composer when its in the correct key set 😅

morawskim:
This PR closes #901 

At the moment in recipe we can use `composer-scripts`, but those commands will be running only during `composer install` or `composer update`.

This PR adds a new configuration option `composer-commands`.
We can add to recipe:

```
.....
"composer-commands": {
    "command": "vendor/bin/foo"
}
```
and flex will add to `composer.json` necessary entry and we will able to call `composer run command` to execute `vendor/bin/foo`.

This behavior is very similar to current Makefile configurator as mentioned in #901.

### DIFF
--- a/src/Configurator.php
+++ b/src/Configurator.php
@@ -43,6 +43,7 @@ class Configurator
             'container' => Configurator\ContainerConfigurator::class,
             'makefile' => Configurator\MakefileConfigurator::class,
             'composer-scripts' => Configurator\ComposerScriptsConfigurator::class,
+            'composer-commands' => Configurator\ComposerCommandsConfigurator::class,
             'gitignore' => Configurator\GitignoreConfigurator::class,
             'dockerfile' => Configurator\DockerfileConfigurator::class,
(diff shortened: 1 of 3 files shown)


morawskim:
I will fix the build "PHP 7.4 low-deps" at weekend.

maxhelias (approved):

alamirault (approved):
Thanks for the PR, I also need to add commands to script section. I hope it can be merged.

README should probably be updated to document the nex configurator (In another PR ?)

nicolas-grekas (commented):
LGTM thanks! Sorry for the late review. I rebase to 2.x

nicolas-grekas (approved):

PR #931, opened 2022-06-10
Ignore correct PHP CS Fixer config file in .gitattributes

jmsche:


stof (approved):

fabpot:
Good catch, thanks @jmsche.

Issue #864, opened 2022-02-03
Updating Recipe Fails When Using Git Sub-Modules


PythooonUser:
### Symfony version(s) affected

5.4.4

## Project Setup

We are using a project structure that leverages git sub-modules, i.e.
```
root
  -> project1
  -> project2 (checked-out)
```

The `.git` "folder" is actually a file thereof, pointing to the "parent" `.git` folder:
```
# cat root/project2/.git
gitdir: ../.git/modules/project2
```

## Problem
When I now run the command to update a recipe (it doesn't matter which one it is), I'm greeted with this error:

```
user@computer:/root/project2$ composer recipes:update symfony/security-bundle
  Updating recipe for symfony/security-bundle...

There was an error applying the recipe update patch
Failed to create "./.git/objects/e9": mkdir(): Not a directory.

Update the recipe by re-installing the latest version with:
  composer recipes:install symfony/security-bundle --force -v
```

**Issue:** Composer tries to work with the `.git` folder as usual, not being aware of the concept about git sub-modules.

The provided solution here is to just nuke the existing code and "start fresh". That works of course, but could be "cleaner".

### How to reproduce

I think the description says enough already, but:
- Create a git repository
- Add a sub-module with a Symfony project
- Try to update a recipe

### Possible Solution

Not sure about a solution here.

We would probably need to make composer aware about the git sub-module concept and make it follow the link to the "actual" `.git` folder.

But the provided solution to just "nuke" the existing code is also fine, when the above cannot be achieved easily.

nicolas-grekas:
Would you mind working on a fix and sending a PR?

PythooonUser:
Sure, I can try. Could you notch me into the right direction?

nicolas-grekas:
src/Update/RecipePatcher.php might be a good start, that's where the .git folder is referenced.

weaverryan:
I’d be happy to help also - you can also ping me on symfony slack. It seems like you’re likely correct that the code needs to find the “real” .git directory. 

I’m guessing this exact error comes from here: https://github.com/symfony/flex/blob/1.x/src/Update/RecipePatcher.php#L195 - but you could get a better stack trace if you update with -vvv. In that class, you can see a lot of references to rootDir. It’s possible that this needs to change in some or all cases - but I’m doing a lot of guessing :)

PythooonUser:
Okay, I tried creating a reproducer first. Another thing that I found out:

### Install from sub-project
[...]

PR #937, opened 2022-07-13
Updating Recipe Fails When Project Located in Subfolder in Git Repository

### LINKED ISSUE #918: Updating Recipe Fails When Project Located in Subfolder in Git Repository
### Description
Updating Recipes is not fully applied when symfony project is located in sub-folder of the git repository.

Recipe update runs without errors, Changelog is calculated properly and some changes (file deletions and symfony.lock) are shown. Everything looks fine, unless you know that modifications are actually missing. (Like when [following course video](https://symfonycasts.com/screencast/symfony6-upgrade/framework-bundle-recipe))

Repository layout:
```
root
-> projectB
```

### How to reproduce
* Create new symfony 4.4 project in subfolder of a git repo
* Update symfony to 5.4 (replace `4.4.*` with `5.4.*` in composer.json and run composer update, commit changes)
* Run `composer recipes:update symfony/framework-bundle` (or any recipe update that should modify files)

Result:
```
$ git status
On branch main
Changes to be committed:
  (use "git restore --staged <file>..." to unstage)
 	deleted:    projectB/config/bootstrap.php
	deleted:    projectB/config/packages/test/framework.yaml
	deleted:    projectB/config/routes/dev/framework.yaml
	modified:   projectB/symfony.lock
```
Expected Result:
```
$ git status
On branch main
Changes to be committed:
  (use "git restore --staged <file>..." to unstage)
	modified:   projectB/.env
	deleted:    projectB/config/bootstrap.php
	modified:   projectB/config/packages/framework.yaml
	deleted:    projectB/config/packages/test/framework.yaml
	modified:   projectB/config/preload.php
	deleted:    projectB/config/routes/dev/framework.yaml
	new file:   projectB/config/routes/framework.yaml
	modified:   projectB/config/services.yaml
	modified:   projectB/public/index.php
	modified:   projectB/src/Kernel.php
	modified:   projectB/symfony.lock
```

### Possible reason and fix
It looks like patch generation ignores the subfolder. And .git -folder location is assumed always the same. 
I think `src-prefix` and `dst-prefix` should be added to patch generation. And .git -folder location could be asked from git binary (`git rev-parse --git-dir`)

I do understand that this scenario is not very common, but outcome was very confusing with SymfonyCasts video where you can see the expected result, but locally get the other. No errors or notifications that update did not go expected.

Peukku:
Closes #918 and #864 

Updating Recipe fails when symfony project is not in the root of git project. 
As a developer the most problematic thing is that it fails without any errors. Some files are just not updated.

There is a simple project with commands needed to reproduce to see what is happening before the fix: https://github.com/Peukku/symfony-flex-reproducer-918-project-root/

I think there are two reasons for this to happen:
* $patchString is created with git diff in tmpPath, which has a temporary flat git repository (with no path prefix), but applied bit later in project context where path would be needed. 
* Location of .git directory is assumed to be in project root, when it actually is in git root level. (or in .git/modules/[subproject] when using git submodules). The blobs do get written, but in the wrong location.

This PR uses git binary to resolve location of .git folder and current relative path to be used as the prefix for git diff.

### DIFF
--- a/src/Update/RecipePatcher.php
+++ b/src/Update/RecipePatcher.php
@@ -168,9 +168,8 @@ class RecipePatcher
     private function addMissingBlobs(array $blobs): array
     {
         $addedBlobs = [];
-        $gitDir = trim($this->execute('git rev-parse --absolute-git-dir', $this->rootDir));
         foreach ($blobs as $hash => $contents) {
-            $blobPath = $gitDir.'/'.$this->getBlobPath($hash);
+            $blobPath = $this->getBlobPath($this->rootDir, $hash);
             if (file_exists($blobPath)) {
                 continue;
             }
@@ -188,7 +187,6 @@ class RecipePatcher
     private function generateBlobs(array $originalFiles, string $originalFilesRoot): array
     {
         $addedBlobs = [];
-        $originalFilesGitDir = trim($this->execute('git rev-parse --absolute-git-dir', $originalFilesRoot));
         foreach ($originalFiles as $filename => $contents) {
             // if the file didn't originally exist, no blob needed
             if (!file_exists($originalFilesRoot.'/'.$filename)) {
@@ -196,18 +194,20 @@ class RecipePatcher
             }
 
             $hash = trim($this->execute('git hash-object '.ProcessExecutor::escape($filename), $originalFilesRoot));
-            $addedBlobs[$hash] = file_get_contents($originalFilesGitDir.'/'.$this->getBlobPath($hash));
+            $addedBlobs[$hash] = file_get_contents($this->getBlobPath($originalFilesRoot, $hash));
         }
 
         return $addedBlobs;
     }
 
-    private function getBlobPath(string $hash): string
+    private function getBlobPath(string $gitRoot, string $hash): string
     {
+        $gitDir = trim($this->execute('git rev-parse --absolute-git-dir', $gitRoot));
+
         $hashStart = substr($hash, 0, 2);
         $hashEnd = substr($hash, 2);
 
-        return 'objects/'.$hashStart.'/'.$hashEnd;
+        return $gitDir.'/objects/'.$hashStart.'/'.$hashEnd;
     }
 
     private function _applyPatchFile(RecipePatch $patch)
--- a/src/Update/RecipePatcher.php
+++ b/src/Update/RecipePatcher.php
@@ -83,6 +83,8 @@ class RecipePatcher
             }
         }
 
+        // Use git binary to get project path from repository root
+        $prefix = trim($this->execute('git rev-parse --show-prefix', $this->rootDir));
         $tmpPath = sys_get_temp_dir().'/_flex_recipe_update'.uniqid(mt_rand(), true);
         $this->filesystem->mkdir($tmpPath);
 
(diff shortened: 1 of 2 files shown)


src/Update/RecipePatcher.php
```diff
@@ -203,7 +207,7 @@ private function getBlobPath(string $hash): string
         $hashStart = substr($hash, 0, 2);
         $hashEnd = substr($hash, 2);
 
-        return '.git/objects/'.$hashStart.'/'.$hashEnd;
+        return 'objects/'.$hashStart.'/'.$hashEnd;
```
weaverryan: Small code simplification. Move `$originalFilesGitDir = trim($this->execute('git rev-parse --absolute-git-dir', $originalFilesRoot));` into THIS function, then return:

```php
return $originalFilesGitDir.'/objects/'.$hashStart.'/'.$hashEnd;
```

If you do this, then the two methods above need zero changes (their diffs can be reverted).

weaverryan (commented):
Big apologies for the slow review on this! Amazing then, when I hit this problem today, you had a PR ready for this complex problem ❤️ .

I've just tried this code and it works beautifully.

It would also be awesome to add a simple test case for this in https://github.com/symfony/flex/blob/1.x/tests/Update/RecipePatcherTest.php - you could duplicate `testApplyPatch`, but make it work without a data provider (just initialize the git repo, add a file, then try to apply some simple patch from a sub-directory).

Peukku:
Thank you for the feedback!I'm extremely busy this week, but i will continue after that

src/Update/RecipePatcher.php
Peukku: Hmm, this seems not to work in case of subfolders.
I think the reason is that this getBlobPath() is called from two different places which are refering to different git paths:

- addMissingBlobs() is modifying the current project in $this->rootDir (git rev-parse resulting to either .git or ../.git/modules/projectName)
- generateBlobs() is manually generating the blobs for patch in temp location $tmpPath (git rev-parse always resulting to .git). 

I'll try a bit different approach simplifying the code

weaverryan:
I see some new work on this - awesome :). I don't know if you're finished, yet, but 2 small things:

A) Please rebase the PR to remove the merge commit
B) On fabbot, there is a curl command you can run to tweak any coding standards - d429f654e42ab691c1fb684d99501d8fcf8903d

Thanks!

Peukku:
Extremely busy week turned into two .. But now I was finally able to continue with this!

I moved the "git rev-parse" into getBlobPath() as you suggested, makes much more sense that getBlobPath() returns the full path. The full blob path consists of two parts; git root path (which can be project repository or the temp location used in creating the patch) and the hash. These are now the arguments for the getBlobPath(string $gitRoot, string $hash).

I also made the duplicate of testApplyPatch; testApplyPatchOnSubfolder. And made it use the same dataprovider as the original one. Same files and patches can be used, but subfolder had to be added in the generated patch. 
For that I made the dataprovider getApplyPatchTests to check the name of the test method calling it to prefix the patches with required path.

weaverryan (approved):
I just tested this locally and it works! I also tested on a traditional project (where the git root is the project root) and there were no problems there either. This should be merged. Great work and big thanks @Peukku 👍

fabpot:
Thank you @Peukku.

PR #942, opened 2022-08-04
Fixing bug where removed parameters were not handled correctly during recipes update

weaverryan:
Hi again!

Reported at https://symfonycasts.com/screencast/symfony6-upgrade/upgrade-recipes2#comment-5935619544

Just an oversight - the original implementation was a bit naive, not taking into account a situation where the `container` configurator in the OLD recipe contained a parameter, but this parameter was *removed* in the updated recipe.

Cheers!


maxhelias (approved):

fabpot (approved):

fabpot:
Thank you @weaverryan.

PR #947, opened 2022-09-09
Fixes SYMFONY_REQUIRE environment variable [2.x]

ncoleshaw-oncorps:
Closes https://github.com/symfony/flex/issues/946 (`2.x`)

When requiring a package in `symfony/symfony`, Flex attempts to determine the `symfony.require` version in two different ways:
- `Symfony\Flex\PackageResolver`
  - Uses the value of `extra.symfony.require` defined in the root `composer.json`; or
  - Uses the version constraint for `symfony/framework-bundle` defined in the root `composer.json`; or
  - Applies no constraint (`SYMFONY_REQUIRE` environment variable is ignored)
- `Symfony\Flex\Flex`
  - Uses the value of `SYMFONY_REQUIRE` environment variable; or
  - Uses the in-memory value of the root package's `extra.symfony.require`; or
  - Applies no constraint (`symfony/framework-bundle` version constraint is ignored)

As a result of the above, when requiring packages in `symfony/symfony`, Flex may report a package restriction without applying it.

This PR ensures `Symfony\Flex\PackageResolver` and `Symfony\Flex\Flex` both use the same `symfony.require` version, determined in the following priority order:
- `SYMFONY_REQUIRE` environment variable; or
- In-memory value of `extra.symfony.require`; or
- In-memory value of the `symfony/framework-bundle` version constraint in the `requires` list (unless the `symfony/framework-bundle` version constraint given is `*`); or
- No version constraint applies

PR #948, opened 2022-09-09
Fixes SYMFONY_REQUIRE environment variable [1.x]

### LINKED ISSUE #946: Symfony package versions not restricted by SYMFONY_REQUIRE environment variable
The `flex` plugin reads the value of `SYMFONY_REQUIRE` from the environment and may use this value to report that packages in "symfony/symfony" have been restricted, but `PackageResolver` ignores this environment variable and instead attempts to obtain the `symfony.require` version directly from the root package's `composer.json` file.

Steps to reproduce:
- Set `SYMFONY_REQUIRE` environment variable
  - e.g. `export SYMFONY_REQUIRE=5.4.*`
- Verify that the variable is set
  - `printenv | grep SYMFONY_REQUIRE`
- Make sure `extra.symfony.version` is not set in the root `composer.json`
- Make sure `symfony/framework` is not in the `requires` list in the root `composer.json`
- Require a package in `symfony/symfony`
  - e.g. `composer require symfony/serializer`

Expected result:
For the above example, `symfony/serializer 5.4.*` is added to the `requires` list, and the latest patch version corresponding to that version constraint is installed. 

Actual result:
For the above example, Flex reports `Restricting packages listed in "symfony/symfony" to "5.4.*"` but the most recent available version of `symfony/serializer` (`^6.1` at the time of writing) is installed.

This issue appears to be referenced in https://github.com/symfony/flex/issues/933

ncoleshaw-oncorps:
Closes https://github.com/symfony/flex/issues/946 (`1.x`)

Backports https://github.com/symfony/flex/pull/947/files to `1.x`

### DIFF
--- a/src/Flex.php
+++ b/src/Flex.php
@@ -103,6 +103,7 @@ class Flex implements PluginInterface, EventSubscriberInterface
         'unpack' => true,
     ];
     private $filter;
+    private $symfonyRequire = '';
 
     public function activate(Composer $composer, IOInterface $io)
     {
@@ -126,15 +127,15 @@ class Flex implements PluginInterface, EventSubscriberInterface
         $this->config = $composer->getConfig();
         $this->options = $this->initOptions();
 
-        $symfonyRequire = preg_replace('/\.x$/', '.x-dev', getenv('SYMFONY_REQUIRE') ?: ($composer->getPackage()->getExtra()['symfony']['require'] ?? ''));
+        $this->symfonyRequire = preg_replace('/\.x$/', '.x-dev', $this->symfonyRequireVersion($this->composer));
 
         if ($composer2 = version_compare('2.0.0', PluginInterface::PLUGIN_API_VERSION, '<=')) {
             $rfs = Factory::createHttpDownloader($this->io, $this->config);
 
             $this->downloader = $downloader = new Downloader($composer, $io, $rfs);
 
-            if ($symfonyRequire) {
-                $this->filter = new PackageFilter($io, $symfonyRequire, $this->downloader);
+            if ($this->symfonyRequire) {
+                $this->filter = new PackageFilter($io, $this->symfonyRequire, $this->downloader);
             }
 
             $setRepositories = null;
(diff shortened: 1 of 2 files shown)


maxhelias:
I didn't see this one before : https://github.com/symfony/flex/pull/947. 

This PR should be enough and it will go into 2.x at merge. And some tests would be useful

ncoleshaw-oncorps:
@maxhelias Thank you for your review. It may be better to close this PR as `Flex.php` has some differences between `1.x` and `2.x` that affect the way my proposal is implemented. It may therefore be simpler for this to be a `2.x` and later suggestion?

maxhelias:
I don't know how difficult it is to maintain the 1.x and 2.x branches together so let the Symfony Team decide

src/Flex.php
```diff
@@ -241,7 +245,6 @@ class_exists(__NAMESPACE__.str_replace('/', '\\', substr($file, \strlen(__DIR__)
             } elseif ('update' === $command) {
                 $this->displayThanksReminder = 1;
             } elseif ('outdated' === $command) {
-                $symfonyRequire = null;
```
nicolas-grekas, line 244: `composer outdated` should report the non-restricted packages. This breaks it.

src/Flex.php
```diff
@@ -1103,4 +1106,27 @@ private function doesRecipeConflict(array $recipeData, OperationInterface $opera
...
+        $package = $composer->getPackage();
+        $version = getenv('SYMFONY_REQUIRE') ?: '';
+
+        if (empty($version)) {
+            $version = $package->getExtra()['symfony']['require'] ?? '';
+        }
+
+        if (empty($version)) {
+            $require = $package->getRequires();
+            $link = $require['symfony/framework-bundle'] ?? null;
```
nicolas-grekas, line 1121: This means that having framework-bundle in your deps will restrict the possible symfony packages.
This is not how it behaves right now, nor what we want.

nicolas-grekas (commented):
Thanks for the PR, but I cannot reproduce the issue you describe, and the patch doesn't make sense to me (yet). Can you please provide more specific instructions to reproduce what you have so that I can really understand what the problem is?

PR #950, opened 2022-09-26
Resolve directories for copy from recipe in update case

shyim:
The update case of `CopyFromRecipeConfigurator` does not respect the folders in the config. 
https://symfony.com/doc/current/setup/flex.html#customizing-flex-paths

### DIFF
--- a/tests/Configurator/CopyFromRecipeConfiguratorTest.php
+++ b/tests/Configurator/CopyFromRecipeConfiguratorTest.php
@@ -190,6 +190,84 @@ class CopyFromRecipeConfiguratorTest extends TestCase
         $this->assertSame($newRecipeFiles, $recipeUpdate->getNewFiles());
     }
 
+    public function testUpdateResolveDirectories()
+    {
+        $configurator = $this->createConfigurator();
+
+        $lock = $this->createMock(Lock::class);
+        $lock->expects($this->once())
+            ->method('add')
+            ->with(
+                'test-package',
+                [
+                    'files' => [
+                        'config/packages/framework.yaml',
+                        'test.yaml',
+                    ],
+                ]
+            );
+
+        $originalRecipeFiles = [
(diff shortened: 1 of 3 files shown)


maxhelias (approved):

stof:
This should add tests covering this behavior (in CopyFromRecipeConfiguratorTest) to prevent regressions.

shyim:
the 7.4 fails because of packagist woofoo with composer 1

tests/Configurator/CopyFromRecipeConfiguratorTest.php
```diff
@@ -190,6 +190,84 @@ public function testUpdate()
...
+            $newRecipe,
+            $lock,
+            FLEX_TEST_DIR
+        );
+
+        $configurator->update(
+            $recipeUpdate,
+            [
+                'root/' => '',
+                'symfony8config/' => '%CONFIG_DIR%/',
```
stof, line 250: Having the recipe using `symfony8config` and mapping it to the usual config dir named `config` is not at all the use case that you want to solve (it is exactly the opposite use case). I suggest making the test cover the intended use case instead, as that's the use case for which we want to prevent regressions.

shyim: I wanted to test, when the path is empty or you mapped to it %SOME-VAR%

I can remove it if you want. mapping `config` to `%CONFIG_DIR%` makes no difference

Ocarthon:
@stof Does this feature have the potential to be merged and released? I just encountered the same bug with our internal recipes.

If yes, i would also create a PR for flex 2.x (if @shyim agrees)

shyim:
It will be backported when merged I think automatically

nicolas-grekas (approved):
Let's ship this fix.
PR targeting 2.x as 1.x is not maintained anymore.

nicolas-grekas:
Thank you @shyim.

PR #959, opened 2022-11-08
Allow updating recipes without Git project

shyim:
We want to migrate at Shopware our Zip Based Installation users to Composer. For that reason, we are building an own application which updates a project using Composer on shared hosters. 

People using Zip files to deploy their stuff are not using Git for their project 🙈 . So I just added the option to use normal Unix `patch` to patch the files of the changed files instead of expecting a Git repository there.

The Git way is still supported. I just fallback to `patch` when a git project is not present

### DIFF
--- a/src/Command/UpdateRecipesCommand.php
+++ b/src/Command/UpdateRecipesCommand.php
@@ -27,7 +27,8 @@ use Symfony\Flex\GithubApi;
 use Symfony\Flex\InformationOperation;
 use Symfony\Flex\Lock;
 use Symfony\Flex\Recipe;
-use Symfony\Flex\Update\RecipePatcher;
+use Symfony\Flex\Update\GitRecipePatcher;
+use Symfony\Flex\Update\PlainRecipePatcher;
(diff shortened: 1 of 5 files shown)


stof:
The git-based workflow relies on git to provide a way to review the changes done by applying the new recipes. How do you plan to do that in a workflow without git ? Reviewing such changes is an important step of the workflow (as changing the configuration of packages in an existing project will have impacts not covered by the BC policy, precisely because you change the config)

shyim:
Does it maybe make sense to output the list of updated files as a warning? 🤔 So the user should review them?

shyim:
I will look for an better way maybe and may come back 🙈

stof:
When using git, `git diff` gives you exactly that.

shyim:
I guess I will go the way of `recipe:install --force --reset` and document that the configs will be resettet when done using web updater 😅 

Just merge the old `.env` of the user with the new one 😅 

I guess only non tech people will use that without git anyway  😅

PR #960, opened 2022-11-21
Correctly resolve the lock alias

pierredup:
`composer require lock` doesn't resolve the `lock` alias because the package name resolution is skipped when the package argument is `lock`. This PR bypasses this specifically for the `require` command, so that `composer require lock` correctly resolved the alias, but `composer update lock` still skips the name resolution.

Reference: https://github.com/symfony/recipes/issues/1150


nicolas-grekas (approved):

kbond (approved):

ogizanagi (approved):

fabpot:
Thank you @pierredup.

PR #961, opened 2022-12-09
Fix recipe history url

rosier:
Recipes that are installed from the branch master now point to a history url that results in a 404.

Before:
flex-before

After:
flex-after

### DIFF
--- a/src/Command/RecipesCommand.php
+++ b/src/Command/RecipesCommand.php
@@ -165,6 +165,10 @@ class RecipesCommand extends BaseCommand
         $lockBranch = $recipeLock['recipe']['branch'] ?? null;
         $lockVersion = $recipeLock['recipe']['version'] ?? $recipeLock['version'] ?? null;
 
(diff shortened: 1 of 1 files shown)


src/Command/RecipesCommand.php
```diff
@@ -165,6 +165,10 @@ private function displayPackageInformation(Recipe $recipe)
         $lockBranch = $recipeLock['recipe']['branch'] ?? null;
         $lockVersion = $recipeLock['recipe']['version'] ?? $recipeLock['version'] ?? null;
 
+        if ('master' === $lockBranch && 'github.com/symfony/recipes' === $lockRepo) {
```
maxhelias: We should also do it for the recipes-contrib :
```suggestion
        if ('master' === $lockBranch && \in_array($lockRepo, ['github.com/symfony/recipes', 'github.com/symfony/recipes-contrib'])) {
```

maxhelias (approved):

xabbuh (approved):

weaverryan (approved):

nicolas-grekas:
Thank you @rosier.

PR #963, opened 2023-01-05
Add --yes flag to recipes:install command to overwrite all files without asking

shyim:
When I run `recipes:install --reset --force -n` **no files are overwritten**, as the default as been set to false.

There is currently also no way to accept that, so I changed it back to `true`

### DIFF
--- a/src/Options.php
+++ b/src/Options.php
@@ -62,7 +62,7 @@ class Options
         return file_exists($rootDir.'/'.$otherPhpunitDistFile) ? $otherPhpunitDistFile : $result;
     }
 
-    public function shouldWriteFile(string $file, bool $overwrite): bool
+    public function shouldWriteFile(string $file, bool $overwrite, bool $skipQuestion): bool
     {
         if (isset($this->writtenFiles[$file])) {
             return false;
@@ -78,6 +78,10 @@ class Options
         }
 
         if (!filesize($file)) {
+            return true;
+        }
+
+        if ($skipQuestion) {
             return true;
         }
 
--- a/tests/Command/InstallRecipesCommandTest.php
+++ b/tests/Command/InstallRecipesCommandTest.php
@@ -1,0 +1,42 @@
+<?php
+
+/*
+ * This file is part of the Symfony package.
+ *
+ * (c) Fabien Potencier <fabien@symfony.com>
+ *
+ * For the full copyright and license information, please view the LICENSE
+ * file that was distributed with this source code.
+ */
+
+namespace Symfony\Flex\Tests\Command;
+
+use Composer\Console\Application;
+use PHPUnit\Framework\TestCase;
+use Symfony\Component\Console\Tester\CommandTester;
+use Symfony\Flex\Command\InstallRecipesCommand;
+use Symfony\Flex\Event\UpdateEvent;
+use Symfony\Flex\Flex;
+
+class InstallRecipesCommandTest extends TestCase
+{
+    public function testCommandFlagsPassedDown()
+    {
+        $flex = $this->createMock(Flex::class);
+        $flex->method('update')->willReturnCallback(function (UpdateEvent $event) {
+            $this->assertTrue($event->reset());
+            $this->assertTrue($event->assumeYesForPrompts());
+        });
+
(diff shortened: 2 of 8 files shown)


maxhelias (approved):

src/Options.php
```diff
@@ -68,7 +68,7 @@ public function shouldWriteFile(string $file, bool $overwrite): bool
         exec('git status --short --ignored --untracked-files=all -- '.ProcessExecutor::escape($file).' 2>&1', $output, $status);
 
         if (0 !== $status) {
-            return $this->io && $this->io->askConfirmation(sprintf('Cannot determine the state of the "%s" file, overwrite anyway? [y/N] ', $file), false);
+            return $this->io && $this->io->askConfirmation(sprintf('Cannot determine the state of the "%s" file, overwrite anyway? [y/N] ', $file));
```
stof: you should change which of the `y` and `n` is a capital letter to reflect the default, so that users will know what will happen if they press `Enter` without writing any letter.

stof:
Note that this default should impact only files for which git cannot give us the modified state or untracked files. Existing files created by recipes are supposed to be tracked files. What is your workflow in which no files are being overwritten ?

shyim:
We have many customers that don't use Git (yes it's bad). Therefore they can't use the recipes:update command, so we tell them to use `recpies:install --reset --force -n` and diff them on their own (or create only new yaml files, don't touch existing ones).

As no file is tracked, flex asks for like 30 files this question, and for non interactive sessions, it did no changes as the default was false

nicolas-grekas:
Did you try using the `yes` command instead?

nicolas-grekas:
I'd rather handle this with a new command line argument instead, so that the default is safe. eg `-y` to swap the defaults.

Please rebase to target 2.x also (and revert the unrelated changes while doing so)

shyim:
I will add an flag `-y`, is fine for me :)

tests/Command/InstallRecipesCommandTest.php
```diff
@@ -0,0 +1,45 @@
...
+use Symfony\Component\Console\Tester\CommandTester;
+use Symfony\Flex\Command\DumpEnvCommand;
+use Symfony\Flex\Command\InstallRecipesCommand;
+use Symfony\Flex\Event\UpdateEvent;
+use Symfony\Flex\Flex;
+use Symfony\Flex\Options;
+
+class InstallRecipesCommandTest extends TestCase
+{
+    public function testCommandFlagsPassedDown(): void
```
nicolas-grekas:
```suggestion
    public function testCommandFlagsPassedDown()
```

tests/OptionsTest.php
```diff
@@ -0,0 +1,36 @@
...
+
+namespace Symfony\Flex\Tests;
+
+use Composer\IO\IOInterface;
+use PHPUnit\Framework\TestCase;
+use Symfony\Flex\Options;
+
+class OptionsTest extends TestCase
+{
+    public function testShouldWrite(): void
```
nicolas-grekas:
```suggestion
    public function testShouldWrite()
```

tests/OptionsTest.php
```diff
@@ -0,0 +1,36 @@
...
+
+        // We have an IO, and it allowed to write the file
+        $io = $this->createMock(IOInterface::class);
+        $io->expects($this->once())->method('askConfirmation')->willReturn(true);
+        $this->assertTrue((new Options([], $io))->shouldWriteFile(__FILE__, true, false));
+
+        // We skip all questions, so we're able to write
+        $this->assertTrue((new Options([], null))->shouldWriteFile(__FILE__, true, true));
+    }
+}
```
nicolas-grekas: missing EOL

nicolas-grekas (approved):
Some nitpicking and GTM

nicolas-grekas:
>1) Symfony\Flex\Tests\OptionsTest::testShouldWrite
>Failed asserting that true is false.

test case failing though

shyim:
is now finally green :)

nicolas-grekas:
Thank you @shyim.

Issue #938, opened 2022-07-14
Don't update package.json or controllers.json if vendor not updated


jmsche:
Hi,

Currently Flex seems to keep changing the values we set manually in package.json or controllers.json to the ones provided by the vendors we install (eg. https://github.com/symfony/ux/pull/394).

Maybe this package.json/controllers.json update behaviour should only happen if the given vendor has just been installed/updated, and not all the time?

bendavies:
yes, flex is basically rolling back my manual changes every time any composer install/update command completes.
this doesn't happen with any other receipe, which is confusing:

![image](https://user-images.githubusercontent.com/625392/187874812-b9140d7b-b427-477a-b2b9-e439aadec4bc.png)

oleg-andreyev:
this is quite annoying 

I have `"@hotwired/turbo": "^7.2.2"` and each `composer update` (e.g. resolving conflict after merge) flex is downgrading it to "@hotwired/turbo": "^7.1.0"

weaverryan:
Btw, a pragmatic, and perhaps easier-to-implement, solution would be this: only update the user's version in `package.json` if the version in the package is HIGHER than the one in `package.json`.

So, for example, suppose the user has `"tom-select": "^2.0.2"` and Flex wants to update this to `^2.0.1`. In this case, do NOT do the update. However, if Flex wants to update to `^2.1.0`, then YES do the update.

The `composer/semver` package (which is already available inside Flex) should have some tools to compare these... though I'm not sure the exact method/class to use.

stof:
Be careful. `composer/semver` parses Composer constraints. npm constraints are not exactly the same syntax (for instance, https://github.com/symfony/webpack-encore/releases/tag/v4.1.1 was caused by such a mistake, where the dev used a syntax supported in Composer but not in npm).

weaverryan:
Good point, but it's probably "close enough". And if we're wrong because the user has some fancier syntax, making the "wrong" decision is not a huge deal. In other words, we might do the wrong thing in some edge cases vs doing the wrong thing in ALL cases like right now :)

smnandre:
> Btw, a pragmatic, and perhaps easier-to-implement, solution would be this: only update the user's version in `package.json` if the version in the package is HIGHER than the one in `package.json`.
> 
[...]

Issue #843, opened 2021-12-03
PackageSynchronizer changes version of dependencies


weaverryan:
Hi there!

This is pretty easy to replicate:

```
symfony new flex_package_sync_test
cd flex_package_sync_test
composer require encore
composer require symfony/ux-turbo

# at this point, package.json will contain `@hotwired/turbo` at some version - e.g. `^7.0.0-beta.5`
# CHANGE this to something else - e.g. `^7.1.0`

composer update
# this will change the version back to `^7.0.0-beta.5`
```

Reported originally to me by @alexander-schranz

nicolas-grekas:
This doesn't happen on install anymore, but it does on update. The reason is that `PackageJsonSynchronizer` works without caring about symfony.lock nor existing constraints. Ignoring existing constraints looks on purpose to me, since that's how updates would work. Ignoring symfony.lock on the contrary doesn't look desired.

I don't know if it's possible to implement mutating package.json as a recipe configurator, but that could be the solution.

Anyone affected up for a PR?

PR #965, opened 2023-01-19
Keep existing constraints in package.json

smnandre:
When a dependency constraint is already set in the package.json, the PackageSynchroniser replaces it with even if the constraint is valid.

This PR changes this behaviour and let the constraint untouched if it matches the one provided by the PHP package.

Related issues:
* https://github.com/symfony/flex/issues/843
* https://github.com/symfony/flex/issues/938

### DIFF
--- a/src/PackageJsonSynchronizer.php
+++ b/src/PackageJsonSynchronizer.php
@@ -13,6 +13,7 @@ namespace Symfony\Flex;
 
 use Composer\Json\JsonFile;
 use Composer\Json\JsonManipulator;
+use Composer\Semver\VersionParser;
 use Seld\JsonLint\ParsingException;
 
 /**
@@ -23,11 +24,13 @@ class PackageJsonSynchronizer
 {
     private $rootDir;
     private $vendorDir;
+    private $versionParser;
 
     public function __construct(string $rootDir, string $vendorDir = 'vendor')
     {
         $this->rootDir = $rootDir;
         $this->vendorDir = $vendorDir;
+        $this->versionParser = new VersionParser();
     }
 
     public function shouldSynchronize(): bool
@@ -136,8 +139,10 @@ class PackageJsonSynchronizer
                 $content['devDependencies'][$dependency] = $constraint;
                 $didChangePackageJson = true;
             } elseif ($constraint !== $content[$parentNode][$dependency]) {
-                $content[$parentNode][$dependency] = $constraint;
-                $didChangePackageJson = true;
+                if ($this->shouldUpdateConstraint($content[$parentNode][$dependency], $constraint)) {
+                    $content[$parentNode][$dependency] = $constraint;
+                    $didChangePackageJson = true;
+                }
             }
         }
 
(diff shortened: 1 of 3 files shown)


src/PackageJsonSynchronizer.php
```diff
@@ -163,6 +166,19 @@ private function registerDependencies(array $flexDependencies): bool
         return $didChangePackageJson;
     }
 
+    private function shouldUpdateConstraint(string $existingConstraint, string $constraint)
+    {
+        try {
+            $versionParser = new VersionParser();
```
maxhelias: `VersionParser`, could be instantiated only once in the constructor.

smnandre: Fixed! thank you

maxhelias (dismissed):

tests/Fixtures/packageJson/stricter_constraints_package.json
```diff
@@ -0,0 +1,11 @@
+{
+   "name": "symfony/fixture",
+   "devDependencies": {
+      "@hotcookies": "^2",
+      "@hotdogs": "2.4.1",
```
weaverryan: The `@hotdogs` constraint in the package is `^2`. So let's use the following, so that we can test that the constraint WILL be updated if it doesn't match (so we'll use `@hotcookies` to test that satisfying constraints are NOT changed and we'll use `@hotdogs` to test that non-satisfying ARE changed):

```suggestion
      "@hotdogs": "^1.9",
```

tests/PackageJsonSynchronizerTest.php
```diff
@@ -245,4 +245,32 @@ public function testExistingElevatedPackage()
             json_decode(file_get_contents($this->tempDir.'/package.json'), true)
         );
     }
+
+    public function testStricterConstraintsAreKept()
```
weaverryan:
```suggestion
    public function testStricterConstraintsAreKeptNonMatchingAreReplaced()
```

tests/PackageJsonSynchronizerTest.php
```diff
@@ -245,4 +245,32 @@ public function testExistingElevatedPackage()
...
+
+        // Should keep existing constraints when stricter than packages ones
+        $this->assertSame(
+            [
+                'name' => 'symfony/fixture',
+                'devDependencies' => [
+                    '@hotcookies' => '^2',
+                    '@hotdogs' => '2.4.1',
+                    '@symfony/existing-package' => 'file:vendor/symfony/existing-package/Resources/assets',
+                ],
```
weaverryan, line 270: I changed to:

```
                'devDependencies' => [
                    // this satisfies the constraint, so it's kept
                    '@hotcookies' => '^2',
                    // this was too low, so it's replaced
                    '@hotdogs' => '^2',
                    '@symfony/existing-package' => 'file:vendor/symfony/existing-package/Resources/assets',
                ],
```

weaverryan (commented):
😍 

I've just made a small suggestion so that the test case will cover BOTH situations (when a constraint should not be updated but also when it SHOULD be updated). I've tested things locally, including these test changes. It's all working awesomely!

Thanks @smnandre!

tests/PackageJsonSynchronizerTest.php
smnandre: Thank you for your review and your suggestions :)

smnandre: I'm sorry i must have messed with git, i lost your changes so i put them in another commit

smnandre:
Thank you both for your review and your suggestions 😃

weaverryan (dismissed):
This is ready to go - excellent work - it seems to work great when I played with the test locally!

nicolas-grekas:
Thank you @smnandre.

PR #975, opened 2023-05-19
New "add-lines" configurator for simple file patching + importmap support

weaverryan:
Hi!

Today, WebpackEncoreBundle's recipe contains the UX/Stimulus files. Soon, I will introduce a new StimulusBundle - https://github.com/weaverryan/stimulus-bundle - so that the WebpackEncoreBundle can be used without Stimulus and (more importantly) the `stimulus_()` functions can be used with AssetMapper (i.e. without Encore).

This makes the recipe setup more... interesting :). This PR adds 2 things:

## importmap support in `JsonSynchronizer`

`JsonSynchronizer` now has 2 modes, based on the presence/absence of the `importmap.php` file. If that file is present, then:

* A) A new [symfony.importmap](https://github.com/weaverryan/stimulus-bundle/blob/148f6f9412e7063f9945d0947f206081d3311d7a/assets/package.json#L8-L11) config is read from the bundle's `package.json` file and these are added to the `importmap.php` file by running the `bin/console importmap:require` command. The `path:` prefix is used to refer to a "local" file in the bundle. Sometimes the importmap entries will be different than what's needed for `package.json`, hence having both configs.

* B) The `controllers.json` file is updated like normal

Also, a new [symfony.needsPackageAsADependency](https://github.com/weaverryan/stimulus-bundle/blob/148f6f9412e7063f9945d0947f206081d3311d7a/assets/package.json#LL7C10-L7C35) config key was added specifically for StimulusBundle. If `true`, no `file:/vendor/...` package will be added to `package.json` when using Encore.

## add-lines Configurator

The new `add-lines` configurator is able to add entire lines to the `top`, `bottom` of `after_target` of existing files (if they exist). Example usage:

```json
"add-lines": [
    {
        "file": "webpack.config.js",
        "content": "\n    // enables the Symfony UX Stimulus bridge (used in assets/bootstrap.js)\n    .enableStimulusBridge('./assets/controllers.json')",
        "position": "after_target",
        "target": ".splitEntryChunks()"
    },
    {
        "file": "assets/app.js",
        "content": "import './bootstrap.js';",
        "position": "top",
        "warn_if_missing": true
    }
]
```

There is also a `requires` key to only run if another package is installed. This is needed because StimulusBundle will need a [different assets/bootstrap.js](https://github.com/weaverryan/recipes/blob/3ab0e996ae22af665ec83bf0634f1163b533bc36/symfony/stimulus-bundle/1.0/manifest.json#L23-L33) based on if Encore vs AssetMapper is installed

Cheers!

### DIFF
--- a/src/Configurator/AddLinesConfigurator.php
+++ b/src/Configurator/AddLinesConfigurator.php
@@ -1,0 +1,230 @@
+<?php
+
+namespace Symfony\Flex\Configurator;
+
+use Composer\IO\IOInterface;
+use Symfony\Flex\Lock;
+use Symfony\Flex\Recipe;
+use Symfony\Flex\Update\RecipeUpdate;
+
+/**
+ * @author Kevin Bond <kevinbond@gmail.com>
+ * @author Ryan Weaver <ryan@symfonycasts.com>
+ */
+class AddLinesConfigurator extends AbstractConfigurator
+{
+    private const POSITION_TOP = 'top';
+    private const POSITION_BOTTOM = 'bottom';
+    private const POSITION_AFTER_TARGET = 'after_target';
+
+    private const VALID_POSITIONS = [
+        self::POSITION_TOP,
+        self::POSITION_BOTTOM,
+        self::POSITION_AFTER_TARGET,
+    ];
+
+    public function configure(Recipe $recipe, $config, Lock $lock, array $options = []): void
+    {
+        foreach ($config as $patch) {
+            if (!isset($patch['file'])) {
+                $this->write(sprintf('The "file" key is required for the "add-lines" configurator for recipe "%s". Skipping', $recipe->getName()));
+
+                continue;
+            }
+
+            if (isset($patch['requires']) && !$this->isPackageInstalled($patch['requires'])) {
+                continue;
+            }
+
+            if (!isset($patch['content'])) {
+                $this->write(sprintf('The "content" key is required for the "add-lines" configurator for recipe "%s". Skipping', $recipe->getName()));
+
+                continue;
+            }
+            $content = $patch['content'];
+
+            $file = $this->path->concatenate([$this->options->get('root-dir'), $patch['file']]);
+            $warnIfMissing = isset($patch['warn_if_missing']) && $patch['warn_if_missing'];
+            if (!is_file($file)) {
+                $this->write([
+                    sprintf('Could not add lines to file <info>%s</info> as it does not exist. Missing lines:', $patch['file']),
+                    '<comment>"""</comment>',
+                    $content,
+                    '<comment>"""</comment>',
+                    '',
+                ], $warnIfMissing ? IOInterface::NORMAL : IOInterface::VERBOSE);
+
+                continue;
+            }
+
+            $this->write(sprintf('Patching file "%s"', $patch['file']));
+
+            if (!isset($patch['position'])) {
+                $this->write(sprintf('The "position" key is required for the "add-lines" configurator for recipe "%s". Skipping', $recipe->getName()));
+
+                continue;
+            }
+            $position = $patch['position'];
+            if (!\in_array($position, self::VALID_POSITIONS, true)) {
+                $this->write(sprintf('The "position" key must be one of "%s" for the "add-lines" configurator for recipe "%s". Skipping', implode('", "', self::VALID_POSITIONS), $recipe->getName()));
+
+                continue;
+            }
+
+            if (self::POSITION_AFTER_TARGET === $position && !isset($patch['target'])) {
+                $this->write(sprintf('The "target" key is required when "position" is "%s" for the "add-lines" configurator for recipe "%s". Skipping', self::POSITION_AFTER_TARGET, $recipe->getName()));
+
+                continue;
+            }
+            $target = isset($patch['target']) ? $patch['target'] : null;
+
+            $this->patchFile($file, $content, $position, $target, $warnIfMissing);
+        }
+    }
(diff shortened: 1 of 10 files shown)


weaverryan:
This is ready for review 🖖

javiereguiluz (approved):
This looks OK to me. Thanks Ryan (and Kevin)!

src/PackageJsonSynchronizer.php
```diff
@@ -24,22 +24,30 @@ class PackageJsonSynchronizer
...
-        return $this->rootDir && file_exists($this->rootDir.'/package.json');
+        return $this->rootDir && (file_exists($this->rootDir.'/package.json') || file_exists($this->rootDir.'/importmap.php'));
     }
 
     public function synchronize(array $phpPackages): bool
     {
+        if (file_exists($this->rootDir.'/importmap.php')) {
+            $this->synchronizeForAssetMapper($phpPackages);
+
+            return false;
```
nicolas-grekas, line 48: should we configure also package.json in case both exist?

src/Configurator/AddLinesConfigurator.php
```diff
@@ -0,0 +1,196 @@
...
+                        array_splice($lines, $key + 1, 0, $value);
+                        $targetFound = true;
+
+                        break;
+                    }
+                }
+                $fileContents = implode("\n", $lines);
+
+                if (!$targetFound) {
+                    // TODO warn?
```
nicolas-grekas: display a message? Could not add lines after "%s" as no such string was found in "%s": $the-lines

nicolas-grekas (commented):
LGTM. Do we need a rule to replace a line by another?

src/PackageJsonSynchronizer.php
weaverryan: It's a good question - e.g. "if `package.json` exists, why not update it?". But one use-case will be that someone is using AssetMapper but has a small package.json file so that they can process Tailwind files or React files. In that case, we wouldn't want to update their `package.json`. So my thinking currently is: there's a boolean choice between Encore vs AssetMapper / `importmap.php` vs `package.json`.

src/Configurator/AddLinesConfigurator.php
weaverryan: Good catch! It looks like this now both here and if the file is missing (with wording differences, of course):

Screenshot 2023-05-22 at 6 37 10 AM

The user would only see this if `warn_if_missing: true` in `manifest.json` (or, of course, if they have verbosity high).

weaverryan:
> LGTM. Do we need a rule to replace a line by another?

Thank you! I didn't NEED the replace yet, so I skipped it - it could be a `position: replace_target`. I can add it now if you'd like. Otherwise, wait until we need it.

src/Configurator/AddLinesConfigurator.php
```diff
@@ -0,0 +1,206 @@
...
+    ];
+
+    public function configure(Recipe $recipe, $config, Lock $lock, array $options = []): void
+    {
+        foreach ($config as $patch) {
+            if (!isset($patch['file'])) {
+                throw new \InvalidArgumentException(sprintf('The "file" key is required for the "add-lines" configurator for recipe "%s".', $recipe->getName()));
+            }
+
+            if (isset($patch['requires']) && !$this->isPackageInstalled($patch['requires'])) {
```
stof, line 35: should this be a string only or should we support a list of package names ?

weaverryan: My gut is: let's keep it simple as a string for now. We could expand it later to an array if needed - I can't think of a use-case yet. 

But I CAN add it now if we want it - I've been trying to focus the scope.

stof: turning it into an array later would require using a different key, to avoid making recipes incompatible with older versions of Flex that would break badly when receiving something else than a string. So if we don't support an array already, we should at least properly handled unsupported values.

weaverryan: Good point - I've added support for an array, as it turned out to be just as easy as checking for the type and throwing an exception.

stof (approved):

src/Configurator/AddLinesConfigurator.php
```diff
@@ -0,0 +1,216 @@
...
+            // Ignore "requires": the target packages may have just become uninstalled.
+            // Checking for a "content" match is enough.
+
+            $file = $this->path->concatenate([$this->options->get('root-dir'), $patch['file']]);
+            if (!is_file($file)) {
+                continue;
+            }
+
+            if (!isset($patch['content'])) {
+                throw new \InvalidArgumentException(sprintf('The "value" key is required for the "add-lines" configurator for recipe "%s".', $recipe->getName()));
```
stof: should we throw or should we ignore the invalid recipe ? throwing would break the whole composer command.

stof: All those validation rules should be implemented as part of the recipe validator for the CI of the recipes repo.

weaverryan: Good catch! I've made a PR to add the linting rules - https://github.com/symfony-tools/recipes-checker/pull/14

I've also relaxed these to low-priority warnings (and skipping the rule). It will also help if we add more capabilities in the future to recipes, but they are perhaps on an older flex version (it won't work correctly, but no explosion).

kbond (approved):

fabpot:
Thank you @weaverryan.

PR #976, opened 2023-05-26
Avoid errors with an older global install of Flex

weaverryan:
Hi!

When I tried 2.3.0 locally, I it exploded, complaining about:

> Call to undefined method Symfony\Flex\Configurator::postInstall()
The reason is that I have a global `symfony/flex` install, which was NOT on 2.3.0. And so... I guess (?) your global `symfony/flex` is used when originally creating a new project / installing `symfony/flex` into a project... at least partially? It's a bit opaque to me, but this may be needed to avoid issues.

Btw, the fix for this is running `composer global up symfony/flex`.

### DIFF
--- a/src/Flex.php
+++ b/src/Flex.php
@@ -544,10 +544,12 @@ class Flex implements PluginInterface, EventSubscriberInterface
             }
         }
 
-        foreach ($postInstallRecipes as $recipe) {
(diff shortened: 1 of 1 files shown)


fabpot (approved):

fabpot:
Thank you @weaverryan.

PR #978, opened 2023-05-29
[Bug] Fix Flex not installing recipes while it is being upgraded

### LINKED ISSUE #977: Flex does not install recipes if Flex itself is being upgraded
Hi!

Pretty sure we stumbled on a bug where if (A) `symfony/flex` is being upgraded AND (B) a new package is being installed that has a recipe, the recipe will be ignored.

We're seeing this when upgrading apps using certain UX packages, as a `composer update` grabs the new `symfony/flex` version AND installs `symfony/stimulus-bundle`, which is a new dependency for some UX packages.

Reproducer:

```
git clone git@github.com:weaverryan/flex-recipe-upgrade-bug.git
cd flex-recipe-upgrade-bug
composer update
```

You'll notice that it installs `symfony/stimulus-bundle`, but skips installing its recipe. Running `composer update` a 2nd time causes it to be installed.

## Cause

This appears to be due to Composer being smart. When a plugin is upgraded, it's PluginManager::update() method unloads the old plugin, and loads the new one. The install of StimulusBundle is recorded into a `Flex` instance... but then that `Flex` instance is discarded and a new instance is created that is *not* aware of what packages were just installed.

## Solution

Possibly, in `Flex::deactivate()`, we could... store `$this->operations` onto a static `Flex` property - e.g. `public static $storedOperations`. Then in `Flex::activate()`, we could look for `Symfony\Flex\Flex::$storedOperations` and use that (after the update, Composer uses a class called `Symfony\Flex\Flex_composer_tmp0`... so I'm assuming that `Symfony\Flex\Flex` would still be available, and we could look directly for it.

Cheers!

weaverryan:
Fixes #977

Here's the lifecycle:

A) Packages are installed/updated and `Flex::recordOperations()` is called.
B) Composer sees that `symfony/flex` is updated and unloads the plugin and reloads it under a new `Symfony\Flex\Flex_composer_tmp0` class. This results in the `Flex::$operations` property being lost.
C) And so, no recipes are installed.

This PR "stashes" the operations directly on `Symfony/Flex/Flex` where we then read them later. Tested locally. Amazing, it seems to work perfectly.

Also, backports #976 to 1.x - my fault for making the PR on the wrong branch.

Cheers!

### DIFF
--- a/src/Flex.php
+++ b/src/Flex.php
@@ -60,6 +60,8 @@ use Symfony\Thanks\Thanks;
  */
 class Flex implements PluginInterface, EventSubscriberInterface
 {
+    public static $storedOperations = [];
+
     /**
      * @var Composer
      */
@@ -125,6 +127,13 @@ class Flex implements PluginInterface, EventSubscriberInterface
         $this->io = $io;
         $this->config = $composer->getConfig();
         $this->options = $this->initOptions();
+
+        // if Flex is being upgraded, the original operations from the original Flex
+        // instance are stored in the static property, so we can reuse them now.
+        if (property_exists(self::class, 'storedOperations') && self::$storedOperations) {
+            $this->operations = self::$storedOperations;
+            self::$storedOperations = [];
+        }
 
         $symfonyRequire = preg_replace('/\.x$/', '.x-dev', getenv('SYMFONY_REQUIRE') ?: ($composer->getPackage()->getExtra()['symfony']['require'] ?? ''));
 
@@ -293,6 +302,8 @@ class Flex implements PluginInterface, EventSubscriberInterface
 
     public function deactivate(Composer $composer, IOInterface $io)
     {
+        // store operations in case Flex is being upgraded
+        self::$storedOperations = $this->operations;
         self::$activated = false;
     }
 
@@ -544,10 +555,12 @@ class Flex implements PluginInterface, EventSubscriberInterface
             }
         }
 
-        foreach ($postInstallRecipes as $recipe) {
-            $this->configurator->postInstall($recipe, $this->lock, [
-                'force' => $event instanceof UpdateEvent && $event->force(),
-            ]);
+        if (method_exists($this->configurator, 'postInstall')) {
+            foreach ($postInstallRecipes as $recipe) {
+                $this->configurator->postInstall($recipe, $this->lock, [
+                    'force' => $event instanceof UpdateEvent && $event->force(),
+                ]);
+            }
         }
 
         if (null !== $manifest) {


src/Flex.php
```diff
@@ -126,6 +127,15 @@ class_exists(__NAMESPACE__.str_replace('/', '\\', substr($file, \strlen(__DIR__)
         $this->config = $composer->getConfig();
         $this->options = $this->initOptions();
 
+        // if Flex is being upgraded, the original operations from the original Flex
+        // instance are stored in the static property, so we can reuse them now.
+        if (property_exists(self::class, 'storedOperations')) {
+            if (null !== self::$storedOperations) {
```
fabpot: Could be combined with the previous condition?

fabpot:
```suggestion
            if (self::$storedOperations) {
```

fabpot (changes requested):

fabpot:
Is it only for 2.x or also for 1.x?

src/Flex.php
```diff
@@ -60,6 +60,7 @@
  */
 class Flex implements PluginInterface, EventSubscriberInterface
 {
+    public static $storedOperations = null;
```
nicolas-grekas:
```suggestion
    public static $storedOperations = [];

```
(including the blank line after)

src/Flex.php
```diff
@@ -126,6 +127,15 @@ class_exists(__NAMESPACE__.str_replace('/', '\\', substr($file, \strlen(__DIR__)
         $this->config = $composer->getConfig();
         $this->options = $this->initOptions();
 
+        // if Flex is being upgraded, the original operations from the original Flex
+        // instance are stored in the static property, so we can reuse them now.
+        if (property_exists(self::class, 'storedOperations')) {
```
nicolas-grekas: to be build on what @fabpot suggested:
```suggestion
        if (property_exists(self::class, 'storedOperations') && self::$storedOperations) {
```

src/Flex.php
```diff
@@ -544,10 +556,12 @@ function ($value) {
             }
         }
 
-        foreach ($postInstallRecipes as $recipe) {
-            $this->configurator->postInstall($recipe, $this->lock, [
-                'force' => $event instanceof UpdateEvent && $event->force(),
-            ]);
+        if (method_exists($this->configurator, 'postInstall')) {
```
nicolas-grekas, line 558: Does this means that post-install configurators are not run when upgrading?
If yes, what about moving the postInstall method to a new class, so that we can load it even when the old configurator class is already loaded?

stof: We would have such issue any time we _edit_ a class of the plugin (except the main plugin class): when Flex is updated, it is possible that the loaded class is the old one rather than the new one.

weaverryan:
> Is it only for 2.x or also for 1.x?

Also in 1.x. My mistake - I updated the branch target - thanks for catching that.

src/Flex.php
weaverryan: Exactly. This part is a backport of #976, which I incorrectly put into 2.x. In that case, I had a globally-installed flex that was on 2.2.0. When I created a new project, I believe the flow that happened was:

A) Composer uses the globally installed Flex at first, and autoloads the `Configurator` class. So we have a 2.2.0 `Configurator` in memory.
B) Composer then unloads the 2.2.0 `Flex` class and loads a new 2.3.0 `Flex` class (aliased as `Flex_composer_tmp0` internally)
C) The new `Flex_composer_tmp0` tries to call `$this->configurator->postInstall`, but it's calling that on the 2.2.0 `Configurator` class.

So, as Stof said, when a plugin is updated, Composer does a cool job of unloading the old version and reloading the new version. But it only applies to the actual plugin `Flex` class itself. All other classes may still be using the old version for the rest of the operation. You probably already know this - but it's new to my understanding :).

weaverryan:
This should be ready to go

fabpot (approved):

fabpot:
Thank you @weaverryan.

PR #987, opened 2023-08-01
Fixing problem where rootDir is ".", which strips all dots from the filename

weaverryan:
Fixes https://github.com/symfony/recipes/issues/1120

In a normal siutation, `$rootDir` is `.`. The intention of this code is to strip the "rootDir" from the START of the `$file` path, so that we are left with only the relative path. The `.` + `str_replace()` was too greedy, and was causing things like `docker-composeryaml`.

The test uses an absolute root dir, and replicating the `.` would be tricky. But I tested this locally and the fix works.

Cheers!


fabpot (approved):

fabpot:
Thank you @weaverryan.

PR #988, opened 2023-08-02
[AddLines] Fix update process to send changes to RecipeUpdate vs make them immediately

weaverryan:
Hi!

This was an oversight when I create `AddLinesConfigurator`.

During `recipes:update`, the purpose of the `update()` method in each configurator is to record what the "original" contents of a file would look like from the original recipe & what the "new" contents of a file would look like with the new recipe. Then, LATER, the patcher system creates a diff from these.

The `AddLinesConfigurator` was going rogue and updating the files immediately, which caused the "patcher" system later to explode because this file was unexpectedly already modified.

The changes are less drastic than they look, as code needed to be moved from some public methods -> private methods.

Cheers!

### DIFF
--- a/src/Configurator/AddLinesConfigurator.php
+++ b/src/Configurator/AddLinesConfigurator.php
@@ -35,30 +35,11 @@ class AddLinesConfigurator extends AbstractConfigurator
 
     public function unconfigure(Recipe $recipe, $config, Lock $lock): void
     {
-        foreach ($config as $patch) {
-            if (!isset($patch['file'])) {
-                $this->write(sprintf('The "file" key is required for the "add-lines" configurator for recipe "%s". Skipping', $recipe->getName()));
+        $changes = $this->getUnconfigureFileChanges($recipe, $config);
 
-                continue;
-            }
-
-            // Ignore "requires": the target packages may have just become uninstalled.
-            // Checking for a "content" match is enough.
-
-            $file = $this->path->concatenate([$this->options->get('root-dir'), $patch['file']]);
-            if (!is_file($file)) {
-                continue;
-            }
-
-            if (!isset($patch['content'])) {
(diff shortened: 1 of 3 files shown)


src/Configurator/AddLinesConfigurator.php
```diff
@@ -23,7 +23,61 @@ class AddLinesConfigurator extends AbstractConfigurator
...
+    {
+        // manually check for "requires", as unconfigure ignores it
+        $originalConfig = array_filter($originalConfig, function ($item) {
+            return !isset($item['requires']) || $this->isPackageInstalled($item['requires']);
+        });
+
+        // reset the file content cache
+        $this->fileContents = [];
+        $this->executeUnconfigure($recipeUpdate->getOriginalRecipe(), $originalConfig);
+        $this->executeConfigure($recipeUpdate->getNewRecipe(), $newConfig);
```
weaverryan, line 67: The old logic "unconfigured" the original recipe, then configured the new recipe. So we're doing the same, except that we "record" the changes and report them to `$recipeUpdate` below instead of actually *making* the changes now.

src/PackageJsonSynchronizer.php
```diff
@@ -296,7 +296,12 @@ private function updateControllersJsonFile(array $phpPackages)
             return;
         }
 
-        $previousControllersJson = (new JsonFile($controllersJsonPath))->read();
+        try {
+            $previousControllersJson = (new JsonFile($controllersJsonPath))->read();
+        } catch (ParsingException $e) {
+            // if controllers.json is invalid (possible during a recipe upgrade), we can't update the file
+            return;
+        }
```
weaverryan, line 304: Unrelated to the PR - but noticed while doing it. We already have this same code in this file to avoid exploding if `package.json` can't be parsed due to a conflict.

fabpot (approved):

fabpot:
Thank you @weaverryan.

PR #990, opened 2023-08-10
[Feature] Add options expansion to Add Lines Configuration

mnocon:
Hi!

I've got a chance to play with the new AddLines configurator and it's an extremely useful feature, thanks for introducing it!

There is one small thing that seems missing to me: support for placeholders that are available in the `copy-from-package` and `copy-from-recipe` configurators.

It's this feature:
```
These are the available placeholders: %BIN_DIR%, %CONF_DIR%, %CONFIG_DIR%, %SRC_DIR% %VAR_DIR% and %PUBLIC_DIR%.

Recipes must use these placeholders instead of hardcoding the paths to be truly reusable. The placeholder values can be overridden in the extra section of your composer.json file (where you can define your own placeholders too):
```
(this text comes from https://github.com/symfony/recipes)

This small PR makes it possible to use placeholders in the add-lines configurator.

I'm not sure about the target branch, please let me know if I should rebase to 1.x - also not sure if I should mention it in the doc somewhere.

### DIFF
--- a/tests/Configurator/AddLinesConfiguratorTest.php
+++ b/tests/Configurator/AddLinesConfiguratorTest.php
@@ -67,6 +67,26 @@ EOF
             $actualContents);
     }
 
+    public function testExpandTargetDirWhenConfiguring()
+    {
+        $this->saveFile('config/file.txt', 'FirstLine');
+
+        $this->runConfigure([
+            [
+                'file' => '%CONFIG_DIR%/file.txt',
+                'position' => 'top',
+                'content' => 'NewFirstLine',
+            ],
+        ]);
+        $actualContents = $this->readFile('config/file.txt');
+        $this->assertSame(<<<EOF
+NewFirstLine
+FirstLine
+EOF
+            ,
+            $actualContents);
+    }
+
     public function testLinesAddedToBottomOfFile()
     {
         $this->saveFile('assets/app.js', <<<EOF
@@ -316,6 +336,28 @@ EOF
         ]);
         $actualContents = $this->readFile('assets/app.js');
         $this->assertSame($expectedContents, $actualContents);
+    }
+
+    public function testExpandTargetDirWhenUnconfiguring()
+    {
+        $this->saveFile('config/file.txt',
+            <<<EOF
+Line1
+Line2
+EOF
+        );
+
+        $this->runUnconfigure([
+            [
+                'file' => '%CONFIG_DIR%/file.txt',
+                'content' => 'Line1',
+            ],
+        ]);
+        $actualContents = $this->readFile('config/file.txt');
+        $this->assertSame(<<<EOF
+Line2
+EOF
+            , $actualContents);
     }
 
     public function getUnconfigureTests()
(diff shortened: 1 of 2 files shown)


tests/Configurator/AddLinesConfiguratorTest.php
```diff
@@ -67,6 +67,26 @@ public function testLinesAddedToTopOfFile()
             $actualContents);
     }
 
+    public function testConvertsOptionsWhenConfiguring()
```
alamirault: As we assert only `expendTargetDir`, what to you think to rename tests ?

```suggestion
    public function testExpandTargetDirWhenConfiguring()
```

tests/Configurator/AddLinesConfiguratorTest.php
```diff
@@ -318,6 +338,27 @@ public function testUnconfigure(string $originalContents, string $value, string
         $this->assertSame($expectedContents, $actualContents);
     }
 
+    public function testConvertsOptionsWhenUnconfiguring()
```
alamirault:
```suggestion
    public function testExpandTargetDirWhenUnconfiguring()
```

tests/Configurator/AddLinesConfiguratorTest.php
mnocon: Sure, thanks for the suggestion!

alamirault (approved):

maxhelias (approved):

nicolas-grekas:
Thanks for the PR. Can you please rebase it for branch 1.x? Please also revert the CS changes while doing so as 1.x still supports PHP 7.1

mnocon:
Thanks for the suggestion!
I've rebased this PR onto 1.x and adapted the code to PHP 7.1 (small change: https://github.com/symfony/flex/pull/990/commits/88672dcc54bfd1477beb41d60eb22139623a5d49)

Questions:
1) Bot still suggests some CS improvements - but from what I see polyfills are not installed when using `--prefer-lowest`, is it ok to ignore this?
2) The `--prefer-lowest` job is failing with:
```
[RuntimeException]                                                                                                                                                           
  Could not load package microweber/microweber in [http://packagist.org:](http://packagist.org/) [UnexpectedValueException] Could not parse version constraint v2.6.*: Invalid version string "v2.6.*"
```
The earliest occurance I found of this failure is here: https://github.com/symfony/flex/actions/runs/7101907556/job/19331086614

I've tried debugging this, but TBH I don't even know where to start (and when running the tests locally I had a different error) - do you have any suggestions where to look?

tests/Configurator/AddLinesConfiguratorTest.php
```diff
@@ -318,6 +338,28 @@ public function testUnconfigure(string $originalContents, string $value, string
...
     }
 
+    public function testExpandTargetDirWhenUnconfiguring()
+    {
+        $this->saveFile('config/file.txt',
+            <<<EOF
+Line1
+Line2
+EOF
+);
```
nicolas-grekas:
```suggestion
        );
```

nicolas-grekas (approved):
Nervemind the failures, they're false-positives.

fabpot:
Thank you @mnocon.

PR #994, opened 2023-10-06
feat: support for compose.yml

### LINKED ISSUE #992: Suggest adding management of the compose.yml file in addition to docker-compose.yml
Hello !

Following my use of Symfony Flex and its wonderful ability to manage Docker configurations via recipes, I noticed that the manifest supports the docker-compose.yml file.

However, consulting the official Docker documentation, it seems that the use of the docker-compose.yml file may be gradually deprecated or, at least, that there is an alternative under the name compose.yml. Details can be found on this page: https://docs.docker.com/compose/multiple-compose-files/merge/

Proposal:

- Add support for compose.yml to the Symfony Flex manifest.
- Check whether a docker-compose.yml already exists and use it, otherwise create a compose.yml.

This will ensure better compatibility with future versions of Docker Compose and offer greater flexibility to developers using Symfony Flex.

I'd love to hear your thoughts on this and whether this is something you'd consider adding to Symfony Flex.

Thank you for your time and consideration!

dunglas:
Docker Compose now recommends using a definition file named `compose.yaml` and also supports `compose.yml` (the legacy `docker-compose.ya?ml` are still supported): https://docs.docker.com/compose/compose-file/03-compose-file/

This PR adds support for `compose.ya?ml` file to Flex, and changes the name of the generated file to `compose.yaml`, according to Docker best practices.

Closes #992.

### DIFF
--- a/src/Configurator/DockerComposeConfigurator.php
+++ b/src/Configurator/DockerComposeConfigurator.php
@@ -23,9 +23,9 @@ use Symfony\Flex\Recipe;
 use Symfony\Flex\Update\RecipeUpdate;
 
 /**
- * Adds services and volumes to docker-compose.yml file.
+ * Adds services and volumes to compose.yaml file.
  *
- * @author Kévin Dunglas <dunglas@gmail.com>
+ * @author Kévin Dunglas <kevin@dunglas.dev>
  */
 class DockerComposeConfigurator extends AbstractConfigurator
 {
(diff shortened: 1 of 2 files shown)


nicolas-grekas (commented):
It might be too early to generate composer.yaml by default. New docker versions still support docker-compose.yaml, isn't it? I think we should plan for a smoother transition period.

nicolas-grekas:
Alternatively, can we detect the version of docker that is in use?

dunglas:
> It might be too early to generate composer.yaml by default.

Docker Compose support `compose.yaml` since 1.28.6 (2021-03-23). Docker Compose v1 stopped receiving updates in July 2023.
All currently maintained Docker Compose installations support this file name.

Also, it's weird to generate files on new projects (existing projects will continue working and are unaffected) that are marked as legacy by the upstream vendor.

> Alternatively, can we detect the version of docker that is in use?

This should be possible, but is it worth it? Starting a new project with an EOLed version of Docker Compose looks very weird.

fabpot (approved):

fabpot:
Thank you @dunglas.

Issue #899, opened 2022-04-23
Add COMPOSER_MEMORY_LIMIT support to Symfony\Flex\ScriptExecutor


PabloKowalczyk:
Hello, this change will allow to define same memory limit for Composer and sub-processes run by `ScriptExecutor`. Any interest in this?
I can provide PR if you are interested.

nicolas-grekas:
Please send a PR if *you* are interested (that's how OSS work ;) )

PR #91, opened 2017-05-23
Check for COMPOSER_ORIGINAL_INIS env otherwise use regular php ini

xawiers:
[ScriptExecutor] added check and use env COMPOSER_ORIGINAL_INIS variable.

After fresh install and `composer req cli` I got errors:

```
Executed script assets:install --symlink --relative web/  [KO]
 [KO]
Script assets:install --symlink --relative web/ returned with error code 1
!!  Cannot load Zend OPcache - it was already loaded
!!  
!!  PHP Warning:  Module 'PDO' already loaded in Unknown on line 0
!!  PHP Warning:  Module 'xml' already loaded in Unknown on line 0
!!  PHP Warning:  Module 'bcmath' already loaded in Unknown on line 0
!!  PHP Warning:  Module 'bz2' already loaded in Unknown on line 0
!!  PHP Warning:  Module 'calendar' already loaded in Unknown on line 0
!!  PHP Warning:  Module 'ctype' already loaded in Unknown on line 0
!!  PHP Warning:  Module 'curl' already loaded in Unknown on line 0
!!  PHP Warning:  Module 'dom' already loaded in Unknown on line 0
!!  PHP Warning:  Module 'exif' already loaded in Unknown on line 0
!!  PHP Warning:  Module 'fileinfo' already loaded in Unknown on line 0
!!  PHP Warning:  Module 'ftp' already loaded in Unknown on line 0
!!  PHP Warning:  Module 'gd' already loaded in Unknown on line 0
!!  
!!  PHP Warning:  Module 'gettext' already loaded in Unknown on line 0
!!  PHP Warning:  Module 'iconv' already loaded in Unknown on line 0
!!  PHP Warning:  Module 'intl' already loaded in Unknown on line 0
!!  PHP Warning:  Module 'json' already loaded in Unknown on line 0
!!  PHP Warning:  Module 'mbstring' already loaded in Unknown on line 0
!!  PHP Warning:  Module 'mcrypt' already loaded in Unknown on line 0
!!  PHP Warning:  Module 'Phar' already loaded in Unknown on line 0
!!  PHP Warning:  Module 'posix' already loaded in Unknown on line 0
!!  PHP Warning:  Module 'readline' already loaded in Unknown on line 0
!!  
!!  PHP Warning:  Module 'shmop' already loaded in Unknown on line 0
!!  PHP Warning:  Module 'SimpleXML' already loaded in Unknown on line 0
!!  PHP Warning:  Module 'sockets' already loaded in Unknown on line 0
[...]

PR #995, opened 2023-10-20
fix: Correctly inherit the Composer restarted process settings

theofidry:
> [!WARNING]
> Note: this PR is more of draft. I am not confident I understand fully
the issue neither that this is the right fix (and I could not test it). Do NOT assume this code just works.

After a lot of digging in https://github.com/box-project/box/issues/988, @maartendekeizer could identify the root of the issue.

My understanding is that Flex tries to detect if the current PHP process was restarted by Composer and forwards its restarted settings to the sub-processes it is going to launch. There is currently two things done:

- #91 which if took code from SensioDistributionBundle.
- #899 which kind of followed the suite.

    I suspect the mentioned code predates `composer/xdebug-handler`. Now
with this package, there is two things to take into account:

- The `composer/xdebug-handler` API is likely much safer, more robust and a lot less hacky to use.
- There is other applications that can restart a PHP process. As an example Box restarts the PHP process to be able to correct the `phar.readonly` setting that cannot be changed at runtime. It matters as the restarted process by be executing a Composer command.

As mentioned in the warning, I could not test this, not even run the tests locally, so I would be careful about this PR. I just wanted to give a base about a potential fix with context about the original issue encountered.

On my side to not have to wait on Flex and avoid the users to have to update it to have the fix, Box launches its Composer commands with `COMPOSER_ORIGINAL_INIS=''` to avoid Flex to trigger the problematic piece of code.

### DIFF
--- a/src/ScriptExecutor.php
+++ b/src/ScriptExecutor.php
@@ -16,6 +16,7 @@ use Composer\EventDispatcher\ScriptExecutionException;
 use Composer\IO\IOInterface;
 use Composer\Semver\Constraint\MatchAllConstraint;
 use Composer\Util\ProcessExecutor;
+use Composer\XdebugHandler\PhpConfig;
 use Symfony\Component\Console\Output\OutputInterface;
 use Symfony\Component\Console\Output\StreamOutput;
 use Symfony\Component\Process\PhpExecutableFinder;
@@ -116,20 +117,8 @@ class ScriptExecutor
 
         $arguments = $phpFinder->findArguments();
 
-        if ($env = (string) getenv('COMPOSER_ORIGINAL_INIS')) {
-            $paths = explode(\PATH_SEPARATOR, $env);
-            $ini = array_shift($paths);
-        } else {
-            $ini = php_ini_loaded_file();
-        }
-
-        if ($ini) {
-            $arguments[] = '--php-ini='.$ini;
-        }
-
-        if ($memoryLimit = (string) getenv('COMPOSER_MEMORY_LIMIT')) {
-            $arguments[] = "-d memory_limit={$memoryLimit}";
-        }
+        $phpConfig = new PhpConfig();
+        $phpConfig->usePersistent();
 
         $phpArgs = implode(' ', array_map([ProcessExecutor::class, 'escape'], $arguments));
         $scriptArgs = implode(' ', array_map([ProcessExecutor::class, 'escape'], $scriptArguments));


src/ScriptExecutor.php
```diff
@@ -116,20 +117,8 @@ private function expandPhpScript(string $cmd, array $scriptArguments): string
...
-            $ini = array_shift($paths);
-        } else {
-            $ini = php_ini_loaded_file();
-        }
-
-        if ($ini) {
-            $arguments[] = '--php-ini='.$ini;
-        }
-
-        if ($memoryLimit = (string) getenv('COMPOSER_MEMORY_LIMIT')) {
```
maartendekeizer, line 130: I think we still needs this if block, the env value is added as argument to the processes which is not done via PhpConfig

theofidry: but the memory limit has been configured and dumped into the tmp ini file, so with persistent settings those changes done to the restarted PHP process should be preserved (I think)

maartendekeizer:
A side note about the original code.

The `COMPOSER_ORIGINAL_INIS` will contain values like:
`/etc/php/8.2/cli/php.ini:/etc/php/8.2/cli/conf.d/10-mysqlnd.ini:/etc/php/8.2/cli/conf.d/10-opcache.ini`

By exploding this string and take the first item (in this example `/etc/php/8.2/cli/php.ini`) and then find the directory of the file (`/etc/php/8.2/cli`) and using this directory as ini scan dir we still not finding other locations. If the env is for example `/etc/php/8.2/cli/php.ini:/etc/php/8.2/cli/conf.d/10-mysqlnd.ini:/etc/php/8.2/cli/conf.d/10-opcache.ini:/home/website/config/99-usersettings.ini` the `/home/website/config` directory is never scanned for the symfony-cmd command.

src/ScriptExecutor.php
maartendekeizer: Nope, XdebugHandler dumped the ini value not the runtime value.

It is not possible to set the PHP memory limit via envs (beside when configured in php.ini). So custom code like this is required.

This code makes it also possible to have another memory limit for a subprocess like Composer then the main processes.
But you can debate if it is correct that this env will be passed to a subprocesses. If you give Composer a limit of 1Gb, you will give the main processes and each subprocess 1Gb, so it is possible that the whole run takes 2Gb.

nicolas-grekas:
Please let us know when you're more confident with the proposed changes.

theofidry:
Pew I have no idea. Part of the issue is, I'm not sure how you test this. How have _you_ tested it so far?

src/ScriptExecutor.php
theofidry: I do not understand what you mean. If you executed composer with a specific memory limit, Composer will have restarted with persistent settings where the current memory limit (and the one that XdebugHandler will use) is that one.

But I admit I think it deserves to be properly checked rather than merged blindly. It needs a fix, but we should make sure this change is the right fix.

theofidry:
Closing as actually this is working as intended. Flex wants to execute a PHP Process as if it was the original command, not the restarted process (if restarted) with modified settings. So if Composer disables Xdebug, the process that flex starts should have Xdebug _on_.

The current code _could_ be refactored to use `XdebugHandler` still instead of the environment variable directly. But it is not a requirement.

PR #996, opened 2023-10-22
AssetMapper: upgrade packages when needed

weaverryan:
Hi!

The scenario:

* User installs `symfony/ux-autocomplete`. Flex adds `tom-select` at version `2.2.3` to `importmap.php`
* 3 months later, user upgrades `symfony/ux-autocomplete`. The new version now requires `tom-select` at `^2.5`.

If we do nothing, the user's `tom-select` is out of date and the user won't even know about it. The `^2.5` constraint is defined in the `symfony/ux-autocomplete` [package.json file](https://github.com/symfony/ux/blob/97fe2cd62e70e329b85570e0b797833882930754/src/Autocomplete/assets/package.json#L23), but nothing enforces that or notifies the user.

With this small change, during a composer update/require, we look at the dependencies of all UX packages and compare them against the version in `importmap.php`. If they do not match, the package is upgraded to a version that matches.

Screenshot 2023-10-22 at 1 45 41 PM

We're not creating a fully-fledge package management by any means, but we don't need to. With a few notifications and things like this, we can keep the user's dependencies in sync with each other.

Unrelated: don't forget about that really cool other PR #978 ;) 

Cheers!

### DIFF
--- a/src/Flex.php
+++ b/src/Flex.php
@@ -587,7 +587,7 @@ class Flex implements PluginInterface, EventSubscriberInterface
         $vendorDir = trim((new Filesystem())->makePathRelative($this->config->get('vendor-dir'), $rootDir), '/');
 
         $executor = new ScriptExecutor($this->composer, $this->io, $this->options);
-        $synchronizer = new PackageJsonSynchronizer($rootDir, $vendorDir, $executor);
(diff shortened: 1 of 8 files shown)


weaverryan:
fabbot failures are incorrect - they're trying to add PHP 8 only functions, while 1.x works on 7.1.

fabpot (approved):

fabpot:
Thank you @weaverryan.

Issue #974, opened 2023-05-03
composer recipes:update fails with curl error


lazka:
Using symfony/flex v1.19.5, when using the "recipes:update" command I get the following for every package:

```bash
$composer recipes:update
Which outdated recipe would you like to update? (default: 0)
  [0] symfony/apache-pack
  [1] symfony/flex
  [2] symfony/lock
  [3] symfony/messenger
  [4] symfony/phpunit-bridge
 > 1
  Updating recipe for symfony/flex...

                        
   Yes! Recipe updated! 
                        

  Run git status or git diff --cached to see the changes.
  When you're ready, commit these changes like normal.

  Calculating CHANGELOG...
In CurlDownloader.php line 623:
                                                                                                                       
  The "https://api.github.com/search/issues?q=fcda4f8f67e48ff06163debe049416c698e4f2ae" file could not be downloaded   
  (HTTP/2 422 ):                                                                                                       
  {"message":"Query must include 'is:issue' or 'is:pull-request'","documentation_url":"https://docs.github.com/rest/r  
  eference/search#search-issues-and-pull-requests"}
```

weaverryan:
Hi!

I'm guessing this was some weird, temporary issue. It shows the URL, and that loads just fine for me right now.

Cheers!

leonboot:
I experienced the same issue when updating a flex recipe. I think it might have something to do with the new fine grained access tokens (`github_pat_` prefixed). I had one in my `~/.composer/auth.json` file and replaced it with a classic token (`ghp_` prefixed). That seemed to solve the issue. Might be a false positive, or perhaps the new fine grained tokens require a specific permission in order for them to work when using with Flex?

micheh:
According to the [documentation](https://docs.github.com/en/rest/search/search?apiVersion=2022-11-28#search-issues-and-pull-requests), it is required to either add `is:issue` or `is:pull-request` to the `?q=` parameter when using the new fine grained access token. This seems to be a new restriction of the access token and not related to the permissions itself, as even a fine grained access token with all permissions requires the query parameter to be set.

PR #1005, opened 2023-12-05
fix(recipes:update): github api now requires is:pull-request parameter

alainsharemat:
This fix is related to issue #974.

While running `recipes:update`  I got this error:

```shell
Calculating CHANGELOG...
In CurlDownloader.php line 630:

  The "https://api.github.com/search/issues?q=95e75e5e424c5f5f611cc57073e399baaf287570" file could not be downloaded (HTTP/2 422 ):                                  
  {"message":"Query must include 'is:issue' or 'is:pull-request'","documentation_url":"https://docs.github.com/rest/search/search#search-issues-and-pull-requests"}
```

According to github's [documentation](https://docs.github.com/en/rest/search/search?apiVersion=2022-11-28#search-issues-and-pull-requests) : `Requests that don't include the is:issue or is:pull-request qualifier will receive an HTTP 422 Unprocessable Entity response`.

### DIFF
--- a/src/GithubApi.php
+++ b/src/GithubApi.php
@@ -128,7 +128,7 @@ class GithubApi
 
     public function getPullRequestForCommit(string $commit, string $repo): ?array
     {
-        $data = $this->requestGitHubApi('https://api.github.com/search/issues?q='.$commit);
+        $data = $this->requestGitHubApi('https://api.github.com/search/issues?q='.$commit.'+is:pull-request');
 
         if (0 === \count($data['items'])) {
             return null;


maxhelias (approved):

xabbuh (approved):

src/GithubApi.php
```diff
@@ -142,7 +142,7 @@ public function getPullRequestForCommit(string $commit, string $repo): ?array
         $bestItem = null;
         foreach ($data['items'] as $item) {
             // make sure the PR referenced isn't from a different repository
-            if (false === strpos($item['html_url'], sprintf('%s/pull', $repositoryName))) {
+            if (!str_contains($item['html_url'], sprintf('%s/pull', $repositoryName))) {
```
fabpot: Can you revert the 2 unrelated changes?

fabpot (changes requested):
Should also be rebased on 1.x.

src/GithubApi.php
alainsharemat: I did. Had to commit this code to make fabbot.io happy in the first place.

fabpot:
Thank you @alainsharemat.

PR #1011, opened 2024-02-08
fix jsonPath in ComposerScriptsConfigurator::update

### LINKED ISSUE #1010: ComposerScriptsConfigurator can break "composer.json" filename in patch
https://github.com/symfony/flex/blob/bec213c39511eda66663baa2ee7440c65f89c695/src/Configurator/ComposerScriptsConfigurator.php#L52

In case when root dir = ".", it replaces the dot in filename, and then applying the patch is broken with: 
```
NOTE:
  The file composerjson was not updated because it doesn't exist in your app.
```

Reported in https://github.com/symfony/recipes/issues/1271

klkvsk:
Fixes https://github.com/symfony/flex/issues/1010

https://github.com/symfony/flex/blob/bec213c39511eda66663baa2ee7440c65f89c695/src/Configurator/ComposerScriptsConfigurator.php#L52

In case when root dir = ".", it replaces the dot in filename, and then applying the patch is broken with: 
```
NOTE:
  The file composerjson was not updated because it doesn't exist in your app.
```

Change to using Path::makeRelative for proper result.

### DIFF
--- a/src/Configurator/ComposerScriptsConfigurator.php
+++ b/src/Configurator/ComposerScriptsConfigurator.php
@@ -14,7 +14,6 @@ namespace Symfony\Flex\Configurator;
 use Composer\Factory;
 use Composer\Json\JsonFile;
 use Composer\Json\JsonManipulator;
-use Symfony\Component\Filesystem\Path;
 use Symfony\Flex\Lock;
 use Symfony\Flex\Recipe;
 use Symfony\Flex\Update\RecipeUpdate;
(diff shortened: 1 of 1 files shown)


src/Configurator/ComposerScriptsConfigurator.php
```diff
@@ -49,7 +50,7 @@ public function unconfigure(Recipe $recipe, $scripts, Lock $lock)
     public function update(RecipeUpdate $recipeUpdate, array $originalConfig, array $newConfig): void
     {
         $json = new JsonFile(Factory::getComposerFile());
-        $jsonPath = ltrim(str_replace($recipeUpdate->getRootDir(), '', $json->getPath()), '/\\');
+        $jsonPath = Path::makeRelative($json->getPath(), $recipeUpdate->getRootDir());
```
maxhelias: Flex has no dependencies other than `composer-plugin-api`, so it is not possible to use an external package.

maxhelias (changes requested):

PR #1021, opened 2024-04-17
Fix a merge-conflict residue

heiglandreas:
Due to the merge-conflict marker the code could not be executed and failed CI

### DIFF
--- a/src/Downloader.php
+++ b/src/Downloader.php
@@ -51,7 +51,7 @@ class Downloader
     private $enabled = true;
     private $composer;
 
-    public function __construct(Composer $composer, IoInterface $io, HttpDownloader $rfs)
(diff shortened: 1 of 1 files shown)


src/Downloader.php
```diff
@@ -51,11 +51,7 @@ class Downloader
     private $enabled = true;
     private $composer;
 
-<<<<<<< HEAD
     public function __construct(Composer $composer, IoInterface $io, HttpDownloader $rfs)
```
xabbuh:
```suggestion
    public function __construct(Composer $composer, IOInterface $io, HttpDownloader $rfs)
```

nicolas-grekas:
Oops, thanks!

Issue #1009, opened 2024-01-10
Remove importmap entries when uninstalling a package


weaverryan:
When you install a UX package, if you have AssetMapper installed, we `importmap:require` the packages you need. We should also `importmap:remove` those when the package is uninstalled.

https://github.com/symfony/recipes/issues/1089#issuecomment-1885459144

PR #1026, opened 2024-05-21
[AssetMapper] Allow to define entrypoint in importmap.php

Jibbarth:
Hello :wave:

I wanted that one of my package to be able to define an asset as entrypoint inside importmap.php to be allow call `importmap('@mypackage/entry.js')` inside a template of the package. 

I made a naive modification, not sure if it was the purpose of the `[symfony][entrypoints]` node of the package.json. 

Maybe it would be fit directly inside the `[symfony][importmap]` node. 

WDYT ? 

_UPDATE_ : see latest proposal here https://github.com/symfony/flex/pull/1026#issuecomment-2629386272

### DIFF
--- a/src/PackageJsonSynchronizer.php
+++ b/src/PackageJsonSynchronizer.php
@@ -104,7 +104,7 @@ class PackageJsonSynchronizer
 
         foreach (['dependencies' => $jsDependencies, 'devDependencies' => $jsDevDependencies] as $key => $packages) {
             foreach ($packages as $name => $version) {
-                if ('@' !== $name[0] || 0 !== strpos($version, 'file:'.$this->vendorDir.'/') || false === strpos($version, '/assets')) {
+                if ('@' !== $name[0] || !str_starts_with($version, 'file:'.$this->vendorDir.'/') || !str_contains($version, '/assets')) {
                     continue;
                 }
                 if (file_exists($this->rootDir.'/'.substr($version, 5).'/package.json')) {
@@ -149,20 +149,36 @@ class PackageJsonSynchronizer
         $dependencies = [];
 
         foreach ($packageJson->read()['symfony']['importmap'] ?? [] as $importMapName => $constraintConfig) {
-            if (\is_array($constraintConfig)) {
-                $constraint = $constraintConfig['version'] ?? [];
-                $package = $constraintConfig['package'] ?? $importMapName;
-            } else {
+            if (\is_string($constraintConfig)) {
+                // Matches string constraint, like "^3.0" or "path:%PACKAGE%/script.js"
                 $constraint = $constraintConfig;
                 $package = $importMapName;
+                $entrypoint = false;
+            } elseif (\is_array($constraintConfig)) {
+                // Matches array constraint, like {"version":"^3.0"} or {"version":"path:%PACKAGE%/script.js","entrypoint":true}
+                // Note that non-path assets can't be entrypoint
+                $constraint = $constraintConfig['version'] ?? '';
+                $package = $constraintConfig['package'] ?? $importMapName;
+                $entrypoint = $constraintConfig['entrypoint'] ?? false;
+            } else {
+                throw new \InvalidArgumentException(\sprintf('Invalid constraint config for key "%s": "%s" given, array or string expected.', $importMapName, var_export($constraintConfig, true)));
             }
 
-            if (0 === strpos($constraint, 'path:')) {
+            // When "$constraintConfig" matches one of the following cases:
+            // - "entrypoint:%PACKAGE%/script.js"
+            // - {"version": "entrypoint:%PACKAGE%/script.js"}
+            if (str_starts_with($constraint, 'entrypoint:')) {
+                $entrypoint = true;
+                $constraint = substr_replace($constraint, 'path:', 0, \strlen('entrypoint:'));
+            }
+
+            if (str_starts_with($constraint, 'path:')) {
                 $path = substr($constraint, 5);
                 $path = str_replace('%PACKAGE%', \dirname($packageJson->getPath()), $path);
 
                 $dependencies[$importMapName] = [
                     'path' => $path,
+                    'entrypoint' => $entrypoint,
                 ];
 
                 continue;
@@ -239,7 +255,7 @@ class PackageJsonSynchronizer
     }
 
     /**
-     * @param array<string, array{path?: string, package?: string, version?: string}> $importMapEntries
+     * @param array<string, array{path?: string, package?: string, version?: string, entrypoint?: bool}> $importMapEntries
      */
     private function updateImportMap(array $importMapEntries): void
     {
@@ -264,11 +280,15 @@ class PackageJsonSynchronizer
                     continue;
                 }
 
-                $this->io->writeError(sprintf('Updating package <comment>%s</> from <info>%s</> to <info>%s</>.', $name, $version, $versionConstraint));
+                $this->io->writeError(\sprintf('Updating package <comment>%s</> from <info>%s</> to <info>%s</>.', $name, $version, $versionConstraint));
             }
 
             if (isset($importMapEntry['path'])) {
                 $arguments = [$name, '--path='.$importMapEntry['path']];
+                if (isset($importMapEntry['entrypoint']) && true === $importMapEntry['entrypoint']) {
+                    $arguments[] = '--entrypoint';
+                }
+
                 $this->scriptExecutor->execute(
                     'symfony-cmd',
                     'importmap:require',
@@ -293,7 +313,7 @@ class PackageJsonSynchronizer
                 continue;
             }
 
-            throw new \InvalidArgumentException(sprintf('Invalid importmap entry: "%s".', var_export($importMapEntry, true)));
+            throw new \InvalidArgumentException(\sprintf('Invalid importmap entry: "%s".', var_export($importMapEntry, true)));
         }
     }
 
--- a/tests/Fixtures/packageJson/vendor/symfony/importmap-invalid-constraint-package/assets/package.json
+++ b/tests/Fixtures/packageJson/vendor/symfony/importmap-invalid-constraint-package/assets/package.json
@@ -1,0 +1,7 @@
+{
+    "symfony": {
+        "importmap": {
+            "@symfony/test": true
+        }
+    }
+}
--- a/tests/Fixtures/packageJson/vendor/symfony/new-package/assets/package.json
+++ b/tests/Fixtures/packageJson/vendor/symfony/new-package/assets/package.json
@@ -14,6 +14,11 @@
             "@hotcake/foo": "^1.9.0",
             "@symfony/new-package": {
                 "version": "path:%PACKAGE%/dist/loader.js"
+            },
+            "@symfony/new-package/entry.js": "entrypoint:%PACKAGE%/entry.js",
+            "@symfony/new-package/entry2.js": {
+                "version": "path:%PACKAGE%/entry2.js",
+                "entrypoint": true
             }
         }
     },
--- a/tests/PackageJsonSynchronizerTest.php
+++ b/tests/PackageJsonSynchronizerTest.php
@@ -323,11 +323,16 @@ class PackageJsonSynchronizerTest extends TestCase
         file_put_contents($this->tempDir.'/importmap.php', '<?php return [];');
 
         $fileModulePath = $this->tempDir.'/vendor/symfony/new-package/assets/dist/loader.js';
-        $this->scriptExecutor->expects($this->exactly(2))
+        $entrypointPath = $this->tempDir.'/vendor/symfony/new-package/assets/entry.js';
+        $secondEntrypointPath = $this->tempDir.'/vendor/symfony/new-package/assets/entry2.js';
+
+        $this->scriptExecutor->expects($this->exactly(4))
             ->method('execute')
             ->withConsecutive(
                 ['symfony-cmd', 'importmap:require', ['@hotcake/foo@^1.9.0']],
-                ['symfony-cmd', 'importmap:require', ['@symfony/new-package', '--path='.$fileModulePath]]
+                ['symfony-cmd', 'importmap:require', ['@symfony/new-package', '--path='.$fileModulePath]],
+                ['symfony-cmd', 'importmap:require', ['@symfony/new-package/entry.js', '--path='.$entrypointPath, '--entrypoint']],
+                ['symfony-cmd', 'importmap:require', ['@symfony/new-package/entry2.js', '--path='.$secondEntrypointPath, '--entrypoint']],
             );
 
         $this->synchronizer->synchronize([
@@ -396,14 +401,19 @@ class PackageJsonSynchronizerTest extends TestCase
                 'version' => '1.8.0',
             ],
         ];
-        file_put_contents($this->tempDir.'/importmap.php', sprintf('<?php return %s;', var_export($importMap, true)));
+        file_put_contents($this->tempDir.'/importmap.php', \sprintf('<?php return %s;', var_export($importMap, true)));
 
         $fileModulePath = $this->tempDir.'/vendor/symfony/new-package/assets/dist/loader.js';
-        $this->scriptExecutor->expects($this->exactly(2))
+        $entrypointPath = $this->tempDir.'/vendor/symfony/new-package/assets/entry.js';
+        $secondEntrypointPath = $this->tempDir.'/vendor/symfony/new-package/assets/entry2.js';
+
+        $this->scriptExecutor->expects($this->exactly(4))
             ->method('execute')
             ->withConsecutive(
                 ['symfony-cmd', 'importmap:require', ['@hotcake/foo@^1.9.0']],
-                ['symfony-cmd', 'importmap:require', ['@symfony/new-package', '--path='.$fileModulePath]]
+                ['symfony-cmd', 'importmap:require', ['@symfony/new-package', '--path='.$fileModulePath]],
+                ['symfony-cmd', 'importmap:require', ['@symfony/new-package/entry.js', '--path='.$entrypointPath, '--entrypoint']],
+                ['symfony-cmd', 'importmap:require', ['@symfony/new-package/entry2.js', '--path='.$secondEntrypointPath, '--entrypoint']]
             );
 
         $this->synchronizer->synchronize([
@@ -421,19 +431,41 @@ class PackageJsonSynchronizerTest extends TestCase
                 // constraint in package.json is ^1.9.0
                 'version' => '1.9.1',
             ],
+            '@symfony/new-package/entry2.js' => [
+                'path' => './vendor/symfony/new-package/assets/entry2.js',
+                'entrypoint' => true,
+            ],
         ];
-        file_put_contents($this->tempDir.'/importmap.php', sprintf('<?php return %s;', var_export($importMap, true)));
+        file_put_contents($this->tempDir.'/importmap.php', \sprintf('<?php return %s;', var_export($importMap, true)));
 
         $fileModulePath = $this->tempDir.'/vendor/symfony/new-package/assets/dist/loader.js';
-        $this->scriptExecutor->expects($this->once())
+        $entrypointPath = $this->tempDir.'/vendor/symfony/new-package/assets/entry.js';
+
+        $this->scriptExecutor->expects($this->exactly(2))
             ->method('execute')
             ->withConsecutive(
-                ['symfony-cmd', 'importmap:require', ['@symfony/new-package', '--path='.$fileModulePath]]
+                ['symfony-cmd', 'importmap:require', ['@symfony/new-package', '--path='.$fileModulePath]],
+                ['symfony-cmd', 'importmap:require', ['@symfony/new-package/entry.js', '--path='.$entrypointPath, '--entrypoint']],
             );
 
         $this->synchronizer->synchronize([
             [
                 'name' => 'symfony/new-package',
+                'keywords' => ['symfony-ux'],
+            ],
+        ]);
+    }
+
+    public function testExceptionWhenInvalidImportMapConstraint()
+    {
+        file_put_contents($this->tempDir.'/importmap.php', '<?php return [];');
+
+        $this->expectException(\InvalidArgumentException::class);
+        $this->expectExceptionMessage('Invalid constraint config for key "@symfony/test": "true" given, array or string expected.');
+
+        $this->synchronizer->synchronize([
+            [
+                'name' => 'symfony/importmap-invalid-constraint-package',
                 'keywords' => ['symfony-ux'],
             ],
         ]);


Jibbarth:
relate to https://github.com/symfony/symfony/issues/53912

src/PackageJsonSynchronizer.php
```diff
@@ -267,8 +268,14 @@ private function updateImportMap(array $importMapEntries): void
                 $this->io->writeError(sprintf('Updating package <comment>%s</> from <info>%s</> to <info>%s</>.', $name, $version, $versionConstraint));
             }
 
+            $arguments = [];
```
smnandre: A package with version cannot be an entrypoint... so entrypoints should not be concerned here, no ?

Jibbarth: Yeah, i see what you mean. 

If i remember correctly, it was to avoid too many `if` levels but I can add it inside the following condition

smnandre:
Do we allow bundle to change the public/index.php file? Or to create new webpack configs ? 

I really find weird  that a bundle could install things like this. And a bit dangerous.

And still wait for a use case that require to create a entrypoint .... maybe you could describe yours for the package in question ? That really could help me understand the need.

src/PackageJsonSynchronizer.php
```diff
@@ -147,7 +147,7 @@ private function resolveImportMapPackages($phpPackage): array
         }
 
         $dependencies = [];
-
+        $entrypoints = $packageJson->read()['symfony']['entrypoints'] ?? [];
```
smnandre: Took some time to refresh my memory and re-read all this code 😅 

So .. nope : this is not related to importmap entrypoints, but to stimulus bundle + controllers.json file. Thus should not be used here :) 

Should not be used here

src/PackageJsonSynchronizer.php
```diff
@@ -163,6 +163,7 @@ private function resolveImportMapPackages($phpPackage): array
 
                 $dependencies[$importMapName] = [
                     'path' => $path,
+                    'entrypoint' => \in_array($importMapName, $entrypoints, true),
                 ];
```
smnandre: <details>
<summary>(update) collapsed my comment here: it was not <em>constructive</em></summary>

---

By definition, a dependency cannot be an importmap entrypoint :) 

</details>

lanfisis:
> Do we allow bundle to change the public/index.php file? Or to create new webpack configs ?
> 
> I really find weird that a bundle could install things like this. And a bit dangerous.
> 
> And still wait for a use case that require to create a entrypoint .... maybe you could describe yours for the package in question ? That really could help me understand the need.


[TL;DR] This feature is cool, we need it in flex. As it or in other way, but it would be useful to describe a bundle js dependency as an entry point. 

Hello @smnandre,

I was concern by the topic "Asset Mapper with Bundle" because I work on Sylius applications, and we use themes (drove by the [Sylius Theme Bundle](https://github.com/Sylius/SyliusThemeBundle)). 

The idea behind was to distribute Sylius Theme as Symfony Bundle / Sylius Plugin (that's the same thing) to offer the ability for a developer to make a `composer require my-vendor/my-beautiful-theme`, go to the Sylius admin panel, select the new "My Beautiful Theme" theme, save, on go to the front to see the magic happen.

A Sylius theme package / bundle / plugin it's almost templates (twig files), CSS and JS. Since we have introduced Asset Mapper in our Sylius project, I encountered all the problems describe in this [looong topic](https://github.com/symfony/symfony/issues/53912). Basically: how to smoothly install a bundle with JS and CSS dependencies in a Symfony project where assets are Asset Mapper drove. 

I'm not going to rehash the game and I understand both side arguments. I resolve a lot of my problems using the same methods as Symfony UX packages: add `symfony-ux'` in the `keywords` part of the `composer.json` of my theme Bundle that allow **flex** to install in the `importmap.php` dependencies declared in the `assets/packages.json` like this: 

```json
{
  "name": "my-vendor/my-beautiful-theme",
  "description": "Beautiful Theme for Sylius",
  "license": "MIT",
  "version": "1.0.0",
  "symfony": {
    "importmap": {
      "alpinejs": "3.13.2",
      "flickity": "3.0.0",
      "my-vendor/my-beautiful-theme": "path:%PACKAGE%/my-entrypoint.js"
    }
  }
}
```

 I use the idea describe [in this comment ](https://github.com/symfony/symfony/issues/53912#issuecomment-2591275217) to declare theme assets folder as Asset Mapper path. (Part 2: How do I expose my bundle's own assets to a consumer of the bundle?)

And it's work like a charm. 🎉

But now there is my point: theme bundles comes with the whole Sylius app templates. Whole app means that somewhere we use `{{ importmap('my-vendor/my-beautiful-theme') }}` in a twig file to import. But, if flex have correctly added `my-vendor/my-beautiful-theme` in the `impormap.php` file, it was not declared as an entry point and there is an exception when front page is loaded. I have to ask my theme bundle users to manually transform 
```php
    // Automatically added by flex FTW but it miss that this is an entrypoint
    'my-vendor/my-beautiful-theme' => [
        'path' => './vendor/my-vendor/my-beautiful-theme/assets/my-entrypoint.js'
    ],
```
to
```php
    '@monsieurbiz/master-theme' => [
        'path' => './vendor/my-vendor/my-beautiful-theme/assets/my-entrypoint.js',
        'entrypoint' => true,
    ],
```
in the `importmap.php` after theme bundle installation and it's not cool.

That's why I agree with @Jibbarth: It woul be very very cool to have this feature!

Jibbarth:
Thanks @lanfisis 👍

 I totally forgot to reply here, sorry @smnandre, but what I want to achieve is exactly what @lanfisis describe above ☺️

smnandre:
Well, once more.... we're not talking about "bundle assets" but `third-party external dependencies` that are not present `in` your bundle. 

Because assets from your bundle _**are exposed**_ via both asset package and AssetMapper.

I have no power to prevent you for moving on this topic, and no intention to do so...
As long I don't have megabytes of data coming from nowhere and auto-installed next time I `composer update` my projects 😅 

Meaning this should, i think, be a new configuration key somewhere.. 
(cannot be the [Stimulus one](https://github.com/symfony/flex/pull/1026#pullrequestreview-2504263279) that has another usage)

lanfisis:
Hi @smnandre,

To be honest, I'm not sure to understand your reply because it seems not related to my post.  You mention `third-party external dependencies` instead of `bundle dependencies` but I mainly speak about `bundle dependencies`. 

To be clear:
* I am the developper of a third party bundle called `my-beautiful-theme`.
* I've added `symfony-ux` in the keyword of the `composer.json` file [like this UX bundle](https://github.com/symfony/ux/blob/ec09cdff7dfc00b71cc7097618df4a7274c24deb/src/Translator/composer.json#L6)
*  I've added an `assets/package.json` file with **BOTH** external package (e.g. `"alpinejs": "3.13.2"`) and local to the bundle JS files (e.g. `"my-vendor/my-beautiful-theme": "path:%PACKAGE%/my-entrypoint.js"`) [always like this UX bundle](https://github.com/symfony/ux/blob/ec09cdff7dfc00b71cc7097618df4a7274c24deb/src/Translator/assets/package.json#L18)
* thanks to flex, because of the 2 previous actions, when I require this bundle in my Symfony app, **BOTH** external package and bundle JS file are added to the `importmap.php` of my app.

This process exists and works perfectly. No hack, no trick, no billions of external packages. This is how Symfony UX bundle works too. 


The only purpose of this PR, and my only point is: let us the capability to precise that one or more bundle JS local files are entry points in `importmap.php`. That's all. No dirty extra features with hidden package download. 

I you `composer require` a Symfony US package, do you think that you
> have megabytes of data coming from nowhere and auto-installed next time I composer update my projects 😅

If the answer is _yes_, you probably not use Symfony UX and you are probably not the right person to deal with our point about entrypoint.

If the answer is _no_, you have to understand that this small feature could be a game changer for peoples who work on third party bundle with app sized JS file (theme bundle, full admin bundle, dashboard bundle, ....)

Move from actual `package.json` file like :
```json
{
  "name": "my-vendor/my-beautiful-theme",
  "description": "Beautiful Theme for Sylius",
  "license": "MIT",
  "version": "1.0.0",
  "symfony": {
    "importmap": {
      "alpinejs": "3.13.2",
      "flickity": "3.0.0",
      "my-vendor/my-beautiful-theme": "path:%PACKAGE%/my-entrypoint.js"
    }
  }
}
```
to
```json
{
  "name": "my-vendor/my-beautiful-theme",
  "description": "Beautiful Theme for Sylius",
  "license": "MIT",
  "version": "1.0.0",
  "symfony": {
    "entrypoints": ["my-vendor/my-beautiful-theme"]
    "importmap": {
      "alpinejs": "3.13.2",
      "flickity": "3.0.0",
      "my-vendor/my-beautiful-theme": "path:%PACKAGE%/my-entrypoint.js"
    }
  }
}
```
for a result from actual `importmap.php` file:
```php
    // Automatically added by flex FTW but it miss that this is an entrypoint
    'my-vendor/my-beautiful-theme' => [
        'path' => './vendor/my-vendor/my-beautiful-theme/assets/my-entrypoint.js'
    ],
```
to
```php
    '@monsieurbiz/master-theme' => [
        'path' => './vendor/my-vendor/my-beautiful-theme/assets/my-entrypoint.js',
        'entrypoint' => true,
    ],
```
should not create such a debate to my mind.

But, I know you have to address tons of different problems for the Symfony community, and that's a huge and time-consuming work. I'm open to try to answer to your questions and ready to try to calm your fears for this point.

Cheers

smnandre:
Hi @lanfisis! 

The general asset situation is currently a bit of a mess, and I can assure you I've been once or twice using Symfony UX packages :) 

I'm saying again and again bundles can expose _**their**_ assets, so it's only a question about external dependencies and automatic installation. And i will precise this everytime it's needed, I don't want anyone to believe this is not the case currently. 

> should not create such a debate to my mind.

This is why I'm telling you I won't nor want (nor can) oppose this feature. And I continue to answer on the technicalities and consequences on some of the suggested implementations (see my last comment on this matter. https://github.com/symfony/symfony/issues/53912#issuecomment-2591599401)

lanfisis:
@smnandre,

Thanks for the quick answer (as usual). 

>  I can assure you I've been once or twice using Symfony UX packages :)

I was no doubt about this, my question was more rhetoric than really relevant.

In a previous message, you said:

> Do we allow bundle to change the public/index.php file? Or to create new webpack configs ?
>    
> I really find weird that a bundle could install things like this. And a bit dangerous.

But this is a flex feature[ implemented 2 years ago by Ryan and Kevin](https://github.com/symfony/flex/blame/14189260e2513403467d8897604ae85578713219/src/PackageJsonSynchronizer.php#L81), and used by a lot of official bundles. Almost all the UX bundle add external dependencies to `importmap.json`. 
That's a common practice and a fact: dependencies into `assets/packages.json` under node `["symfony"]["importmap"]` are automatically added to `importmap.php`. 

>So .. nope : this is not related to importmap entrypoints, but to stimulus bundle + controllers.json file. Thus should not be used here :)

That's not true. It is used to manage any bundle located js file as [you can see here](https://github.com/symfony/ux/blob/ec09cdff7dfc00b71cc7097618df4a7274c24deb/src/React/assets/package.json#L28).  There is absolutely nothing in the code that constraint the use of this feature for Stimulus controller. 

> And still wait for a use case that require to create a entrypoint .... maybe you could describe yours for the package in question ? That really could help me understand the need.

I hope I have to show you at least one example of why we need this feature. 

> I'm saying again and again bundles can expose their assets, so it's only a question about external dependencies.

Nope, you are the only one who deal with this topic at the beginning. Adding external dependencies and / or JS bundle files into the `importmap.php` file automatically is a reality. It's use by official bundle and probably community and private bundle too. You don't think it's a good practice: that's ok, you could promote the idea to remove this feature from Flex.  But to me, this is a discussion you should have with your partners of the UX core team. Not into the reply of this PR whose only goal is to add a small improvement to an existing feature.

smnandre:
> That's not true. It is used to manage any bundle located js file as [you can see here](https://github.com/symfony/ux/blob/ec09cdff7dfc00b71cc7097618df4a7274c24deb/src/React/assets/package.json#L28). There is absolutely nothing in the code that constraint the use of this feature for Stimulus controller.

There is not in this file any Symfony entrypoint key. When i comment something specific about the code, it is only about the code. No thing more .


—- 


You keep comparing to multiple UX packages. But you must have seen none of them add a new entrypoint.  Which is specifically what you say you want here. So yeah i’m not getting the argument here. 

—-

Flex is not managed by the UX Core Team, nor is AssetMapper. The StimulusBundle is, but still follows the BC promise of Symfony in general. 

—- 

The fact you seem to ignore or let aside the difference between entrypoints and other dependencies proves me the entire system must be re-documented and/or clarified, probably changed too.

—

On the other PR I explain certain things about technical points and try to separate the very différent things asked related to bundles. 
Hzre, the PR implémentation asked for feedback on the key used and I say: I dont think this is the food key to do so. This key had another meaning. 

—

You wont convince me and this is not a problem and should not be one. 

One last time: I cannot and will not block this feature. I can point some technicalities to avoid messing with users that never signed for it. But allow me to not be celebrating it either.

lanfisis:
Ok, so the key is probably not good named. I agree with this. And this comment is surely relevant in the scope of this PR. 
But too me this is the only one. 

>You keep comparing to multiple UX packages. But you must have seen none of them add a new entrypoint. Which is specifically what you say you want here. So yeah i’m not getting the argument here.

I use UX bundle as example because this is the only way to add dependencies on `importmap.php` file. But not only UX bundle want to to this.  Maybe the keyword should be rename `asset-mapper-aware` instead of `symfony-ux` but that's a technical debate.

My point is:

UX packages have to add JS dependencies on `importmap.php`file => no problemo, Kevin and Ryan add some code in flex to do this and it works like a charm. [Absolutely no one has any reservations about the  idea.](https://github.com/symfony/flex/pull/975) 

Peoples of the community said: in the context certain kind of bundle, as I describe and I can describe again if needed, we want to do exacltly the same thing but we want to precise that one of this package is an entrypoint because the bundle comes with templates that include this entrypoint => we get comments about the fact that's "weird" and that there is no concrete application. 

Let's assume "that" is weird to me.

> The fact you seem to ignore or let aside the difference between entrypoints and other dependencies proves me the entire system must be re-documented and/or clarified, probably changed too.

Be sure that I really understand what an entry point is. I'm working on the ability to make Sylius Theme bundle distributable with an asset mapper compatibility for a year. Because there's no documentation, it takes me hours to find how to smoothly, backed by official Symfony tool and trying to avoid dirty fixes, make this working by reading codes source of Asset Mapper, Flex, UX bundles, .... So you're right: this need to be documented and clarified. But this is a group work that have to be backed by the advices and the need of the entire Symfony community, not your own personal opinion.

> You wont convince me and this is not a problem and should not be one.
One last time: I cannot and will not block this feature. I can point some technicalities to avoid messing with users that never signed for it. But allow me to not be celebrating it either.

You said "I'm saying again and again bundles can expose their assets ..." and we (@Jibbarth and I) just add "... with the capability to precise that one of this asset is an entry point".  

Sorry, but it's to little drive me crazy to see you give your negative point of you  like "I really find weird that a bundle could install things like this. And a bit dangerous."  or " But allow me to not be celebrating it either." on PR like this where people have takes time to produce code, are aware about technical observation and describe real life use case.  This is only based on your guts feeling too me, no based on fact. But, agree to disagree about that an asset bundle could be or not an entrypoint.

I hope that @Jibbarth will takes time to change his code according to your needs and I hope that this feature will be available on flex soon because it's a real life needs.

Have a nice day.

smnandre:
@lanfisis reached out to propose a private discussion to address and smooth out differences in perspective and clarify a few points. I gladly accept (we’ll aim to do this as soon as possible) and truly appreciate the positive and constructive approach.

src/PackageJsonSynchronizer.php
```diff
@@ -267,8 +268,14 @@ private function updateImportMap(array $importMapEntries): void
                 $this->io->writeError(sprintf('Updating package <comment>%s</> from <info>%s</> to <info>%s</>.', $name, $version, $versionConstraint));
             }
 
+            $arguments = [];
+            if (isset($importMapEntry['entrypoint']) && true === $importMapEntry['entrypoint']) {
+                $arguments[] = '--entrypoint';
+            }
```
smnandre:
```suggestion
            if (isset($importMapEntry['entrypoint']) && true === $importMapEntry['entrypoint']) {
                $arguments[] = '--entrypoint';
            }
```

This must be updated (moved in the code -- see below) as it will potentially generate an incorrect state.

## Entrypoints cannot be remote packages.

In ImportMapGenerator

```php
        if ($rootImportEntries->get($entryName)->isRemotePackage()) {
            throw new \InvalidArgumentException(\sprintf('The entrypoint "%s" is a remote package and cannot be used as an entrypoint.', $entryName));
        }
```        

## Remote package = package with a version

In ImportMapEntry.php

```php
    public function isRemotePackage(): bool
    {
        return null !== $this->version;
    }
```


## Version and path are exclusive properties

In ImportMapConfigReader.php

```php
                if (isset($data['version'])) {
                    throw new RuntimeException(\sprintf('The importmap entry "%s" cannot have both a "path" and "version" option.', $importName));
                }
```                

## Package have either version or path

In ImportMapConfigReader.php

```php
            if (null === $version) {
                throw new RuntimeException(\sprintf('The importmap entry "%s" must have either a "path" or "version" option.', $importName));
            }
```            


> [!TIP]
>  So entrypoints must be local. And this _must_ be moved line `277` (inside the `if (isset($importMapEntry['path'])) {`)

--- 

Another question: should we allow overriding existing user entrypoints? Do we want to allow that ?

src/PackageJsonSynchronizer.php
smnandre: On a pure technical level, we may want to use **another key**, and definitively one nested under the `importmap` key. 

The [symfony][entrypoints] is used by flex and  is or has been used by stimulus bundle / stimulus bridge / webpack encore bundle.

https://github.com/symfony/flex/blob/14189260e2513403467d8897604ae85578713219/src/PackageJsonSynchronizer.php#L365-L370

Meaning it would have different _meaning_, and `behaviour`, wether someone uses Webpack or AssetMapper.

_Maybe_ there is something to do in the symfony[importmap][ PACKAGE_NAME ] key.

Currently we allow here two shapes

```php
$importMapName =>  [
    'package' => 'foobar',
    'version' => 'foobar',
]
```
and 
```php
$importmapName => 'path:foobar'
```

As seen in my other comment, the first one is out of scope (entrypoints cannot have version).

But I suppose you could here allow another array syntax, like 

```php
$importMapName =>  [
    'package' => 'path:foobarr',
    'entrypoiit' => true,
]
```

Or something like replace `path:`  with `entrypoint:` 

```php
$importMapName => 'entrypoint:foobar',
```

(that would give the path and specifu an entrypointà)

Jibbarth:
Let's sum up.

Currently, we can declare assets for asset-mapper with following ways in package.json

```json
{
  "symfony": {
    "importmap": {
	  "jquery": "^3.7.1", // Remote
      "myjquery": { // custom name
          "version":"^3.7.1",
          "package": "jquery"
      },
      "@mybundle/script.js": {
        "version": "path:%PACKAGE%/script.js",
      },
      "@mybundle/script2.js": "path:%PACKAGE%/script2.js"
    }
  }
}
```

To be able to use one of these as entrypoint, as proposed by @smnandre, I added following possibilies

```json
{
  "symfony": {
    "importmap": {
      "@mybundle/script.js": {
        "version": "path:%PACKAGE%/script.js",
        "entrypoint": true,
      },
      "@mybundle/script2.js": "entrypoint:%PACKAGE%/script2.js",
      "@mybundle/script3.js": {
        "version": "entrypoint:%PACKAGE%/script3.js"
      }
    }
  }
}
```

I guess, for a while, third-party bundle should use the first way 
```
      "@mybundle/script.js": {
        "version": "path:%PACKAGE%/script.js",
        "entrypoint": true,
      }
```
because it will not throw any error while requiring the asset for app that don't have latest flex version, but user still have to add the "entrypoint => true" in their importmap.php.

src/PackageJsonSynchronizer.php
```diff
@@ -147,14 +147,24 @@ private function resolveImportMapPackages($phpPackage): array
...
 
         $dependencies = [];
-
         foreach ($packageJson->read()['symfony']['importmap'] ?? [] as $importMapName => $constraintConfig) {
-            if (\is_array($constraintConfig)) {
-                $constraint = $constraintConfig['version'] ?? [];
-                $package = $constraintConfig['package'] ?? $importMapName;
-            } else {
+            if (\is_string($constraintConfig)) {
+                // Case "jquery": "^3.0" | "@mybundle/script.js": "path:%PACKAGE%/script.js"
```
Jibbarth: I can clean this comment if needed

src/PackageJsonSynchronizer.php
```diff
@@ -267,8 +278,14 @@ private function updateImportMap(array $importMapEntries): void
                 $this->io->writeError(sprintf('Updating package <comment>%s</> from <info>%s</> to <info>%s</>.', $name, $version, $versionConstraint));
             }
 
+            $arguments = [];
```
Jibbarth: this can be revert and handled directly inside the if below

src/PackageJsonSynchronizer.php
smnandre: I'm not sure to read it correctly to be fair 😅

smnandre:
Seems ok!

lanfisis:
Ok to me too !  I've personally adopted another way to add entrypoints after some talks with @smnandre, this small feature could be useful and have no side effects I think.

Jibbarth:
Friendly ping @nicolas-grekas @fabpot :slightly_smiling_face: Is there anything else I can do here ?

src/PackageJsonSynchronizer.php
```diff
@@ -149,12 +149,23 @@ private function resolveImportMapPackages($phpPackage): array
         $dependencies = [];
 
         foreach ($packageJson->read()['symfony']['importmap'] ?? [] as $importMapName => $constraintConfig) {
-            if (\is_array($constraintConfig)) {
-                $constraint = $constraintConfig['version'] ?? [];
-                $package = $constraintConfig['package'] ?? $importMapName;
-            } else {
+            if (\is_string($constraintConfig)) {
+                // Case "jquery": "^3.0" | "@mybundle/script.js": "path:%PACKAGE%/script.js"
```
Kocal:
```suggestion
                // Matches string constraint, like "^3.0" or "path:%PACKAGE%/script.js"
```

src/PackageJsonSynchronizer.php
```diff
@@ -149,12 +149,23 @@ private function resolveImportMapPackages($phpPackage): array
...
                 $constraint = $constraintConfig;
                 $package = $importMapName;
+                $entrypoint = false;
+            } else {
+                // Case "jquery": {"version": "^3.0"} | "@mybundle/script.js": {"version": "path:%PACKAGE%/script.js", "entrypoint": true}
```
Kocal:
```suggestion
                // Matches array constraint, like {"version":"^3.0"} or {"version":"path:%PACKAGE%/script.js","entrypoint":true}
```

src/PackageJsonSynchronizer.php
```diff
@@ -149,12 +149,23 @@ private function resolveImportMapPackages($phpPackage): array
...
-            if (\is_array($constraintConfig)) {
-                $constraint = $constraintConfig['version'] ?? [];
-                $package = $constraintConfig['package'] ?? $importMapName;
-            } else {
+            if (\is_string($constraintConfig)) {
+                // Case "jquery": "^3.0" | "@mybundle/script.js": "path:%PACKAGE%/script.js"
                 $constraint = $constraintConfig;
                 $package = $importMapName;
+                $entrypoint = false;
+            } else {
```
Kocal: Let's check if `$constraintConfig` is an array before accessing properties from it, like it was done before.

Then, we can add a final `else` that throws about `$constraintConfig` invalid type

src/PackageJsonSynchronizer.php
```diff
@@ -149,12 +149,23 @@ private function resolveImportMapPackages($phpPackage): array
...
+                // Case "jquery": {"version": "^3.0"} | "@mybundle/script.js": {"version": "path:%PACKAGE%/script.js", "entrypoint": true}
+                // Note that non-path assets can't be entrypoint
+                $constraint = $constraintConfig['version'] ?? '';
```
Kocal, line 160: I didn't read the whole PR, but before your changes, it was `$constraint = $constraintConfig['version'] ?? []`. 
I believe that's on purpose, `$constraint` must be a string IINW

src/PackageJsonSynchronizer.php
```diff
@@ -149,12 +149,23 @@ private function resolveImportMapPackages($phpPackage): array
...
+                $package = $constraintConfig['package'] ?? $importMapName;
+                $entrypoint = $constraintConfig['entrypoint'] ?? false;
+            }
+
+            // Case "@mybundle/script.js": "entrypoint:%PACKAGE%/script.js" | "@mybundle/script.js": {"version": "entrypoint:%PACKAGE%/script.js"}
```
Kocal:
```suggestion
            // When "$constraintConfig" matches one of the following cases:
            // - "entrypoint:%PACKAGE%/script.js"
            // - {"version": "entrypoint:%PACKAGE%/script.js"}
```

Kocal (approved):
With minor comments

src/PackageJsonSynchronizer.php
Jibbarth: Should we throw an exception or bypass the current package sync, maybe with a warn ?

I guess it's better to not stop all the flex process when a bundle didn't do properly the configuration. WDYT ?

src/PackageJsonSynchronizer.php
Jibbarth: yep, it must be a string.

src/PackageJsonSynchronizer.php
Kocal: If a package contains invalid entries, I think it's better to explicitly throw an exception.

Jibbarth:
Thanks @Kocal for the review :heart_hands: Comments addressed

Kocal (approved):
Thanks!

src/PackageJsonSynchronizer.php
```diff
@@ -149,20 +149,36 @@ private function resolveImportMapPackages($phpPackage): array
...
+                throw new \InvalidArgumentException(\sprintf('Invalid constraint config for key "%s": "%s" given, array or string expected.', $importMapName, var_export($constraintConfig, true)));
             }
 
-            if (0 === strpos($constraint, 'path:')) {
+            // When "$constraintConfig" matches one of the following cases:
+            // - "entrypoint:%PACKAGE%/script.js"
+            // - {"version": "entrypoint:%PACKAGE%/script.js"}
+            if (str_starts_with($constraint, 'entrypoint:')) {
+                $entrypoint = true;
+                $constraint = str_replace('entrypoint:', 'path:', $constraint);
```
nicolas-grekas:
```suggestion
                $constraint = substr_replace($constraint, 'path:', 0, \strlen('entrypoint:'));
```

nicolas-grekas (approved):

nicolas-grekas:
Thank you @Jibbarth.

nicolas-grekas:
Maybe you'd like to have a look at #1009 @Jibbarth? :angel:

PR #1028, opened 2024-10-01
Add DotenvConfigurator, for .env.etc files

nicolas-grekas:
Needed for https://github.com/symfony/recipes/pull/1342


stof (approved):

fabpot (approved):

fabpot:
Thank you @nicolas-grekas.

stof:
how will this work for projects with multiple developers ? As the `.env.local` file is not committed, the dev running the recipe will get the change, but it will be missing for everyone else in the team.

nicolas-grekas:
We could put that secret in .env.dev, WDYT?

PR #1029, opened 2024-10-07
Fix invalid version string "9999999.9999999"

### LINKED ISSUE #887: Running a recipe on dev-master branch generates a version of 9999999.9999999 in lock file
When installing a package it sets the version correctly as "dev-master" but after running the recipe it changes the version to "9999999.9999999" which then throws an error on a cache:clear.

```
"thedrum-developers/production-bundle": {
        "version": "9999999.9999999",
        "recipe": {
            "repo": "gitlab.com/thedrum-developers/recipes",
            "branch": "master",
            "version": "dev-master",
            "ref": "091baa18e93cc037058429db21ce6b3d7dc82ca4"
        },
        "files": [
            "config/packages/prod/the_drum_production.yaml"
        ]
    }
```

I am using private repositories so I am setting the repository in composer.json as below and then running ``` composer require thedrum-developers/production-bundle:dev-master ``` This installs the bundle correctly.

```
{
    "type": "vcs",
    "url": "git@gitlab.com:thedrum-developers/production-bundle.git"
}
```

After running the ``` recipes:install thedrum-developers/production-bundle ``` the issue appears.

### LINKED ISSUE #983: Private bundle 99999.99999 version when branch doesn't contain specific recipe version.
Hi folks!

First of all my issue is related to this one https://github.com/symfony/flex/issues/887 which seems unresolved.

The issue happens to me as well.

When I use private bundle + private recipe repository and the composer.json in "require" section contains "my-org/my-bundle":"dev-master" on the 1st composer install (or update regarding if there is a composer.lock file) all other recipes from symfony 3rd party bundles are installed. At this moment my custom recipe for the particular custom bundle is postponed.

After a bit research inside Symfony\Flex\Downloader I saw that all of the bundles containing recipes when they are on branch (dev-master, dev-trunk, dev-develop, etc) it searches for branch aliasing for this package and when there is no such alias set, it automatically transforms the current version of the recipe to (master|develop|trunk, etc).999999 and lately version_compare function is invoked in order to check if the version (this with .999999) is lower than the version from the recipe. If the version is lower then the recipe is never generated and you need to run it manually. Although manually run, the recipe is written in the symfony.lock file with version 99999.99999 which lately on each of the commands throws:

```
In VersionParser.php line 186:
                                            
  Invalid version string "9999999.9999999"  
```

A workaround for this is to create branch-alias for your dev-master: 1.0-dev inside the bundle's repository.

But my question here is, imagine that you're working in organisation where you have a custom bundle + custom recipe for it and during development the dev- branches in the root project which points to the custom bundle are subject of changing (which is normal, like testing specific change in the core bundle on staging). How in this case would you manage branch changes and running the composer commands which interact particularly with the recipes version?

Thanks in advance.

### LINKED ISSUE #1016: Invalid version string "master.9999999"
Hi,

i'm using private recipes. During development i add a master version instaead of "1.0".
It's cool and Flex found the version and installed it. But in the symfony.lock File in the Version it is 

```
    "XXX/YYY": {
        "version": "master.9999999",
        "recipe": {
            "repo": "XXX.YYY",
            "branch": "master",
            "version": "master",
            "ref": "15ed78413939e240efb1037133e852ac3bdc293c"
        }
    },

```

When i now add a new package oder update an existing Symfony/Flex throws the Error

```
In VersionParser.php line 186:

  [UnexpectedValueException]
  Invalid version string "master.9999999"  

```

When i remove the entry from the symfony.lock file i can update/add my packages, but Symfony/Flex added the bad lines again.

I'm using symfony/flex v2.4.4

nicolas-grekas:
Fix #887
Fix #983
Fix #1016

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -21,6 +21,9 @@
         "symfony/phpunit-bridge": "^4.4.12|^5.0|^6.0",
         "symfony/process": "^4.4|^5.0|^6.0"
     },
+    "conflict": {
+        "composer/semver": "<1.7.2"
+    },
     "autoload": {
         "psr-4": {
             "Symfony\\Flex\\": "src"
--- a/src/Flex.php
+++ b/src/Flex.php
@@ -391,6 +391,10 @@ class Flex implements PluginInterface, EventSubscriberInterface
         $versionParser = new VersionParser();
         $packages = [];
         foreach ($this->lock->all() as $name => $info) {
+            if ('9999999.9999999' === $info['version']) {
+                // Fix invalid versions found in some lock files
+                $info['version'] = '99999.9999999';
+            }
             $packages[] = new Package($name, $versionParser->normalize($info['version']), $info['version']);
         }
 


nicolas-grekas:
I'm not sure why we end up with `9999999.9999999` in symfony.lock but at least this works around it.
The issue is that composer accepts max 5 digits for the major version.
I'm not sure because in Downloader.php, we append `.9999999`, but I don't see where we prefix by `9999999`

PR #1032, opened 2024-11-29
feat: add ability to skip package.json synchronization

### LINKED ISSUE #1017: Make packages.json synchronization optional
There are cases where a project uses Yarn workspaces and Javascript dependencies may be in multiple `packages.json` files, not just the one at the root.
In this case we don't want them added.

wuchen90:
For those who doesn't want the package.json synchronization, this PR offers to opt-out this behavior by setting the config below in extra section of composer.json:

```json
{
  "extra": {
    "symfony/flex": {
      "synchronize_package_json": false
    }
  }
}

```

fix https://github.com/symfony/flex/issues/1017

### DIFF
--- a/src/Flex.php
+++ b/src/Flex.php
@@ -600,6 +600,12 @@ class Flex implements PluginInterface, EventSubscriberInterface
 
     private function synchronizePackageJson(string $rootDir)
     {
+        if (!($this->composer->getPackage()->getExtra()['symfony/flex']['synchronize_package_json'] ?? true)) {
+            $this->io->writeError('<info>Skip synchronizing package.json with PHP packages</>');
+
+            return;
+        }
+
         if (!$this->downloader->isEnabled()) {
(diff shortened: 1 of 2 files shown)


skmedix (commented):
Perhaps it would be better if it logs what exactly has been skipped?

wuchen90:
> Perhaps it would be better if it logs what exactly has been skipped?

@skmedix Currently it already outputs this `<info>Skip synchronizing package.json with PHP packages</>`. What kind of logs do you have in mind?

dmaicher (approved):
I also need this option :+1:

94noni (approved):

skmedix (approved):
I think that's good enough, so LGTM

fabpot (approved):

fabpot:
Thank you @wuchen90.

PR #1034, opened 2025-01-07
feat: Allow multiple lines addition with array for add-lines Configurator

pocky:
Adding multiple lines via the add-lines configurator is currently complex due to the JSON format. This PR introduces support for adding multiple lines through an array.

The array is processed in a straightforward manner using an implode with ~~PHP_EOL~~ `\n`. This approach has minimal impact on the existing code, avoids any regressions, and only modifies a private method. It also maintains compatibility with multi-line strings using "\n".

Test scenarios have been added to ensure proper functionality and validate the new feature.

### DIFF
--- a/src/Configurator/AddLinesConfigurator.php
+++ b/src/Configurator/AddLinesConfigurator.php
@@ -164,9 +164,13 @@ class AddLinesConfigurator extends AbstractConfigurator
         }
     }
 
-    private function getPatchedContents(string $file, string $value, string $position, ?string $target, bool $warnIfMissing): string
+    private function getPatchedContents(string $file, string|array $value, string $position, ?string $target, bool $warnIfMissing): string
     {
         $fileContents = $this->readFile($file);
+
+        if (\is_array($value)) {
+            $value = implode(\PHP_EOL, $value);
+        }
 
         if (false !== strpos($fileContents, $value)) {
             return $fileContents; // already includes value, skip
--- a/tests/Configurator/AddLinesConfiguratorTest.php
+++ b/tests/Configurator/AddLinesConfiguratorTest.php
@@ -566,6 +566,70 @@ console.log(Turbo);
 EOF
             ],
         ];
+
+        yield 'recipe_changes_with_multiline_string' => [
+            ['assets/app.js' => $appJsOriginal],
+            [
+                ['file' => 'assets/app.js', 'position' => 'top', 'content' => "import './bootstrap';"],
+            ],
+            [
+                ['file' => 'assets/app.js', 'position' => 'top', 'content' => "import './stimulus_bootstrap';\nimport { useIntersection } from 'stimulus-use';"],
+            ],
+            ['assets/app.js' => <<<EOF
+import './stimulus_bootstrap';
+import { useIntersection } from 'stimulus-use';
+import * as Turbo from '@hotwired/turbo';
+
+console.log(Turbo);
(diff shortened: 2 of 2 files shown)


pocky:
Should I apply the suggested patch even if it's larger than my changes?

mtarld:
@pocky, yes, you should apply suggestions about `\is_array` and trailing `,` related to your changes.
But you can ignore other suggestions :slightly_smiling_face:

src/Configurator/AddLinesConfigurator.php
```diff
@@ -164,10 +164,14 @@ public function executeUnconfigure(Recipe $recipe, $config): void
         }
     }
 
-    private function getPatchedContents(string $file, string $value, string $position, ?string $target, bool $warnIfMissing): string
+    private function getPatchedContents(string $file, string|array $value, string $position, ?string $target, bool $warnIfMissing): string
     {
         $fileContents = $this->readFile($file);
 
+        if (is_array($value)) {
+            $value = implode(PHP_EOL, $value);
```
stof: Please use `"\n"` instead of `PHP_EOL`. We don't want to generate a mix of LF and CRLF on Windows (and we generate code with LF)

stof:
Instead of adding this in the configurator (which won't be available for older versions of Flex, making recipes incompatible with them if they use an array), shouldn't we rather support using a an array in the source code of the recipes repository but converting that to a string when generating the Flex endpoints with https://github.com/symfony-tools/recipes-checker ? It would make it compatible with existing Flex versions (allowing to start using arrays in `symfony/recipes` immediately instead of having to wait first until support for arrays is widespread enough among installed versions of Flex).

pocky:
@mtarld Thanks! Done (except for the last one)

```diff
@@ -615,7 +615,7 @@
                 ['file' => 'assets/app.js', 'position' => 'top', 'content' => [
                     "import './stimulus_bootstrap';",
                     "import { useIntersection } from 'stimulus-use';",
-                    "",
+                    '',
                     "  console.log(years['2025'] !== years['0225']);",
                 ]],
             ],
```

src/Configurator/AddLinesConfigurator.php
pocky: Thanks for your feedback!

pocky:
@stof 

> Instead of adding this in the configurator (which won't be available for older versions of Flex, making recipes incompatible with them if they use an array), shouldn't we rather support using a an array in the source code of the recipes repository but converting that to a string when generating the Flex endpoints with https://github.com/symfony-tools/recipes-checker ? It would make it compatible with existing Flex versions (allowing to start using arrays in `symfony/recipes` immediately instead of having to wait first until support for arrays is widespread enough among installed versions of Flex).

I don’t know. I wrote my recipe using add-lines and an array like Docker and Compose keys because I thought: "same problem, same solution." Then I realized this wasn’t the case and made this PR.

pocky:
Hello,

I'd like to bring this PR back to life because I still need it. I’ve completely rebased my branch and applied the suggested patch so there are no more pipeline errors.

@nicolas-grekas, what do you think?

nicolas-grekas:
@stof's proposal makes sense to me, it shouldn't be very hard to achieve the same DX without opening compat issues.
Up to give it a try instead?
Let me close here as it's pretty sure we'll prefer the seamless way.
Thanks for submitting!

PR #1035, opened 2025-02-01
Add more entropy to the generated secret

rosier:
Increase the "randomness" of the generated secret by extending the used characters from hexadecimal to base58.

I opted for base58 to keep the secret readable.

### DIFF
--- a/src/Configurator/EnvConfigurator.php
+++ b/src/Configurator/EnvConfigurator.php
@@ -197,9 +197,24 @@ class EnvConfigurator extends AbstractConfigurator
         return $value;
     }
 
-    private function generateRandomBytes($length = 16)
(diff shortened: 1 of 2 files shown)


nicolas-grekas (commented):
:-1: on my side:
- this changes the definition of the $length argument, which was about entropy size and now becomes result's size
- base58 is not worth the extra logic

`return base64_encode(random_bytes($length))` would be just fine if the purpose is having a shorter result. (we could increase the default to 18 bytes to free the base64 string from any trainling `=`.)

rosier:
Hmm, I guess my focus was to much on putting more entropy in the default 32 long secret and overlooked you can just as well increase the entropy by generating a longer secret

PR #1036, opened 2025-02-01
CI: Bump actions and run tests also on newer php versions

rosier:
Fixes:
- The `set-output` command is deprecated and will be disabled soon. 
- The `save-state` command is deprecated and will be disabled soon.


fabpot (approved):

fabpot:
Thank you @rosier.

PR #1037, opened 2025-02-03
CS fixes

rosier:
I did split this in separate commits to make it easier to review.

I can squash this if you prefer that.


rosier:
The fabbot suggests a few more changes but I not sure if all work on PHP 7.1

rosier:
The fabbot suggest one more change but that requires PHP 8

```
-                if (0 === strpos(\get_class($plugin), 'Hirak\Prestissimo\Plugin')) {
+                if (0 === strpos($plugin::class, 'Hirak\Prestissimo\Plugin')) {
```

fabpot (approved):

fabpot:
Thank you @rosier.

PR #1039, opened 2025-02-25
Do not synchronize package.json if symfony/flex is not installed

leofeyer:
Just as the project is not configured and no recipes are loaded if symfony/flex is not in the root `composer.json`, the `package.json` should not be synchronized in this case.


Kocal (approved):

fabpot (approved):

fabpot:
Thank you @leofeyer.

PR #1043, opened 2025-04-05
Added a feature to optionally override existing environment variables set by the system

### LINKED ISSUE #1042: Allow override of existing environment variables set by the system optionally via a flag in DumpEnvCommand
https://github.com/symfony/flex/pull/1043

VarunAgw:
Fixes #1042

### DIFF
--- a/src/Command/DumpEnvCommand.php
+++ b/src/Command/DumpEnvCommand.php
@@ -42,6 +42,7 @@ class DumpEnvCommand extends BaseCommand
                 new InputArgument('env', InputArgument::OPTIONAL, 'The application environment to dump .env files for - e.g. "prod".'),
             ])
             ->addOption('empty', null, InputOption::VALUE_NONE, 'Ignore the content of .env files')
+            ->addOption('override', null, InputOption::VALUE_NONE, 'Whether existing environment variables set by the system should be overridden')
         ;
     }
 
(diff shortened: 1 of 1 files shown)


nicolas-grekas:
Can you expand? Why can't you set your real env var to the value you need? Or why can't you remove said real env vars instead?
You can already override real env vars by listing them in the SYMFONY_DOTENV_VARS env var. Isn't this a better way to cover your use case?
Adding options adds complexity. Better not if possible.

VarunAgw:
Hi @nicolas-grekas 

Here is my code structure

.env
- Some data

.env.local
- APP_ENV=prod_conf1

.env.prod_conf1
- APP_ENV=prod
- APP_NAME=conf1
- some more data

.env.prod_conf2
- APP_ENV=prod
- APP_NAME=conf2
- some more data

Without an override flag, I can't override the value of APP_ENV in .env.prod_conf1

VarunAgw:
I can't override APP_ENV using SYMFONY_DOTENV_VARS feature.

VarunAgw:
Bump @nicolas-grekas

VarunAgw:
Hi @nicolas-grekas Bump

nicolas-grekas:
I'm :-1:
Thanks for the explanations but I think this is too specific.
I think you can achieve what you want using `cp`: instead of editing `.env.local` to switch APP_ENV, use `cp .env.prod_conf1 .env.prod` instead.

VarunAgw:
Sure, thanks for letting me know

PR #1044, opened 2025-05-10
Always include locked dev packages

wouterj:
It took me ages to find this hidden bug.

When fetching the locked repository, Composer by default does not include dev packages. This means that conflicts for dev packages defined by Flex recipes were always completely ignored (like the `phpunit/phpunit: <9.6` conflict in our `symfony/phpunit-bridge` recipe).

This creates big issues for my new PHPUnit recipe (https://github.com/symfony/recipes/pull/1401), as the 10+ recipe is run for applications using PHPUnit 9.

### DIFF
--- a/src/Flex.php
+++ b/src/Flex.php
@@ -600,6 +600,12 @@ class Flex implements PluginInterface, EventSubscriberInterface
 
     private function synchronizePackageJson(string $rootDir)
     {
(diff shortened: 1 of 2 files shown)


fabpot (approved):

fabpot:
Merged in 2.x

fabpot:
Thank you so much @wouterj

PR #1045, opened 2025-05-19
Handle phpunit.xml.dist as an alias of phpunit.dist.xml

nicolas-grekas:
Should smoothen the transition from phpunit.xml.dist to phpunit.dist.xml

Tested on a small app where the file is renamed between installation and removal of phpunit/phpunit, flex tracks the move successfully.


stof (approved):

fabpot:
Thank you @nicolas-grekas.

PR #1056, opened 2025-05-29
Upgrade to PHPUnit 12

### LINKED ISSUE #1051: Bump to PHP 8.1 and upgrade CI to PHPUnit 10
Not PHPUnit 11 because we still need to support for PHP 8.1

e.g. at the moment we have stuff like:

> 1) Symfony\Flex\Tests\Configurator\CopyFromPackageConfiguratorTest::testConfigureAndOverwriteFiles
The at() matcher has been deprecated. It will be removed in PHPUnit 10. Please refactor your test to not rely on the order in which methods are invoked.

Help wanted 🙏

HypeMC:
Resolves #1051

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -16,9 +16,9 @@
     },
     "require-dev": {
         "composer/composer": "^2.1",
+        "phpunit/phpunit": "^12.4",
         "symfony/dotenv": "^6.4.41|^7.4.13|^8.0.13",
         "symfony/filesystem": "^6.4|^7.4|^8.0",
-        "symfony/phpunit-bridge": "^6.4|^7.4|^8.0",
         "symfony/process": "^6.4|^7.4|^8.0"
     },
     "conflict": {
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -1,10 +1,10 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" backupGlobals="true" backupStaticAttributes="false" colors="true" convertErrorsToExceptions="true" convertNoticesToExceptions="true" convertWarningsToExceptions="true" processIsolation="false" stopOnFailure="false" bootstrap="tests/bootstrap.php" xsi:noNamespaceSchemaLocation="https://schema.phpunit.de/9.3/phpunit.xsd">
-  <coverage>
+<phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" backupGlobals="true" colors="true" failOnDeprecation="true" failOnNotice="true" failOnWarning="true" bootstrap="tests/bootstrap.php" xsi:noNamespaceSchemaLocation="vendor/phpunit/phpunit/phpunit.xsd" cacheDirectory=".phpunit.cache">
+  <source ignoreIndirectDeprecations="true">
     <include>
       <directory>./src/</directory>
     </include>
-  </coverage>
+  </source>
   <testsuites>
     <testsuite name="Symfony Flex Test Suite">
       <directory>./tests</directory>
@@ -12,7 +12,5 @@
   </testsuites>
   <php>
     <ini name="error_reporting" value="-1"/>
-    <env name="SYMFONY_DEPRECATIONS_HELPER" value="weak"/>
-    <env name="LC_ALL" value="C"/>
   </php>
 </phpunit>
--- a/tests/Configurator/DockerComposeConfiguratorTest.php
+++ b/tests/Configurator/DockerComposeConfiguratorTest.php
@@ -14,6 +14,8 @@ namespace Symfony\Flex\Tests\Configurator;
 use Composer\Composer;
 use Composer\IO\IOInterface;
 use Composer\Package\RootPackage;
+use PHPUnit\Framework\Attributes\DataProvider;
+use PHPUnit\Framework\MockObject\Stub;
 use PHPUnit\Framework\TestCase;
 use Symfony\Component\Filesystem\Filesystem;
 use Symfony\Flex\Configurator\DockerComposeConfigurator;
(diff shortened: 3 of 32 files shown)


composer.json
```diff
@@ -11,14 +11,14 @@
...
-        "php": ">=8.0",
+        "php": ">=8.1",
         "composer-plugin-api": "^2.1"
     },
     "require-dev": {
         "composer/composer": "^2.1",
+        "phpunit/phpunit": "^10.5",
         "symfony/dotenv": "^5.4|^6.0",
         "symfony/filesystem": "^5.4|^6.0",
-        "symfony/phpunit-bridge": "^5.4|^6.0",
```
xabbuh: I would not entirely drop the bridge as PHPUnit 10 does not fully support handling deprecations IIRC.

stof: the bridge is not compatible with PHPUnit 10. So you cannot keep it (expecting it to report deprecations) when using PHPUnit 10

nicolas-grekas:
Let's bump to v12?

HypeMC:
@nicolas-grekas PHPUnit 12 requires PHP 8.3, so I had to allow older versions for lower PHP versions.

nicolas-grekas:
I did the bump to PHP >= 8.1 in #1067
Rebase needed for any follow up here :)

HypeMC:
> I did the bump to PHP >= 8.1 in #1067 Rebase needed for any follow up here :)

@nicolas-grekas Done

nicolas-grekas:
I forced-pushed a rebase, then realized that while tests are green, they have things to fix (one risky, and `N` ones?)

HypeMC:
@nicolas-grekas Oh, the N test are because of PHPUnit 12.5-dev, guess that's something new, I'll have to check it out.

HypeMC:
@nicolas-grekas Should be good now, see https://github.com/symfony/flex/pull/1056/commits/4dc8127fd171072fe07b47107337f2bd212c8c39.

The main issue was:

> No expectations were configured for the mock object for \<class\>. You should refactor your test code and use a test stub instead.

Also, using `$this->any()` resulted in the same error, eg:

```php
$repository = $this->createMock(InstalledRepositoryInterface::class);
$repository->expects($this->any());
```

tests/Configurator/DockerComposeConfiguratorTest.php
```diff
@@ -496,10 +497,6 @@ public function testUnconfigureFileWithManyMarks()
         $dockerComposeFile = FLEX_TEST_DIR.'/docker-compose.yml';
         file_put_contents($dockerComposeFile, $originalContent);
 
-        /** @var Recipe|\PHPUnit\Framework\MockObject\MockObject $recipe */
-        $recipe = $this->getMockBuilder(Recipe::class)->disableOriginalConstructor()->getMock();
-        $recipe->method('getName')->willReturn('symfony/messenger');
-
```
HypeMC, line 505: This was unused.

phpunit.xml.dist
```diff
@@ -1,18 +1,16 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" backupGlobals="true" backupStaticAttributes="false" colors="true" convertErrorsToExceptions="true" convertNoticesToExceptions="true" convertWarningsToExceptions="true" processIsolation="false" stopOnFailure="false" bootstrap="tests/bootstrap.php" xsi:noNamespaceSchemaLocation="https://schema.phpunit.de/9.3/phpunit.xsd">
-  <coverage>
+<phpunit xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" backupGlobals="true" colors="true" failOnNotice="true" failOnWarning="true" bootstrap="tests/bootstrap.php" xsi:noNamespaceSchemaLocation="vendor/phpunit/phpunit/phpunit.xsd" cacheDirectory=".phpunit.cache">
```
stof: We should use `failOnDeprecation` to match the behavior of the PHPUnit bridge

composer.json
```diff
@@ -16,9 +16,9 @@
     },
     "require-dev": {
         "composer/composer": "^2.1",
+        "phpunit/phpunit": "^12.4",
```
stof, line 19: should we use `^10.5 || ^11.5 || ^12.4` to make it easier for contributors running things locally, instead of relying on the CI replacing the requirement ?

composer.json
```diff
@@ -16,9 +16,9 @@
...
         "symfony/dotenv": "^6.4|^7.4|^8.0",
         "symfony/filesystem": "^6.4|^7.4|^8.0",
-        "symfony/phpunit-bridge": "^6.4|^7.4|^8.0",
```
stof, line 21: the SymfonyExtension defined in the bridge still does the automatic registration of the DebugClassLoader to report deprecations triggered by this autoloader. If the bridge is not used anymore, it might still make sense to depend on error-handler and to register that autoloader in the PHPUnit bootstrap file.

composer.json
HypeMC: The `ignoreIndirectDeprecations` would still be a problem.  I guess we could have two xml files.

stof: it will indeed be an issue for PHPUnit 10 (used on PHP 8.1 only). Maybe we should bump our PHP requirement to 8.2 instead of the bump to 8.1 done in https://github.com/symfony/flex/pull/1067
/cc @nicolas-grekas

nicolas-grekas: I wouldn't until 6.4 is maintained.

composer.json
HypeMC: @stof I just double-checked, the `DebugClassLoader` is registered only if the `error-handler` component is installed, which was never the case. So removing the bridge doesn't change this behavior.

phpunit.xml.dist
HypeMC: Done.

composer.json
stof: ah, I thought the phpunit-bridge was depending on error-handler in newer version. So this is indeed a separate improvement idea.

nicolas-grekas:
Thank you @HypeMC.

PR #1057, opened 2025-05-30
Don't remove still-referenced files when updating recipes

wouterj:
A twin to #1054 

When a recipe removes a file in a newer recipe version, we should also check if other recipes still reference that file.

Real use-case: This occurs when you need to keep the PhpunitBridge in 7.3 (e.g. you need the clock mock).

### DIFF
--- a/src/Update/RecipePatcher.php
+++ b/src/Update/RecipePatcher.php
@@ -15,6 +15,7 @@ use Composer\IO\IOInterface;
 use Composer\Util\ProcessExecutor;
 use Symfony\Component\Filesystem\Exception\IOException;
 use Symfony\Component\Filesystem\Filesystem;
+use Symfony\Flex\Lock;
 
 class RecipePatcher
 {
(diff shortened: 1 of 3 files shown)


HypeMC (approved):

src/Update/RecipePatcher.php
```diff
@@ -39,11 +42,30 @@ public function __construct(string $rootDir, IOInterface $io)
...
 
+        $nonRemovableFiles = [];
         foreach ($patch->getDeletedFiles() as $deletedFile) {
-            if (file_exists($this->rootDir.'/'.$deletedFile)) {
-                $this->execute(\sprintf('git rm %s', ProcessExecutor::escape($deletedFile)), $this->rootDir);
+            if (!file_exists($this->rootDir.'/'.$deletedFile)) {
+                continue;
+            }
+
+            if (isset($lockedFiles[$deletedFile])) {
```
nicolas-grekas, line 53: Won't this be always true, since the file will be listed all the time because the currently updated recipe has it?

stof: it would have to exclude the patched package from the creation of `lockedFiles`, so that only _other_ recipes lock a file.

fabpot (approved):

fabpot:
Thank you @wouterj.

PR #1059, opened 2025-06-30
UpdateRecipesCommand: add `--next` option

gharlan:
With `symfony:recipes:update` you can update the recipes, but you have to do it one by one.
I suggest to add a `--next` option to update the next outdated package. So you do not need to select the next package interactively.

### DIFF
--- a/tests/Fixtures/update_recipes/composer.json
+++ b/tests/Fixtures/update_recipes/composer.json
@@ -3,11 +3,18 @@
     "license": "proprietary",
     "require": {
         "php": ">=7.1",
-        "symfony/console": "5.4.*"
+        "symfony/console": "5.4.*",
+        "symfony/flex": "@dev"
     },
     "config": {
         "allow-plugins": {
             "symfony/flex": true
         }
-    }
+    },
+    "repositories": [
+        {
+            "type": "path",
+            "url": ".."
+        }
+    ]
 }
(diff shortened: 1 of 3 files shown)


gharlan:
I'm not sure why my new test case fails with 

> Symfony recipes are disabled: "symfony/flex" not found in the root composer.json

But first I want to ask what do you think about this new option?

tests/Fixtures/update_recipes/composer.json
```diff
@@ -3,11 +3,18 @@
...
+        "symfony/console": "5.4.*",
+        "symfony/flex": "@dev"
     },
     "config": {
         "allow-plugins": {
             "symfony/flex": true
         }
-    }
+    },
+    "repositories": [
```
fabpot, line 14: The changes in composer.json should be removed, right?

fabpot (changes requested):
Makes sense to me.

tests/Fixtures/update_recipes/composer.json
gharlan: Without the changes I get this error for the new phpunit test case:

```
1) Symfony\Flex\Tests\Command\UpdateRecipesCommandTest::testCommandUpdatesRecipe with data set #1 (array(true))
Failed asserting that '<warning>Symfony recipes are disabled: "symfony/flex" not found in the root composer.json\n
All packages appear to be up-to-date!\n
' contains "Recipe updated".
```

That's why I added `symfony/flex` to this composer.json (it is not the main one). Don't know if there is maybe a better solution?

nicolas-grekas:
Looks like running in non-interactive mode provides the same experience, no?

`composer recipes:update -n`

PR #1061, opened 2025-07-08
feat: Make DockerComposeConfigurator compatible with "false" string

### LINKED ISSUE #1060: `DockerComposeConfigurator` do not honors `composer.json` configuration
# Description

`DockerComposeConfigurator` do not honors `composer.json` configuration. For example, the following configuration will create _docker compose_ files:

```json
{
...
  "extra": {
    "symfony": {
      "docker": "false"
    }
  }
}
```

However follwing configurations will work as expected, they will not create nor update _docker compose_ files:

```json
{
...
  "extra": {
    "symfony": {
      "docker": false
    }
  }
}
```

```json
{
...
  "extra": {
    "symfony": {
      "docker": "0"
    }
  }
}
```

# How to reproduce

1. Configure your project with the following Composer command `composer config extra.symfony.docker false`.
2. Install a pack that updates `docker compose` files, for example `composer require orm`.
3. Docker compose files will be created/updated even if this is not desired.

# Problem

So clearly the catch is that a string `"false"` is set with the `composer config extra.symfony.docker false` command, and this string is not recognized by PHP as a proper boolean (unlike `"0"` string in my examples). However because `composer config` command can only set strings! There is no way to set a proper boolean.

IMHO, Flex should be capable to recognize `"false"` string as a boolean since `composer config` is an a configuration command provided by Composer itself.

I identified this bug because I run `composer config extra.symfony.docker false` programatically in every project I create.

# Fix

I created an MR that fixes this issue: #1061

jawira:
This fixes #1060 

With this MR `DockerComposeConfigurator` will recognize `"false"` string a proper boolean.

I use `filter_var` function, I think this is a good solution because it requires only a single method call to fix the issue. 

With `filter_var` function:

* true, 1, "true", "yes", "1", and "on" are recognized as `true` as well.
* any other value is `false`.

Additionally this function is already used in the same method some lines below:

https://github.com/symfony/flex/blob/423c36e369361003dc31ef11c5f15fb589e52c01/src/Configurator/DockerComposeConfigurator.php#L115-L117

### DIFF
--- a/src/Configurator/DockerComposeConfigurator.php
+++ b/src/Configurator/DockerComposeConfigurator.php
@@ -101,7 +101,7 @@ class DockerComposeConfigurator extends AbstractConfigurator
         }
 
         if (null !== $dockerPreference = $composer->getPackage()->getExtra()['symfony']['docker'] ?? null) {
-            self::$configureDockerRecipes = $dockerPreference;
+            self::$configureDockerRecipes = filter_var($dockerPreference, \FILTER_VALIDATE_BOOLEAN);
 
             return self::$configureDockerRecipes;
         }


stof (approved):

dunglas (approved):

maxhelias (approved):

fabpot:
Thank you @jawira.

PR #1064, opened 2025-10-08
[UX][Bug] Avoid considering "false" as a truthy value for "synchronize_package_json"

Crovitche-1623:
The current command `composer config "extra.symfony/flex.synchronize_package.json" false` write a string containing `"false"`:

e.g.
```json
{
    {
        "symfony/flex": {
            "synchronize_package_json": "false",
        }
    }
}
```

It does not skip the `package.json` as expected so I suggest to add the `filter_var`, so for all those who use thed previous "wrong" command, it'll work as expected now.

ping @smnandre and @Kocal

related: https://github.com/symfony/ux/pull/3091

### DIFF
--- a/src/Flex.php
+++ b/src/Flex.php
@@ -497,7 +497,7 @@ class Flex implements PluginInterface, EventSubscriberInterface
 
     private function synchronizePackageJson(string $rootDir)
     {
-        if (!($this->composer->getPackage()->getExtra()['symfony/flex']['synchronize_package_json'] ?? true)) {
+        if (!(filter_var($this->composer->getPackage()->getExtra()['symfony/flex']['synchronize_package_json'] ?? true, FILTER_VALIDATE_BOOLEAN))) {
             $this->io->writeError('<info>Skip synchronizing package.json with PHP packages</>');
 
             return;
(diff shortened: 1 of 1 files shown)


Crovitche-1623:
Note: the CI step for CS fails on code that I didn't wrote.

Kocal (approved):

nicolas-grekas:
This doesn't make sense to my. json is a typed language that allows differentiating booleans from strings.

stof:
@nicolas-grekas JSON is typed. But the `composer config` command allowing to change the composer config from the CLI (rather than through manual edition of a JSON file) receives string as input (as any CLI script). And for `extra`, the `config` command of Composer does not have any type conversion rule implemented as it cannot know the expected type of each extra key (as they are precisely not known to Composer)

Crovitche-1623:
@nicolas-grekas and @stof 

The following command with the `--json` argument would also solve the problem :

```composer config --json extra.symfony/flex.synchronize_package_json false```

But I wanted to add this PR for those who already have the `"false"` value and do not want to run the above command one more time.

stof:
I forgot that `composer config` has this `--json` flag.

then I would rather be in favor of reporting an error when they use a string value, instead of adding support for `"false"`

nicolas-grekas:
Yep
Let's close this then, new PR welcome of course.

PR #1065, opened 2025-10-30
Some love for tests - fix Symfony 8 support

nicolas-grekas:
fabbot failures are false-positives


nicolas-grekas:
This turned out to be fixing compatibility with Symfony 8!

PR #1066, opened 2025-10-30
Fix ignoring unstable releases with --prefer-lowest and --prefer-stable

nicolas-grekas:
This is a behavior that has been broken about one year ago by [a change is composer](https://github.com/composer/composer/commit/685add70ec84d18075a202686288304db0d6d265).

The attached implementation is way more robust since it uses a documented hook for plugins.

I tried to explain why this behavior should be the default also for composer without success so far, see:
https://github.com/composer/composer/pull/12581

This is a must when testing lowest deps seriously, backed by years of practice :)

### DIFF
--- a/src/PackageFilter.php
+++ b/src/PackageFilter.php
@@ -30,14 +30,16 @@ class PackageFilter
     private $symfonyConstraints;
     private $downloader;
     private $io;
+    private $ignoreUnstableReleases;
 
-    public function __construct(IOInterface $io, string $symfonyRequire, Downloader $downloader)
+    public function __construct(IOInterface $io, string $symfonyRequire, Downloader $downloader, bool $ignoreUnstableReleases = false)
     {
         $this->versionParser = new VersionParser();
         $this->symfonyRequire = $symfonyRequire;
-        $this->symfonyConstraints = $this->versionParser->parseConstraints($symfonyRequire);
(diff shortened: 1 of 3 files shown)


src/PackageFilter.php
```diff
@@ -48,6 +50,16 @@ public function __construct(IOInterface $io, string $symfonyRequire, Downloader
...
     {
+        if ($this->ignoreUnstableReleases) {
+            $filteredPackages = [];
+            foreach ($data as $package) {
+                if (\in_array($package->getStability(), ['stable', 'dev'], true)) {
+                    $filteredPackages[] = $package;
+                }
+            }
+            $data = $filteredPackages;
+        }
```
Seldaek, line 61: IMO this is very risky and should only be done if the minimum-stability is dev. If not you will for sure break projects using flex together with --prefer-lowest/--prefer-stable. If they require anything that has only an RC/beta/alpha but dev is not allowed, it will fail resolution.

Ideally it would check per package if dev is available and if yes clean pre-release versions, if not keep everything.

nicolas-grekas: I very much doubt there is a use case where testing with lowest deps and neither dev nor stable are available.
I'd rather wait for a real world use case to be described before doing anything :)

PR #1067, opened 2025-10-30
Bump deps, PHP >= 8.1

nicolas-grekas:


GromNaN (approved):

.github/workflows/ci.yaml
```diff
@@ -20,13 +20,12 @@ jobs:
             fail-fast: false
             matrix:
                 include:
-                  - php: '8.0'
-                    composer: 2.2.x
                   - php: '8.1'
+                    composer: 2.1.x
```
stof, line 24: why testing on composer 2.1.x ? the Composer LTS is 2.2.x

nicolas-grekas: Because this:
> "composer-plugin-api": "^2.1"

PR #1070, opened 2025-11-14
Fix dump-env command when .env files reference other env vars

nicolas-grekas:
At the moment, running `composer dump-env` on a 7.4 project generates a broken `APP_SHARE_DIR`.
The reason is that since https://github.com/symfony/recipes/pull/1465 we define that var as `"$APP_PROJECT_DIR/var/share"`
Yet, when the command runs, all env vars are emptied to make the result context-free.

This fixes it by hooking into calls to the `getenv()` function.


nicolas-grekas:
We might also want to make http-kernel conflict with affected versions of flex so that ppl are forced to update flex when moving to 7.4 and don't fall into this issue.

PR #1075, opened 2025-12-23
Add `create_if_missing` option to `add-lines` configurator

OskarStark:
## Why

In `symfony/ai` we basically have an "empty" `ai.yaml` file, as no stores are configured as default. When adding your first store you are not able to leverage the recipe otherwise.

cc @chr-hertel 

## Summary

This PR adds a `create_if_missing` option to the `add-lines` configurator. When using `position: after_target`, if the target is not found in the file, the content specified in `create_if_missing` will be appended to the file, followed by the regular `content`.

### Example usage:

```json
{
    "add-lines": [{
        "file": "config/packages/ai.yaml",
        "position": "after_target",
        "target": "    store:",
        "create_if_missing": "ai:\n    store:",
        "content": "        azuresearch:\n            default:\n                endpoint: '%env(AZURE_SEARCH_ENDPOINT)%'"
    }]
}
```

### Behavior:
- If `store:` exists → Content is added after it (existing behavior)
- If `store:` doesn't exist but `create_if_missing` is set → The `create_if_missing` content is appended to the file, followed by the `content`
- If `store:` doesn't exist and `create_if_missing` is not set → Warning is logged (existing behavior)

### DIFF
--- a/src/Configurator/AddLinesConfigurator.php
+++ b/src/Configurator/AddLinesConfigurator.php
@@ -129,8 +129,9 @@ class AddLinesConfigurator extends AbstractConfigurator
                 continue;
             }
             $target = isset($patch['target']) ? $patch['target'] : null;
+            $createIfMissing = isset($patch['create_if_missing']) ? $patch['create_if_missing'] : null;
 
-            $newContents = $this->getPatchedContents($file, $content, $position, $target, $warnIfMissing);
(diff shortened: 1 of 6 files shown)


fabpot:
We can maybe close this PR?

stof:
My proposal in https://github.com/symfony/recipes/pull/1503#issuecomment-3688082873 would make this PR unnecessary (adding one more advantage to the list)

PR #1077, opened 2026-01-13
Add `SYMFONY_ALLOW_CONTRIB` variable

GromNaN:
This is equivalent to the composer config `extra.symfony.allow-contrib`. When the variable is set, its value has priority over `extra.symfony.allow-contrib`

### DIFF
--- a/src/Flex.php
+++ b/src/Flex.php
@@ -386,7 +386,12 @@ class Flex implements PluginInterface, EventSubscriberInterface
         }
 
         $this->io->writeError(\sprintf('<info>Symfony operations: %d recipe%s (%s)</>', \count($recipes), \count($recipes) > 1 ? 's' : '', $this->downloader->getSessionId()));
-        $installContribs = $this->composer->getPackage()->getExtra()['symfony']['allow-contrib'] ?? false;
+        if (false === $installContribs = getenv('SYMFONY_ALLOW_CONTRIB')) {
+            $installContribs = $this->composer->getPackage()->getExtra()['symfony']['allow-contrib'] ?? false;
+        }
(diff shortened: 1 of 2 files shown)


src/Flex.php
```diff
@@ -122,13 +121,21 @@ class_exists(__NAMESPACE__.str_replace('/', '\\', substr($file, \strlen(__DIR__)
 
         // if Flex is being upgraded, the original operations from the original Flex
         // instance are stored in the static property, so we can reuse them now.
-        if (property_exists(Flex::class, 'storedOperations') && Flex::$storedOperations) {
-            $this->operations = Flex::$storedOperations;
-            Flex::$storedOperations = [];
+        if (property_exists(self::class, 'storedOperations') && self::$storedOperations) {
```
stof: this must be reverted. Not using `self` is intended here due to the special behavior of composer plugin classes during an update of the plugin (see the comment just above)

src/Flex.php
```diff
@@ -213,7 +213,7 @@ public function deactivate(Composer $composer, IOInterface $io)
     {
         // Using `Flex::` instead of `self::` to avoid issues when
         // composer renames plugin classes when upgrading them
-        Flex::$storedOperations = $this->operations;
+        self::$storedOperations = $this->operations;
```
stof: this must be reverted (see the comment just above for the explanation)

stof (changes requested):

src/Flex.php
GromNaN: Thanks for spotting the reason, I blindly applied the fabbot patch.
This is now reverted.

nicolas-grekas:
Thank you @GromNaN.

PR #1083, opened 2026-04-02
Fix "null as array offset" deprecation in DockerComposeConfigurator

### LINKED ISSUE #1079: PHP 8.5 Deprecation Notice: Using null as an array offset is deprecated in DockerComposeConfigurator.php
When installing a new Symfony 8.0 webapp with Composer under PHP 8.5, a depreaction notice is shown:

```
composer create-project symfony/skeleton:"8.0.*" my_project_directory
cd my_project_directory
composer require webapp
```

```
...
Symfony operations: 22 recipes (028901ec6b3613f936547ee3acc5f42f)
  - Configuring symfony/webapp-pack (>=1.0): From github.com/symfony/recipes:main
  - Configuring doctrine/deprecations (>=1.0): From github.com/symfony/recipes:main
  - Configuring doctrine/doctrine-bundle (>=3.0): From github.com/symfony/recipes:main
  -  WARNING  doctrine/doctrine-bundle (>=3.0): From github.com/symfony/recipes:main
    The recipe for this package contains some Docker configuration.

    This may create/update compose.yaml or update Dockerfile (if it exists).

    Do you want to include Docker configuration from recipes?
    [y] Yes
    [n] No
    [p] Yes permanently, never ask again for this project
    [x] No permanently, never ask again for this project
    (defaults to y):
Deprecation Notice: Using null as an array offset is deprecated, use an empty string instead in /.../my_project_directory/vendor/symfony/flex/src/Configurator/DockerComposeConfigurator.php:288
...
```

GromNaN:
Fix #1079

## Summary

PHP 8.5 introduces a deprecation notice when `null` is used as an array offset:

> Using null as an array offset is deprecated, use an empty string instead

In `configureDockerCompose()`, `$node` is initialized to `null` and used as an array key in `$endAt[$node]` before being assigned — once at the first top-level key encounter during the parsing loop, and once after the loop when the file is empty (e.g. a freshly created `compose.yaml`).

## Fix

Guard both `$endAt[$node]` assignments with `null !== $node` checks, since there is no previous section to record an end position for.

### DIFF
--- a/src/Configurator/DockerComposeConfigurator.php
+++ b/src/Configurator/DockerComposeConfigurator.php
@@ -275,7 +275,9 @@ class DockerComposeConfigurator extends AbstractConfigurator
                 }
 
                 // Keep end in memory (check break line on previous line)
-                $endAt[$node] = !$i || '' !== trim($lines[$i - 1]) ? $i : $i - 1;
+                if (null !== $node) {
+                    $endAt[$node] = !$i || '' !== trim($lines[$i - 1]) ? $i : $i - 1;
+                }
                 $node = $matches[1];
                 if (!isset($nodesLines[$node])) {
                     $nodesLines[$node] = [];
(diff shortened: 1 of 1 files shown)


stof (approved):

maxhelias (approved):

nicolas-grekas:
Thank you @GromNaN.

PR #1084, opened 2026-04-02
Fix testEnvCanBeReferenced for dotenv deferred resolution

GromNaN:
`symfony/dotenv` >= 6.4.35/7.4.7/8.0.7 introduced deferred variable resolution (symfony/symfony#63496): variables like `BAR=$FOO` are now resolved after all `.env` files are loaded, using `$_ENV['FOO']` (set by `populate()` from `.env`) rather than the system env at dump time.

This caused `testEnvCanBeReferenced` to fail: `BAR` now resolves to `'123'` (from `.env`) instead of `'Foo'` (from `$_SERVER` via the runtime PHP expression).

The test is updated to use `InstalledVersions::satisfies()` to assert the correct expected value depending on the installed dotenv version, keeping compatibility with both old and new behavior.

### DIFF
--- a/tests/Command/DumpEnvCommandTest.php
+++ b/tests/Command/DumpEnvCommandTest.php
@@ -13,6 +13,8 @@ namespace Symfony\Flex\Tests\Command;
 
 use Composer\Config;
 use Composer\Console\Application;
+use Composer\InstalledVersions;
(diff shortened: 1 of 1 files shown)


nicolas-grekas:
I fixed this by bumping in composer.json

PR #1086, opened 2026-04-13
ci: add PHP 8.5 default run

Chris53897:
CI is only running in lowest deps for PHP 8.5

Is there a reason that there is no default run for PHP 8.5 ?


nicolas-grekas:
that's an optimization of the CI matrix: adding this line wouldn't help spot more issues.

PR #1092, opened 2026-06-05
Fix `null as array offset` when running composer without command

JorickPepin:
```bash
> composer -V
Deprecation Notice: Using null as an array offset is deprecated, use an empty string instead in /var/www/application/vendor/symfony/flex/src/Flex.php:184
Composer version 2.10.1 2026-06-04 10:25:59
PHP version 8.5.5 (/usr/bin/php8.5)
```


stof (approved):

shyim:
could we get this merged also right now ran into 👀

PR #1095, opened 2026-06-25
[EnvConfigurator] Do not escape literal values in env files

### LINKED ISSUE #1096: [EnvConfigurator] Single-quoted values are incorrectly double-quoted in .env files
## Description

When a recipe defines an environment variable whose value is wrapped in single quotes (e.g. to store a JSON array as a dotenv literal), the `EnvConfigurator` incorrectly wraps the value in double quotes and escapes the inner double quotes.

## Steps to reproduce

Define the following in a recipe manifest:

```json
"dotenv": {
  "dev": {
    "ALLOWED_LANGUAGES": "'[\"en\",\"de\",\"es\"]'"
  }
}
```

### Expected result

`ALLOWED_LANGUAGES='["en","de","es"]'`

The dotenv format treats single-quoted values as fully literal strings — no variable interpolation, no escape sequences. This is the standard way to embed JSON content in a .env file.

### Actual result

`ALLOWED_LANGUAGES="'[\"en\",\"de\",\"es\"]'"`

The value is wrapped in double quotes and the inner " are escaped, which changes the semantics: the single quotes become part of the value instead of acting as dotenv quoting delimiters.

### Root cause

In `EnvConfigurator::configureEnvDist()`, a `strpbrk` check detects special characters (including ") and unconditionally wraps the value in double quotes:

```php
if (false !== strpbrk($value, " \t\n&!\"")) {
    $value = '"'.str_replace(['\\', '"', "\t", "\n"], ['\\\\', '\\"', '\t', '\n'], $value).'"';
}
```

This was introduced in #200 to handle values containing spaces, newlines, and other shell-special characters. However, it does not account for values already wrapped in single quotes, which should be written as-is.

DjLeChuck:
_Closes #1096_

## Problem

When a recipe defines a value wrapped in single quotes - for example to store a JSON array:

```json
{
  "dotenv": {
    "dev": {
      "ALLOWED_LANGUAGES": "'[\"en\",\"de\",\"es\"]'"
    }
  }
}
```

The configurator incorrectly wrapped the value in double quotes and escaped the inner double quotes, producing:

`ALLOWED_LANGUAGES="'[\"en\",\"de\",\"es\"]'"`

Instead of the expected single-quoted dotenv literal:

`ALLOWED_LANGUAGES='["en","de","es"]'`

The dotenv format treats single-quoted values as fully literal strings (no escape processing), which makes them the natural fit for JSON content.

## Solution

Skip the double-quote wrapping when the value already starts and ends with a single quote. Any other value continues through the existing escaping path introduced in #200.

```php
// Before
if (false !== strpbrk($value, " \t\n&!\"")) {
  $value = '"'.str_replace([...], [...], $value).'"';
}

// After
if (!str_starts_with($value, "'") && !str_ends_with($value, "'") && false !== strpbrk($value, "\t\n&!\"")) {
  $value = '"'.str_replace([...], [...], $value).'"';
}
```

## Tests

Added a `ALLOWED_LANGUAGES='["en","de","es"]'` fixture to both EnvConfiguratorTest and
DotenvConfiguratorTest to cover this case, alongside the existing assertions for values that require double-quote escaping.

### DIFF
--- a/src/Configurator/EnvConfigurator.php
+++ b/src/Configurator/EnvConfigurator.php
@@ -93,7 +93,7 @@ class EnvConfigurator extends AbstractConfigurator
                 }
 
                 $value = $this->options->expandTargetDir($value);
-                if (false !== strpbrk($value, " \t\n&!\"")) {
+                if (!str_starts_with($value, "'") && !str_ends_with($value, "'") && false !== strpbrk($value, " \t\n&!\"")) {
                     $value = '"'.str_replace(['\\', '"', "\t", "\n"], ['\\\\', '\\"', '\t', '\n'], $value).'"';
                 }
                 $data .= "$key=$value\n";
--- a/tests/Configurator/DotenvConfiguratorTest.php
+++ b/tests/Configurator/DotenvConfiguratorTest.php
@@ -51,6 +51,7 @@ class DotenvConfiguratorTest extends TestCase
                 '#2' => 'Comment 3',
                 '#TRUSTED_SECRET' => 's3cretf0rt3st"<>',
                 'APP_SECRET' => 's3cretf0rt3st"<>',
+                'ALLOWED_LANGUAGES' => '\'["en","de","es"]\'',
             ],
         ], $lock);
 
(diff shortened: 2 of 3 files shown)


DjLeChuck:
Should I fix Fabbot issues about code style?

PR #1098, opened 2026-07-23
Remove importmap entries when uninstalling a package

### LINKED ISSUE #1009: Remove importmap entries when uninstalling a package

Amoifr:
Fix #1009

When AssetMapper is used, Flex runs `importmap:require` for the entries declared by a UX package on install, but nothing removed them on uninstall: the entries stayed in `importmap.php` forever.

The tricky part is that by the time `PackageJsonSynchronizer` runs (`POST_UPDATE_CMD`), the uninstalled package's files are gone, so its `symfony.importmap` config cannot be read anymore. And blindly diffing `importmap.php` against the entries declared by the remaining packages is not an option either: it would also remove the entries the user added manually with `importmap:require`.

So this PR captures the entry names on `PRE_PACKAGE_UNINSTALL`, while the package files still exist, and passes them to the synchronizer, which then runs a single `importmap:remove` for the captured names, except those that are:

- still declared by an installed package (e.g. two bundles sharing a JS dependency), or
- not present in `importmap.php` (e.g. already removed by hand).

The `synchronize()` signature change is backward compatible (new optional parameter), and the webpack/`package.json` path is untouched (obsolete `file:` links were already cleaned up by `removeObsoletePackageJsonLinks()`).

### DIFF
--- a/src/Flex.php
+++ b/src/Flex.php
@@ -32,6 +32,7 @@ use Composer\IO\IOInterface;
 use Composer\IO\NullIO;
 use Composer\Json\JsonFile;
 use Composer\Json\JsonManipulator;
+use Composer\Package\CompletePackageInterface;
 use Composer\Package\Locker;
 use Composer\Package\Package;
 use Composer\Plugin\PluginEvents;
@@ -76,6 +77,7 @@ class Flex implements PluginInterface, EventSubscriberInterface
     private $installer;
     private $postInstallOutput = [''];
     private $operations = [];
+    private $uninstalledImportMapEntryNames = [];
     private $lock;
     private $displayThanksReminder = 0;
     private $ignorePreleases = false;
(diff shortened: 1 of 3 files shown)


Kocal:
What if an other UX package require the same dependency from the package you are removing... does the dependency stay in importmap.php?

Amoifr:
Good catch to check! Yes, it stays: the removal list is filtered against the entries still declared by the installed packages, so anything another package still needs is skipped. This exact scenario is covered by `testSynchronizeAssetMapperRemovesObsoleteImportMapEntries` (`@hotcake/foo` belongs to the removed package but is still declared by `symfony/new-package`, and it survives the uninstall). 😄

